PR #3, opened 2025-02-18
Add Faostat scraping functions

lbm364dl:
Create helper function `get_faostat_data`. Created by @jgserra18. I adapted it and added some tests.


eduaguilera (approved):
I have reviewed the changes, but I didn't understand some of the code, and I could not run the code (I don't know how). Other than that, everything seems fine! :)

lbm364dl:
> I have reviewed the changes, but I didn't understand some of the code, and I could not run the code (I don't know how). Other than that, everything seems fine! :)

For future reference, there is no need for you to make sure the code runs (we assume the automatic checks do that), but if you do want to try it, you should:
- Get this branch locally in case you do not have it yet (`git fetch origin branch-name`)
- Get into this branch to actually see its contents (`git checkout branch-name`)
- For testing functions from the `R` directory, you can get them ready by using `devtools::load_all()`, which is also common to use while developing, so that you can try your own functions interactively.

jgserra18:
@eduaguilera did you try the examples from get_faostat_data? I just tried it after Catalin's refactoring to tidyverse and they work!
What part of the code didn't you understand?
Captura de ecrã 2025-03-03 182011

PR #6, opened 2025-04-01
Prepare input files for applying FABIO

lbm364dl:
This is a first PR with the FABIO model in mind. The idea is to try to implement it by ourselves in future PRs. For now, here I did the following:

- Read some files already processed by @eduaguilera elsewhere, as input files here (commodity balance sheet, processed product shares, bilateral trade data)
- Create some tests to check data integrity (there are some data observations which should be fixed, see tests TODO comments). For data testing I found the [pointblank R package](https://rstudio.github.io/pointblank/) useful.
- Define small workflow on how to work with input files. Use custom function `get_file_path`, which receives an alias for the input file, and it will be downloaded first if not present and stored in a cache folder, or fetched directly from cache. It relies on input files being uploaded to Google Drive.
- Add tests for the input file reading mechanism.
- Experiment using mocks in tests with the [`testthat::local_mocked_bindings`](https://testthat.r-lib.org/reference/local_mocked_bindings.html) function.


lbm364dl:
> Great! I could not follow everything, but if you want to discuss something in particular, just tell me!

Maybe you can check this last commit I added with function documentation and see if it makes sense. Most of it is for explaining the columns of the data I got from you.

R/bilateral_trade.R
```diff
@@ -1,3 +1,43 @@
...
+#'    item. Equivalences with ISO 3166-1 numeric can be found in the
+#'    _Area Codes_ CSV from the zip file that can be downloaded from
+#'    [FAOSTAT](https://www.fao.org/faostat/en/#data/FBS). TODO: Think about
+#'    this, would be nice to use ISO3 codes but won't be enough for our periods
+#' - `to_code`: FAOSTAT internal code for the country that is importing the
+#'    item. See previous `from_code`.
+#' - `item`: Natural language name for the item that is being traded.
+#' - `unit`: Measure unit for the traded item. It can have two values:
+#'      - `"tonnes"`: Trade amount in tonnes (for crop items)
+#'      - `"heads"`: Trade amount in number of animals (for livestock items)
```
eduaguilera, line 23: It would be more appropriate to say "for live animal items"

R/bilateral_trade.R
```diff
@@ -1,3 +1,43 @@
...
+#' - `from_code`: FAOSTAT internal code for the country that is exporting the
+#'    item. Equivalences with ISO 3166-1 numeric can be found in the
+#'    _Area Codes_ CSV from the zip file that can be downloaded from
+#'    [FAOSTAT](https://www.fao.org/faostat/en/#data/FBS). TODO: Think about
+#'    this, would be nice to use ISO3 codes but won't be enough for our periods
+#' - `to_code`: FAOSTAT internal code for the country that is importing the
+#'    item. See previous `from_code`.
+#' - `item`: Natural language name for the item that is being traded.
+#' - `unit`: Measure unit for the traded item. It can have two values:
+#'      - `"tonnes"`: Trade amount in tonnes (for crop items)
```
eduaguilera, line 22: animal products are also expressed in tonnes, so it would be better to say "crop and livestock products"

R/bilateral_trade.R
```diff
@@ -1,3 +1,43 @@
...
+#'      - `"heads"`: Trade amount in number of animals (for livestock items)
+#' - `value`: The amount traded in the corresponding measure unit.
+#'
+#' For the final data obtained, the export data was preferred, i.e.,
```
eduaguilera, line 26: You did that? I think FABIO now uses the average of the export and import data (I recall Martin said that in the meeting). Anyway, I would follow FABIO's procedure for balancing

R/bilateral_trade.R
```diff
@@ -1,3 +1,43 @@
...
+#' @param file_path The local path where the input CSV is located
+#'
+#' @returns
+#' A tibble with the reported trade between countries.
+#' It contains the following columns:
+#' - `year`: The year when the data was collected.
+#' - `from_code`: FAOSTAT internal code for the country that is exporting the
+#'    item. Equivalences with ISO 3166-1 numeric can be found in the
+#'    _Area Codes_ CSV from the zip file that can be downloaded from
+#'    [FAOSTAT](https://www.fao.org/faostat/en/#data/FBS). TODO: Think about
```
eduaguilera, line 16: This TODO could be discussed in the Harmonization seminar. Maybe we could stablish our own ISO3 codes for past countries which don't have ISO3 code. Alejandra is checking how they do it here:  https://github.com/aourednik/historical-basemaps

R/commodity_balance_sheet.R
```diff
@@ -1,3 +1,54 @@
...
+#' @description
+#' Clean CSV data from the given `file_path`. It is recommended to use an
+#' autogenerated path by `get_file_path()` function.
+#'
+#' @param file_path The local path where the input CSV is located
+#'
+#' @returns
+#' A tibble with the commodity balance sheet data in wide format.
+#' It contains the following columns:
+#' - `year`: The year when the data was collected.
```
eduaguilera, line 12: I think "when the data was collected" is a misleading expression... it makes me think on the date when the survey was made, but a survey may collect data from many previous years. I would say something like "the year in which the flow took place" (this also applies to other datasets)

R/bilateral_trade.R
lbm364dl: I did this by following FABIO's code. There's a part that looks like this:
```r
# Give preference to reported export flows over import flows
btd <- flow_pref(btd, pref = "Export")
```
This is not about the general bilateral trade balancing (I'm working on that separately). Here it's only about using one reporting value from the bilateral trade data among the reporter country and the partner country. While taking the average between both reports could also sound OK, I just followed what FABIO does.
I think what you recall is his explanation about estimating missing data, where they compute a kind of average among imports and exports.
Even when having to use an equal amount for total exports and imports for the balancing algorithm, they don't use an average among both. Instead, they keep the smaller value and they downscale the other to match it.

PR #7, opened 2025-04-02
Remove unnecessary auth

lbm364dl:
Thanks to a previous PR where I had to read public files from Google Drive, I noticed our auth for reading an Excel sheet was unnecessary. Calling `googlesheets::gs4_deauth()` before reading a public Google Drive file is enough.


justin-morgan-csic (approved):
molt be

PR #8, opened 2025-04-09
Estimate and balance bilateral trade

lbm364dl:
Implement FABIO model's choices for bilateral trade data estimation and balancing.
- Use square matrices to represent bilateral trade for each year and item.
- Complete with missing trading pairs and estimate their values following FABIO.
- Balance matrix sums using iterative proportional fitting.
- Add tests for different subfunctions.

Currently, doing all these steps (i.e. calling the `get_bilateral_trade` function) takes roughly 5 mins in my computer. I'm pretty sure it can be improved. I think my method is quite clean but it relies on using `purrr::map` to manipulate non-vectorial type columns, in this case, nested tibbles and matrices, which might be slow. I think it's a decent time for now anyway, so I'll leave that for another time.


eduaguilera (approved):
Everything seems OK!

PR #9, opened 2025-04-28
Optimize bilateral trade and allow public link downloads

lbm364dl:
For the optimization part I was inspired by [the profvis package](https://profvis.r-lib.org/) when looking for profiling tools. Running it I could see which parts took longer.
- Restructure code to use a single call to `purrr::map` equivalents. Check function `.process_bilateral_trade`
- Implement `.complete_total_trade` without direct use of `tidyr::complete`, which was one of the slowest parts
- Now a call to `get_bilateral_trade` takes approx. half of what it took before, and half of this time is the `mipfp::Ipfp` matrix balancing bottleneck

On links, thanks to @eduaguilera for suggesting going back to using Nextcloud storage. I wrote a more general input treatment so that data can be downloaded from any public URL, also making the code neater with the use of [the httr2 package](https://httr2.r-lib.org/). The Google Drive download solution was left there as an option in case someone needs it (basically I think downloads are faster there).

### DIFF
--- a/R/bilateral_trade.R
+++ b/R/bilateral_trade.R
@@ -282,10 +282,7 @@ get_bilateral_trade <- function(file_path) {
 
   total_trade |>
     dplyr::right_join(combs, by = c("year", "item", "area_code")) |>
-    dplyr::mutate(
-      export = ifelse(is.na(export), 0, export),
-      import = ifelse(is.na(import), 0, import)
-    )
+    tidyr::replace_na(list(export = 0, import = 0))
 }
 
 .filter_only_items_in_cbs <- function(btd, cbs) {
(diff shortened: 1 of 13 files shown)


R/bilateral_trade.R
```diff
@@ -249,47 +240,52 @@ get_bilateral_trade <- function(file_path) {
...
+
   total_trade |>
-    tidyr::complete(
-      area_code = codes,
-      fill = list(export = 0, import = 0)
-    ) |>
-    dplyr::arrange(area_code)
+    dplyr::right_join(combs, by = c("year", "item", "area_code")) |>
+    dplyr::mutate(
+      export = ifelse(is.na(export), 0, export),
```
eduaguilera, line 286: you could use tidyr::replace_na(export,0)

eduaguilera (approved):

PR #10, opened 2025-04-30
Avoid using renv in Github actions

lbm364dl:
The initial motivation for this PR was that we were having issues with packages not installed correctly in Github actions because of missing Operating System dependencies (my mistake to ask on https://github.com/r-lib/pak/issues/769, but they were helpful). This can be handled automatically using the `setup-r-dependencies` action (which internally uses [the pak package](https://pak.r-lib.org/)) but I found out it conflicts with `renv` (e.g. cache not working https://github.com/r-lib/actions/issues/911 and other obscure errors I got). In the end, `renv` should not be used in our checks anyway because whoever might want to install this package might not use it, and other dependency versions might be installed from `DESCRIPTION` file. Only we as developers should use `renv`, but it should be removed from Github actions. What I did:

- Change our workflows back to the default ones that you can get using [`usethis::use_github_action(name)`](https://usethis.r-lib.org/reference/github_actions.html)
- Use an env variable `USE_RENV` to conditionally avoid using `renv` in `.Rprofile` (got the idea from [this repository](https://github.com/r-world-devs/GitStats/blob/master/.github/workflows/test-coverage.yaml))
- Add `lintr` check in a separate workflow, just because it's cleaner (it was previously included in the `R cmd check` workflow)


eduaguilera (approved):
It seems reasonable to me to make the change, even if I don't fully understand what you did (but I think there is no need I understand it, as long as it works!)

lbm364dl:
> It seems reasonable to me to make the change, even if I don't fully understand what you did (but I think there is no need I understand it, as long as it works!)

I agree this is not directly related to the project and you can skip it, but whenever there is something you _do_ want to understand, don't be afraid to ask.

PR #11, opened 2025-05-07
Edu/add old functions

eduaguilera:
Create gapfilling functions

### DIFF
--- a/R/gapfilling.R
+++ b/R/gapfilling.R
@@ -18,12 +18,14 @@ sample_tibble <- tibble::tibble(
   change_variable =c(1,1,1,1,1,1,0,0,0,0,0,0)
 )
 
-filled_tibble <- linear_fill(sample_tibble, value, year, category)
+linear_filled_tibble <- linear_fill(sample_tibble, value, year, category)
+proxy_filled_tibble <- proxy_fill(sample_tibble, value, proxy_variable, year, category)
+sum_filled_tibble <- sum_fill(sample_tibble, value, change_variable, category)
 
 # Fills gaps by linear interpolation, when possible, or carrying forward or backwards, when not possible, and labels output accordingly
 # The data should be he data when needed before running the function
-linear_fill <- function(data, var, time_index, ...) { # data= data frame; var = variable to be filled; time_index = time index (usually year); ... = grouping variables
-  data |> 
+linear_fill <- function(df, var, time_index, ...) { # df = data frame; var = variable to be filled; time_index = time index (usually year); ... = grouping variables
+  df |> 
     dplyr::group_by(...) |> 
     dplyr::mutate(
       Value_interpfilled = zoo::na.approx({{ var }},
@@ -59,18 +61,18 @@ linear_fill <- function(data, var, time_index, ...) { # data= data frame; var = 
 
 # Fills gaps by using changes in a proxy variable, using ratios between filled variable and proxy variable, and labels output accordingly
 # Remember to group the data when needed before running the function
-proxy_fill <- function(data, var, proxyvar, time_index, ...) { # data=dataframe, var=variable to be filled, proxyvar=variable used as proxy, Index=time index (usually year)
-  data %>%
-    dplyr::mutate(proxy_ratio = {{ var }} / {{ proxyvar }}) %>%
-    linear_fill(proxy_ratio, time_index) %>%
+proxy_fill <- function(df, var, proxy_var, time_index, ...) { # df = data frame, var = variable to be filled, proxy_var = variable used as proxy, time_index = time index (usually year); ... = grouping variables
+  df |>
+    dplyr::mutate(proxy_ratio = {{ var }} / {{ proxy_var }}) |>
+    linear_fill(proxy_ratio, {{ time_index }}, ...) |>
     dplyr::mutate(
       "Source_{{var}}" := ifelse(!is.na({{ var }}),
                                  "Original",
-                                 ifelse(Source_Proxy_ratio == "Linear interpolation",
+                                 ifelse(Source_proxy_ratio == "Linear interpolation",
                                         "Proxy interpolated",
-                                        ifelse(Source_Proxy_ratio == "Last value carried forward",
+                                        ifelse(Source_proxy_ratio == "Last value carried forward",
                                                "Proxy carried forward",
-                                               ifelse(Source_Proxy_ratio == "First value carried backwards",
+                                               ifelse(Source_proxy_ratio == "First value carried backwards",
                                                       "Proxy carried backwards",
                                                       NA
                                                )
@@ -79,17 +81,39 @@ proxy_fill <- function(data, var, proxyvar, time_index, ...) { # data=dataframe,
       ),
       "{{var}}" := ifelse(!is.na({{ var }}),
                           {{ var }},
-                          proxy_ratio * {{ proxyvar }}
+                          proxy_ratio * {{ proxy_var }}
       )
     )
 }
 
 # Function to fill gaps in a given variable (var) of a time series with the sum of the previous value of var and the value of another column (change_var). The values of var are accumulated along the series.
-sum_fill <- function(var, change_var) {
-  for (i in seq_along(var)) {
-    if (is.na(var[i]) && i > 1) {
-      var1[i] <- var1[i - 1] + change_var[i]
-    }
-  }
-  return(var)
+sum_fill <- function(df, var, change_var, ...) {
+  var_sym <- rlang::ensym(var)
+  change_var_sym <- rlang::ensym(change_var)
+  var_name <- rlang::as_string(var_sym)
+  change_var_name <- rlang::as_string(change_var_sym)
+  
+  group_syms <- rlang::ensyms(...)
+  
+  df <- dplyr::group_by(df, !!!group_syms) |>
+    dplyr::group_modify(function(.x, .y) {
+      var_vec <- .x[[var_name]]
+      change_vec <- .x[[change_var_name]]
+      
+      status_vec <- ifelse(is.na(var_vec), NA, "Original")
+      
+      for (i in seq_along(var_vec)) {
+        if (is.na(var_vec[i]) && i > 1) {
+          var_vec[i] <- var_vec[i - 1] + change_vec[i]
+          status_vec[i] <- "Filled with sum"
+        }
+      }
+      
+      .x[[var_name]] <- var_vec
+      .x[[paste0("Source_",var_name)]] <- status_vec
+      return(.x)
+    }) |>
+    dplyr::ungroup()
+  
+  return(df)
 }
--- a/R/gapfilling.R
+++ b/R/gapfilling.R
@@ -1,29 +1,27 @@
-#' Fill the gaps (NA values) in a time-dependent variable, creating complete time series
-#'
-#' @param trade_sources A tibble dataframe
-#' where each row contains the year range
-#'
-#' @return A tibble dataframe where each row
-#' corresponds to a single year for a given source
+# Simple functions to fill gaps (NA values) in a time-dependent variable, creating complete time series
+
+
+#' Fill gaps in a variable by linear interpolation, when possible, or carrying forward or backwards, 
+#' when not possible, and label output accordingly
+#' 
+#' @param df A tibble data frame containing one observation per row
+#' @param var The variable of df containing gaps to be filled
+#' @param time_index The time index variable (usually year)
+#' @param ... The grouping variables (optional)
+#' 
+#' @return A tibble data frame (ungrouped) where gaps in var have been filled, 
+#' and a new "Source" variable has been created indicating if the value is original or,
+#' in case it has been estimated, the gapfilling method that has been used
 #'
 #' @export
 #'
 #' @examples
-
-sample_tibble <- tibble::tibble(
-  category = c("a", "a", "a", "a", "a", "a", "b", "b", "b", "b", "b", "b"),
-  year = c("2015", "2016", "2017", "2018", "2019", "2020", "2015", "2016", "2017", "2018", "2019", "2020"),
-  value = c(NA, 3, NA, NA, 0, NA, 1, NA, NA, NA, 5, NA),
-  proxy_variable = c(1,2,2,2,2,2,1,2,3,4,5,6),
-  change_variable =c(1,1,1,1,1,1,0,0,0,0,0,0)
-)
-
-linear_filled_tibble <- linear_fill(sample_tibble, value, year, category)
-proxy_filled_tibble <- proxy_fill(sample_tibble, value, proxy_variable, year, category)
-sum_filled_tibble <- sum_fill(sample_tibble, value, change_variable, category)
-
-# Fills gaps by linear interpolation, when possible, or carrying forward or backwards, when not possible, and labels output accordingly
-# The data should be he data when needed before running the function
+#' sample_tibble <- tibble::tibble(
+#'   category = c("a", "a", "a", "a", "a", "a", "b", "b", "b", "b", "b", "b"),
+#'   year = c("2015", "2016", "2017", "2018", "2019", "2020", "2015", "2016", "2017", "2018", "2019", "2020"),
+#'   value = c(NA, 3, NA, NA, 0, NA, 1, NA, NA, NA, 5, NA),
+#' )
+#' linear_filled_tibble <- linear_fill(sample_tibble, value, year, category)
 linear_fill <- function(df, var, time_index, ...) { # df = data frame; var = variable to be filled; time_index = time index (usually year); ... = grouping variables
   df |> 
     dplyr::group_by(...) |> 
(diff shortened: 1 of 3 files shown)


R/gapfilling.R
```diff
@@ -0,0 +1,162 @@
...
+#'
+#' @export
+#'
+#' @examples
+#' sample_tibble <- tibble::tibble(
+#'   category = c("a", "a", "a", "a", "a", "a", "b", "b", "b", "b", "b", "b"),
+#'   year = c("2015", "2016", "2017", "2018", "2019", "2020", "2015", "2016", "2017", "2018", "2019", "2020"),
+#'   value = c(NA, 3, NA, NA, 0, NA, 1, NA, NA, NA, 5, NA),
+#' )
+#' linear_filled_tibble <- linear_fill(sample_tibble, value, year, category)
```
lbm364dl, line 24: If this line is the final output you want to show as an example, it shouldn't be assigned to a variable, because then it doesn't show up in the documentation site. This last line should just be
```r
linear_fill(sample_tibble, value, year, category)
```
without the assignment (same for the other functions).

R/gapfilling.R
```diff
@@ -0,0 +1,162 @@
...
+linear_fill <- function(df, var, time_index, ...) { # df = data frame; var = variable to be filled; time_index = time index (usually year); ... = grouping variables
+  df |> 
+    dplyr::group_by(...) |> 
+    dplyr::mutate(
+      Value_interpfilled = zoo::na.approx({{ var }},
```
lbm364dl, line 29: Remember you have to add new package dependencies in the `DESCRIPTION` file. In this case, `zoo` would need to be added in the `Imports` section.

R/gapfilling.R
```diff
@@ -0,0 +1,162 @@
...
+#'
+#' @examples
+#' sample_tibble <- tibble::tibble(
+#'   category = c("a", "a", "a", "a", "a", "a", "b", "b", "b", "b", "b", "b"),
+#'   year = c("2015", "2016", "2017", "2018", "2019", "2020", "2015", "2016", "2017", "2018", "2019", "2020"),
+#'   value = c(NA, 3, NA, NA, 0, NA, 1, NA, NA, NA, 5, NA),
+#'   proxy_variable = c(1,2,2,2,2,2,1,2,3,4,5,6)
+#' )
+#' proxy_filled_tibble <- proxy_fill(sample_tibble, value, proxy_variable, year, category)
+proxy_fill <- function(df, var, proxy_var, time_index, ...) { # df = data frame, var = variable to be filled, proxy_var = variable used as proxy, time_index = time index (usually year); ... = grouping variables
```
lbm364dl, line 84: As a rule of thumb, comments should never be placed at the end of a code line. They should be on their own line.

renv.lock
```diff
@@ -200,7 +200,7 @@
     },
```
lbm364dl, line 200: I'm not convinced about all these changes in package versions. It this was done correctly, I would assume only the `zoo` part should be added. We can check this at another time though.

lbm364dl (changes requested):
About the error we saw yesterday in the `pkgdown` site examples saying that the functions were not found, I couldn't reproduce it, it worked fine for me (after also adding `zoo` to `DESCRIPTION` imports). Few things that come to mind, maybe try to manually delete the `man` folder and generate again using `devtools::document()`, and also double check if your new functions appear in the `NAMESPACE` file. You may also try rerunning code in clean R sessions when something goes wrong, just in case.

I will leave more comments about the code itself after you complete the PR with style fixes and tests.

lbm364dl:
Check the formatting issues. It can also be fixed directly from here. 

1. Go to files changed tab.
2. Search bot comments. Click _add suggestion to batch_ for all of them.
image
3. In the end, commit those suggestions
image

This can be handy if you don't have a local `air format` installation or it's just a few suggestions. Otherwise you should install it and apply the format locally. Follow instructions at https://posit-dev.github.io/air/formatter.html

tests/testthat/test_gapfilling.R
```diff
@@ -0,0 +1,687 @@
...
+      2020,
+      2015,
+      2016,
+      2017,
+      2018,
+      2019,
+      2020
+    ),
+    value = c(NA, 3, NA, NA, 0, NA, 1, NA, NA, NA, 5, NA),
+    proxy_variable = c(1, 2, 2, 2, 2, 2, 1, 2, 3, 4, 5, 6),
```
lbm364dl, line 20: Only add `proxy_variable`  and `change_variable` when they're used in the test. Otherwise this can be confusing.

R/gapfilling.R
```diff
@@ -0,0 +1,217 @@
...
+#'
+#' @examples
+#' sample_tibble <- tibble::tibble(
+#'   category = c("a", "a", "a", "a", "a", "a", "b", "b", "b", "b", "b", "b"),
+#'   year = c("2015", "2016", "2017", "2018", "2019", "2020", "2015", "2016", "2017", "2018", "2019", "2020"),
+#'   value = c(NA, 3, NA, NA, 0, NA, 1, NA, NA, NA, 5, NA),
+#'   proxy_variable = c(1,2,2,2,2,2,1,2,3,4,5,6)
+#' )
+#'proxy_fill(sample_tibble, value, proxy_variable, year, category)
+proxy_fill <- function(df, var, proxy_var, time_index, ...) {
```
lbm364dl, line 132: I'm having a hard time understanding why this proxy interpolation is meaningful, but it's not my area of expertise. Do you have any good reference?

R/gapfilling.R
```diff
@@ -0,0 +1,217 @@
...
+#' sample_tibble <- tibble::tibble(
+#'   category = c("a", "a", "a", "a", "a", "a", "b", "b", "b", "b", "b", "b"),
+#'   year = c("2015", "2016", "2017", "2018", "2019", "2020", "2015", "2016", "2017", "2018", "2019", "2020"),
+#'   value = c(NA, 3, NA, NA, 0, NA, 1, NA, NA, NA, 5, NA),
+#'   change_variable =c(1,2,3,4,1,1,0,0,0,0,0,1)
+#' )
+#'sum_fill(sample_tibble, value, change_variable, category)
+#'sum_fill(sample_tibble, value, change_variable, TRUE, category)
+sum_fill <- function(df, var, change_var, start_with_zero = FALSE, ...) {
+  var_sym <- rlang::ensym(var)
```
lbm364dl, line 190: I think the whole function can be simplified and it should also be faster. For me for loops in data analysis are a red flag. Whenever you find yourself writing a loop try to think of a better solution. I had fun trying to come up with something better. This is my proposal:

```r 
df |>
  dplyr::group_by(...) |>
  dplyr::mutate(
    groups = cumsum(!is.na({{ var }})),
    prefilled = dplyr::coalesce({{ var }}, {{ change_var }}),
    value = ave(prefilled, groups, FUN = cumsum),
    value = ifelse(groups != 0, value, if (start_with_zero) value else NA),
    source_value = ifelse(is.na({{ var }}), "Filled with sum", "Original")
  ) |>
  dplyr::ungroup()
```

Another thing I don't like is the arbitrary value of 0 when using `start_with_zero = TRUE`. For me the natural behaviour would be to assume there is an invisible 0 value before the first observation, so that we can start filling the next ones with cumulative sum starting from zero. But then if the first value in `change_var` is x, the first value in `var` should also be x, not 0. I think it's more natural. I'm already doing that in my proposed code. I also find it unnecessary to use "Started with zero as a source" and I would just use "Filled with sum" anyway. What do you think?

lbm364dl (changes requested):

R/gapfilling.R
```diff
@@ -0,0 +1,217 @@
...
+#'
+#' @examples
+#' sample_tibble <- tibble::tibble(
+#'   category = c("a", "a", "a", "a", "a", "a", "b", "b", "b", "b", "b", "b"),
+#'   year = c("2015", "2016", "2017", "2018", "2019", "2020", "2015", "2016", "2017", "2018", "2019", "2020"),
+#'   value = c(NA, 3, NA, NA, 0, NA, 1, NA, NA, NA, 5, NA),
+#' )
+#'linear_fill(sample_tibble, value, year, category)
+#'linear_fill(sample_tibble, value, year, category, method = "interpolate")
+linear_fill <- function(df, var, time_index, ..., method = "fill_everything") {
```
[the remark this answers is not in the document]
eduaguilera: I agree this approach is better! implemented!

R/gapfilling.R
```diff
@@ -0,0 +1,217 @@
...
+          Value_interpfilled,
+          ifelse(
+            !is.na(Value_CarriedBackward),
+            Value_CarriedBackward,
+            Value_CarriedForward
+          )
+        )
+      )
+    ) |>
+    dplyr::select(
```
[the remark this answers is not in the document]
lbm364dl: I'm comparing this approach to that of using `dplyr::select` to remove the temporary columns, which is what you do. This alternative looks nicer to me but you can decide, this is subjective I guess.

R/gapfilling.R
eduaguilera: A common example could be estimating the missing yield values of a crop using the trends in another similar crop (e.g. we may use the trends in wheat or barley to estimate the yield of minor cereals with less data availability like rye or oats). This make sense because many variables, like yields change similarly for different crops (e.g. due to wheather conditions). We could also use the trends of a similar country to estimate what happens in a given country. This could make sense for example if we know that they have gone through similar historical processes (e.g. the effects of world wars or economic downturns on trade patterns).
You can see specific examples in my repos Spain_Hist and Global.

R/gapfilling.R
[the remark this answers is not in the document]
eduaguilera: Ok, I get it now... but I'm not sure how to implement it here...

.Rprofile
```diff
@@ -1,5 +1,5 @@
 if (Sys.getenv("USE_RENV", unset = "TRUE") == "TRUE") {
-  source("renv/activate.R")
+  # source("renv/activate.R")
```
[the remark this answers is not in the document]
eduaguilera: Not sure how this should be now...

eduaguilera (commented):
@lbm364dl, I tried to address all your comments. I don't know why but many dissappeared in the process, so I may have missed some...

R/gapfilling.R
eduaguilera: I tried your function, but the tests failed because the values were summed cummulatively while the original values were ignored... I tried many corrections until this one finally worked

R/gapfilling.R
```diff
@@ -0,0 +1,217 @@
...
+#' @examples
+#' sample_tibble <- tibble::tibble(
+#'   category = c("a", "a", "a", "a", "a", "a", "b", "b", "b", "b", "b", "b"),
+#'   year = c("2015", "2016", "2017", "2018", "2019", "2020", "2015", "2016", "2017", "2018", "2019", "2020"),
+#'   value = c(NA, 3, NA, NA, 0, NA, 1, NA, NA, NA, 5, NA),
+#'   change_variable =c(1,2,3,4,1,1,0,0,0,0,0,1)
+#' )
+#'sum_fill(sample_tibble, value, change_variable, category)
+#'sum_fill(sample_tibble, value, change_variable, TRUE, category)
+sum_fill <- function(df, var, change_var, start_with_zero = FALSE, ...) {
```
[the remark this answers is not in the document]
eduaguilera: I agree with the suggestion but I am not sure how the function should be changed... Is the ".by=" also added inside the mutate? Could you directly make the correction to the function?

R/gapfilling.R
```diff
@@ -0,0 +1,192 @@
...
+#'     "2015", "2016", "2017", "2018", "2019", "2020"
+#'   ),
+#'   value = c(NA, 3, NA, NA, 0, NA, 1, NA, NA, NA, 5, NA),
+#'   proxy_variable = c(1, 2, 2, 2, 2, 2, 1, 2, 3, 4, 5, 6)
+#' )
+#' proxy_fill(sample_tibble, value, proxy_variable, year, category)
+proxy_fill <- function(df, var, proxy_var, time_index, ...) {
+  df |>
+    dplyr::mutate(proxy_ratio = {{ var }} / {{ proxy_var }}) |>
+    linear_fill(proxy_ratio, {{ time_index }}, ...) |>
```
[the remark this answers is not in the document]
lbm364dl: I assumed it's already possible. Have you tried it? I thought that's what you do by passing the `...` back into the `linear_fill` call.

R/gapfilling.R
```diff
@@ -0,0 +1,198 @@
...
+#'   ),
+#'   value = c(NA, 3, NA, NA, 0, NA, 1, NA, NA, NA, 5, NA),
+#' )
+#' linear_fill(sample_tibble, value, year, category)
+#' linear_fill(sample_tibble, value, year, category, interpolate = FALSE)
+linear_fill <- function(
+  df,
+  var,
+  time_index,
+  ...,
```
eduaguilera, line 45: It is not clear to me how to replace this approach to grouping data (`...` in parameter definition, then `group_by` and `ungroup` in the function) with the proposal of @lbm364dl (`.by = NULL` in parameter definition, then what in the function?). @lbm364dl, could you please implement your suggestion in the three functions?

R/gapfilling.R
lbm364dl: I understand why we would want to estimate some missing values of one variable using another one. This makes sense if we assume there is a close linear relationship between both. If we were to look at the ratios of non-missing observations, I would expect them to be very similar. Then there wouldn't be a relevant difference between doing a ratio interpolation before multiplying or just multiplying the proxy value by the ratio of the previous observation. 

If on the other hand we have ratios that are very far from each other, I don't really know what kind of relationship the ratio interpolations are trying to model, but it's clearly not a linear one. All I wanted to say is that we should be careful about using this estimate for non-linear relationships between two variables, as I myself don't really know what it would represent.

R/gapfilling.R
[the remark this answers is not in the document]
lbm364dl: I guess the confusion comes from using named vs unnamed arguments in `...`. In the current implementation, all you expected in `...` were grouping variables, so it was passed like a list and treated equally. But if you pass a named argument as part of `...`, it will try to match that name with an existing argument of the function, if there is one. So in this example, if you call something like `proxy_fill(df, var, proxy_var, time_index, category, interpolate = FALSE)`, when sending the `...` to `linear_fill`, the `interpolate = TRUE` would go to the actual `interpolate` argument from `linear_fill` and `category` would go to the `...` of `linear_fill`.

R/gapfilling.R
eduaguilera: Another problem of proxy_fill but also of linear_fill functions is that they take the first and last value of a gap to make the interpolation, which may be problematic if for example one of those values were an outlier for whatever reason. I think it would be better to take an interval instead of a single value. Anyway I think for the moment we can live with them as they are, and leave this as a TODO for the future.

lbm364dl: I encourage you to add any package related `TODO`s as GitHub issues in this repository so that we don't forget about them.

R/gapfilling.R
lbm364dl: I have adapted the code to use `.by` argument and fixed tests, examples and docs accordingly. Have a look at the latest commits, especially [Use .by argument for groups](https://github.com/eduaguilera/whep/pull/11/commits/f6c4117855edccb16dce8788603f064150a1f572).

lbm364dl (changes requested):

R/gapfilling.R
eduaguilera: Thnaks for the suggestion! Done!

tests/testthat/test_gapfilling.R
eduaguilera, line 1: @lbm364dl, I have revised all the test file considering your comments (i.e. removing unnecessary tests, refactoring using `tibble:tribble`, `dplyr::pull` and `pointblank` functions when possible). I have also reorganized it by creating fixture tables that can be used in more than one test, instead of creating a new table in each test from scratch (this was suggested by GitHub Copilot)

tests/testthat/test_gapfilling.R
```diff
@@ -0,0 +1,395 @@
...
+  )
+
+  result <- sum_fill(simple_series, value, change_variable)
+
+  result |>
+    pointblank::expect_col_vals_equal(value, c(10, 12, 15, 16)) |>
+    pointblank::expect_col_vals_in_set(
+      source_value,
+      c("Original", "Filled with sum")
+    )
```
lbm364dl, line 304: Again probably just a matter of style, but I like writing the pipeline directly:
```suggestion
  simple_series |>
    sum_fill(value, change_variable) |>
    pointblank::expect_col_vals_equal(value, c(10, 12, 15, 16)) |>
    pointblank::expect_col_vals_in_set(
      source_value,
      c("Original", "Filled with sum")
    )
```

tests/testthat/test_gapfilling.R
```diff
@@ -0,0 +1,395 @@
...
+  zero_start <- sum_fill(
+    contiguous_gaps,
+    value,
+    change_variable,
+    start_with_zero = TRUE
+  )
+
+  zero_start |>
+    pointblank::expect_col_vals_equal(value, c(1, 3, 6, 10)) |>
+    pointblank::expect_col_vals_equal(source_value, rep("Filled with sum", 4))
```
lbm364dl, line 330: You can write single value and it checks that value for all rows:
```suggestion
    pointblank::expect_col_vals_equal(source_value, "Filled with sum")
```

tests/testthat/test_gapfilling.R
```diff
@@ -0,0 +1,395 @@
...
+
+  result <- sum_fill(
+    grouped,
+    value,
+    change,
+    start_with_zero = TRUE,
+    .by = "grp"
+  )
+
+  totals <- result |>
```
lbm364dl, line 352: I think this approach is a bit unconventional and also inconsistent with the rest of your tests. In the others you do filter on group followed by group values test.

lbm364dl (commented):
I think the tests look better now. I managed to read through all of them so I really think they're easier to maintain now. I also agree that the fixtures usage is clean. Most of the comments I left are related to making the code even cleaner, I hope they help, but as I said in them, you don't have to apply all of them if you think your style is better.

tests/testthat/test_gapfilling.R
```diff
@@ -0,0 +1,395 @@
...
+    dplyr::arrange(grp)
+
+  testthat::expect_equal(totals$grp, c("A", "B"))
+  testthat::expect_equal(totals$values[[1]], c(1, 3, 6))
+  testthat::expect_equal(totals$values[[2]], c(5, 7, 11))
+})
+
+# Integration -------------------------------------------------------------------
+
+testthat::test_that("gapfilling helpers compose in pipelines", {
```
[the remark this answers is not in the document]
eduaguilera: Yes, you are right!

lbm364dl:
> I added @jinfama as a reviewer because he is also working on gapfilling functions and he has many suggestions and also other functions to propose... @lbm364dl, do you think @jinfama should give his feedback on this PR (it may take a few weeks) or we should close this one and make a new one?

When we're done with my suggestions we can merge. @jinfama can add his review when he has time, and then we can create an issue with his suggestions and do them in a new PR. Even if a PR is merged/closed, you can still add your review or comments.

> About the testthat-problems.rds file, I did not create it on purpose...

> Shall I delete it and add it to .gitignore?

You have no snapshot tests in the current code. Maybe you tried to add one at some point. Anyway, I think this file shouldn't appear again if you currently have no snapshot tests. Just delete the file, no need to add to `.gitignore` for now.

lbm364dl (changes requested):
A few small comments. We're almost there!

PS: Remember to also remove the `testthat-comments.rds` file.

tests/testthat/test_gapfilling.R
```diff
@@ -0,0 +1,361 @@
...
+      source_value,
+      c(
+        "Original",
+        "Proxy interpolated",
+        "Proxy carried forward",
+        "Proxy carried backwards"
+      )
+    )
+
+  result |>
```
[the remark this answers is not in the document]
eduaguilera: ok. This could not be checked with pointblak, but given that we no longer group the dataframes, I removed it:   dplyr::is_grouped_df(result) |>
proxy_fill_fixture() |>
testthat::expect_false()

lbm364dl (approved):
LGTM! Great job for the first PR!

PR #16, opened 2025-06-23
Polities harmonization

### LINKED ISSUE #38: Polities and maps in the WHEP project
### Some more specific suggestions for how to start 

**_PART 1_**

**Download data:**

1. Download FAO Country/Region data from FAO. Go to https://www.fao.org/faostat/en/#definitions and choose Country/Region. Download the dataset manually and check it into the Git repository as input data. Name the input dataset after download date (e.g., fao-regions-2025-07-01.csv or similar).
2. Download official M49 data: https://unstats.un.org/unsd/methodology/m49/overview/, rename after download date and check into Git repository too.
3. Also build a file with M49 codes, names and validity years for M49 codes no longer in use. Add this also as an input file in the Git repo.
E.g., based on https://en.wikipedia.org/wiki/ISO_3166-1_numeric#Withdrawn_codes and/or the list at https://unstats.un.org/unsd/methodology/m49/ (under “Recent changes”).

### Build R code to do the following (so that we can change the process later):

**Start building the polity table**

1. Collect all the M49 codes (current and no longer in use) from the M49 datasets.

2. Add the column “polity_code” to uniquely identify each polity in our list. We will let the polity code be a meaningless numeric code 1, 2, 3, …

3. In addition, we can also create a more intuitive “display_code”. For example, it could have a format like {ABC}-{year} where {ABC} is the three-letter ISO code and {year} is the start year for that polity. This display code should also uniquely identify the polity in any given version of the polity table, but it is used for presentation purposes only, and everyone should expect that it can later be changed as needed (e.g., if we change start year or split/merge polities etc).

4. Create a “polity_name” column which initially be based on the names already collected by Alejandra and Eduardo. The exact names are not important for now. We can create shorter/nicer names later if needed.

5. Add the columns “start_year” and “end_year”:
5.1. Where FAO start/end year are available, copy those to the WHEP start/end year.
This is a reasonable choice to begin with because we will use a lot of FAO data. We can change this later if needed (e.g., if the M49 list has earlier or later code validity and if we have substantial other data sources than FAOSTAT).
5.2. For other M49 codes no longer in use, take the official first/last year of validity as WHEP start/end year from that dataset.
5.3. Where still no WHEP Start/End Year is defined, put WHEP Start Year = 1970 (first version of M49 standard) and WHEP End Year = 2025 (this year).

6. Create tests to run some sanity checks/consistency checks:

- All polities have a unique polity_code.
-  There are no two polities with the same M49 code. (But M49 code may be empty.
- All polities have a polity_name, start_year,end_year.
- WHEP Start Year <= WHEP End Year.

**Start building the alias table**

1. Create a table with the following mandatory columns:

1.1. Alias
1.2. Year
1.3. polity_code

2. Start populating the table with data:
2.1. For each polity in the polity list:
2.1.1. For each year from start_year to end_year: add a row in the alias list (name, year, polity code)
2.2. If there are other sources of names (e.g., short names from FAO or ISO datasets) we can add those too in the alias list. We can also later add more aliases manually as needed.
2.3. This can later become quite a long list (imagine we have 300 polities for 150 years, each polity with two different names, that makes 90,000 entries in the alias table). But the upside is that it is very easy to make lookups in it and to check for duplicates.

3. Create tests for sanity checks:
3.1. There are no two rows with the same combination Alias + Year
3.2. All mandatory columns have values
3.3. Each polity_code in the alias table is present in the polity table

**Start building the ISO code table**

1. Based on data from ISO 3166, build the ISO code table(s) in a similar way as the alias table. Most important to start with a country table (ISO 3166-1). Take care so that ISO code + year correctly maps to the right polity for polities that have changed names (e.g., Burma to Myanmar, which will have the same M49 code and the same WHEP Polity Code but two different ISO 3166-1 alpha-3 codes).

**Start collecting map data from relevant places**

1. Suggest to start simple for present-day polities by using a geospatial dataset that has ISO 3166 codes:
1.1. Natural Earth is a good and simple option for current borders.
1.2. Eurostat GISCO data will give country borders back to 2001 (for selected years, which I guess will be sufficient).
1.3. In any case, use the new ISO code table(s) to lookup WHEP Polity Codes.

2. Later, as needed also Natural Earth subnational borders for USA, China, etc.
3. Maybe later also use GAUL data for earlier subnational borders, but let’s avoid it if not needed. Detailed treatment of subnational borders in recent years is not the priority for now!

afuenteshinojosa:
Closes #38.


eduaguilera (commented):
Nice work, @afuenteshinojosa. 
I have added some comments on the code, and I add here some more general comments. Sorry @afuenteshinojosa if you find there are many comments. You are the first team contributor and we are still setting up the procedures!

1. Regarding the routes to files in the input and output folders, I have a few comments:
 - The routes @afuenteshinojosa wrote would not work in my computer, as I have a different folder structure with OneDrive (e.g. I don't have the "Desktop" folder)
 - I would avoid defining the user in each script, as it would imply having to do many changes just to get the code running if more than one script has been modified... That's why I use a "Common_data.R" script to do this kind of setup, but probably there are other better ways to do it... 
 - Any suggestion to address these problems? @lbm364dl 

2. I think the Methods description document (now a Google Doc) should be placed here in the GitHub repository as a markdown document (not sure where, though). In addition, I like having more explanations in the code itself, to avoid having to look at the documentation. As I see it, reading the code with its comments should be enough to understand it, and the documentation should include explanations on the origin of the input files (or even better if a sort of label could be added to these files if they are located in GitHub)

3. It seems that the FT_cleaned.R and FT_and_WHEP.R scripts should be excuted sequentially... For me, it is useful to have a script (or it could be also in the read_me or other documentation document... but in any case it should be accessible and easy to see) showing the order in which scripts have to be run.

4. We should establish conventions regarding file names and script names

inst/scripts/FT_cleaned.R
```diff
@@ -0,0 +1,79 @@
...
+    polity_name = NA,
+    polity_name_full = polity_name_FT_raw,
+    polity_code_full = NA,
+    polity_name_FAO = NA,
+    polity_name_source = "FT",
+    polity_code_source = "FT"
+  )
+
+# Standardize polity names and create final structure
+federico_tena_clean <- federico_tena_clean |>
```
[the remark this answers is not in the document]
lbm364dl: In certain cases I would agree, but here it seems it's just using the variable defined exactly above, so there's no need to create another one, it's OK to just keep writing pipes in the previous statement. If the pipe seems too long, I would create subfunctions for parts that are related.

inst/scripts/FT_and_WHEP.R
```diff
@@ -0,0 +1,127 @@
...
+
+
+# merging the datasets
+polities_whep <- dplyr::bind_rows(whep_final, federico_final)
+
+
+# NA to WHEP
+polities_whep <- polities_whep |>
+  dplyr::mutate(
+    polity_name_source = dplyr::case_when(
```
lbm364dl, line 115: There's the function [`tidyr::replace_na`](https://tidyr.tidyverse.org/reference/replace_na.html) that makes this cleaner:
```r
polities_whep |>
  tidyr::replace_na(list(polity_name_source = "WHEP"))
```
Of course in the list we can add more columns too.

R/polities.R
```diff
@@ -0,0 +1,232 @@
...
+#' - WHEP-polities sheet from the WHEP project database
+#' - Federico-Tena trading polities dataset with temporal coverage
+#'
+#' The resulting dataset provides comprehensive polity information suitable for
+#' historical production and trade analysis with proper source attribution and
+#' standardized naming conventions.
+#'
+#'
+#' @examples
+#' \dontrun{
```
lbm364dl, line 70: I used the `\dontrun{ ... }` in some of my function documentation because they had to load and process a lot of data so it took long to generate the examples. Here you use small input files and the `get_polities()` call is fast, so you can remove the `dontrun` to actually see the examples in the generated doc site.

R/polities.R
```diff
@@ -0,0 +1,232 @@
...
+#' 4. Identifies new records from Federico-Tena not present in WHEP data
+#' 5. Standardizes column structure across both datasets
+#' 6. Combines datasets vertically while maintaining data integrity
+#' 7. Applies final cleaning rules for consistent source attribution
+#'
+#' Data sources:
+#' - WHEP-polities sheet from the WHEP project database
+#' - Federico-Tena trading polities dataset with temporal coverage
+#'
+#' The resulting dataset provides comprehensive polity information suitable for
```
lbm364dl, line 64: This is a bit overlapping with what you already said at start of description section (you could merge both paragraphs in description). Also I'm not sure if the details section is necessary. The user only needs to know how the output data looks. This is something I would probably write in a methodology article.

lbm364dl (changes requested):
You did a great job for the first PR! Please don't feel overwhelmed by all these comments I left. I think it's important to learn good practices soon and try to write clean code (not only code that works). I hope they're helpful. Tell me if something is not clear.

Also, we need to start with some tests! And I would try to add tests first, before trying to do the changes I suggested, so then they can be like your personal check to make sure the code is still working after applying changes.

PR #17, opened 2025-06-26
Build Supply Use Tables

lbm364dl:
It was tough to put everything together, and this part of the model must surely be revisited later for improvements. 
For now I tried to group processes into three groups, each one with its logic as to where to get the data from. Some indirect things I also did:
- Clean many datasets, standardize namings (`_code` and `_name` columns)
- Always use code columns only, and have util functions to get the names back (`add_item_cbs_name`, `add_area_name`, `add_item_prod_name` and their counterparts)
- Create more functions for cleaned datasets coming from @eduaguilera (`get_feed_intake`, `get_primary_production`, `get_primary_residues`). They can be used by anyone who needs to work with those data.

Lastly, I paste here an explanation that I initially wrote in an external task for all the steps followed for building the supply and use tables:

The final goal is to get an input-output matrix (concept from economy https://en.wikipedia.org/wiki/Input%E2%80%93output_model, a.k.a. technical coefficient matrix) for the commodity balance sheet items.
_Why do we need the previous Supply-Use Table (SUT) format?_ In fact we don't, but once we have it, implementing item allocation to get technical coefficients is easier.
_Why do we need to define processes?_ This can also be quite arbitrary. The concept of multi-output process is important because we have to allocate (share) its inputs through each of the outputs. NOTE: If we define in the same process two seemingly unrelated outputs we could get bad results when allocating, but we have to deal with it (see processing items below).
The process name is not that important, since we will be losing it anyway when converting to the input-output square matrix. We will generate process 'names' on the fly.

We will identify a process by a pair (proc_group, proc_cbs_code), where proc_cbs_code is the main item considered in the process (if that makes sense, see cases below) and proc_group is (for now) one of three values, which describe the kind of process (related to how its data will be calculated). These are:
- crop_production:
  - Process ("crop_production", <code_of_main_crop_produced>)
  - Use: Seed of main crop (source: commodity balance sheet use data, Edu's CBS.csv)
  - Supply: 
    - Actual produced crops (source: Edu's Primary_all.csv), 
    - Byproduct residues: Straw, Firewood, other... (source: Edu's Crop_NPPr_NoFallow.csv containing crop residues)
- husbandry:
  - Process ("husbandry", <code_of_animal_involved>)
  - Use: Feed intake, crops used as feed (source: Edu's Feed_intake.csv)
  - Supply: 
    - Actual live animals (in mass) (source: Edu's Primary_all.csv for livestock entries, i.e. those with LU units, then converted to tonnes by 1 LU = 0.65 t)
    - Their animal products (Milk, Offals, Meat, etc) (source: Edu's Primary_all.csv for entries with non-NA Live_anim column)
- processing:
  - Process ("processing", <code_of_item_for_processing_use>)
  - Use: The item used for processing. Note: This is always a single item by how input quantities are stored. This is not ideal but a good starting point. We get this data from Edu's Processing_coefs.csv, we don't really have processes well defined, so there could be more than one process 'grouped' by its input that might give bad results when allocating, but this is the way I thought we could start and have something done.
  - Supply: The multiple outputs for each input item. Again keep in mind they might come from different processes, but for now we assume it is one process. Also comes from Processing_coefs.csv.
    Since we will try to generate some process groupings automatically, we won't really need the process correspondence tables we were trying to fill (items_supply.csv, items_use.csv). We might have to use a similar one for the crop production processes since there are some multioutput crops to treat separately.

This should cover most of the relations between items. There might be others that don't fit in here and we should figure that out later on. I also accept feedback.


R/supply_use.R
```diff
@@ -27,104 +64,197 @@
...
+    .build_livestock_prods_supply(primary_prod, husbandry_items),
+  ) |>
     dplyr::mutate(type = "supply")
+}
 
-  use <- processes |>
-    dplyr::distinct(year, area, proc, item_to_process, value_to_process) |>
-    dplyr::rename(item = item_to_process, value = value_to_process) |>
+.build_livestock_supply <- function(primary_prod, husbandry_items) {
+  k_tonnes_per_livestock_unit <- 0.65
```
eduaguilera, line 206: I think this should not be defined here but in some input file

R/supply_use.R
```diff
@@ -27,104 +64,197 @@
...
+      proc_cbs_code = live_anim_code,
+      item_cbs_code,
+      value = supply
+    ) |>
+    dplyr::mutate(type = "use")
+}
+
+.build_supply_husbandry <- function(husbandry_items, primary_prod) {
+  dplyr::bind_rows(
+    .build_livestock_supply(primary_prod, husbandry_items),
```
eduaguilera, line 199: I think it is only meaningful to have live animals as suppy when they are traded. Otherwise we would be double counting: the same animal would be at the same time a supply item and a product (which is also a supply item)

R/production.R
```diff
@@ -10,11 +10,8 @@
 #' A tibble with the item production data.
 #' It contains the following columns:
 #' - `year`: The year in which the recorded event occurred.
-#' - `area_code`: FAOSTAT internal code for each country from where the data
-#'    is from. Equivalences with ISO 3166-1 numeric can be found in the
-#'    _Area Codes_ CSV from the zip file that can be downloaded from
-#'    [FAOSTAT](https://www.fao.org/faostat/en/#data/FBS). TODO: Think about
-#'    this, would be nice to use ISO3 codes but won't be enough for our periods
+#' - `area_code`: The code of the country where the data is from. For code
```
eduaguilera, line 13: At some point all the references to "area" should be replaced by "polity"

R/livestock.R
eduaguilera, line 1: In the script name I would add something to identify that here we refer specifically to feed intake (as we may end up with many functions/scripts related to livestock). In fact the function only loads and formats the feed intake dataset. At some point we could add the feed intake calculations to this script

R/supply_use.R
eduaguilera, line 1: I have been able to run it and build the supply use table. Very nice! I wonder what would happen if someone had already some of the inputs (such as crop_prod_items, cbs, primary_prod...) in his/her environment (e.g. to use them in other scripts, or even because they had been created as the output of other scripts)... would it be possible to use these objects instead of re-loading/calculating them? 
Also, a related issue is that the function only creates the final output (supply_use) but not the inputs/intermediate objects, so it makes inspecting the process more difficult.

eduaguilera (commented):
Very nice contribution. Overall, the only thing I miss is a guideline explaining the genral purpose of all scripts and inputs and outputs, and in what order to run them (or maybe I missed that?)

inst/extdata/input/raw/crop_production_items.csv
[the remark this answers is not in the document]
lbm364dl: This is not the whole list of of products, it's a particular subset that I happened to need in this calculation. I agree there could be some new column in the correspondence file which can be used to filter, but I'm not sure if it's too arbitrary to start adding seemingly unrelated things there. I can take a look.

R/livestock.R
lbm364dl: The name of the file is on purpose, exactly because right now it only has feed intake, but later on someone could add another function related to livestock, that's why I chose a more general name. If later on we decide that there should be a single file only for feed intake, then we can change it. When we are not sure in advance, general names should be better to start.

R/supply_use.R
lbm364dl: > I think this should not be defined here but in some input file

Thanks for the comment. I had a look at https://r-pkgs.org/data.html which explains how to deal with data in R packages. For constants (in general internal package data) it suggests using `usethis::use_data` with `internal = TRUE`, so I just added a commit changing the way I define the constant. Whenever we need constants from now on, I would add them to `data-raw/constants.R`. See that file for details.

R/supply_use.R
lbm364dl: I understand the issue. I'm still not sure where we want to end up with this table, so I think it's better to raise this issue in a next PR where I integrate the input output matrix with trade.

inst/extdata/input/raw/crop_production_items.csv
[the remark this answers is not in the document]
lbm364dl: Ok, so I got rid of these code-only CSV files and added a new column `item_type` in both `items_prod.csv` and `items_cbs.csv`, and used this with a specific value to filter the items I want.
I spent no time trying to come up with a meaningful classification for the rest of the items, I just wrote a value of `other`. I don't really like trying to do many things ahead of time because later on they might prove useless and we would have to think them again. That's why I prefer leaving the `other` value and letting the next person who needs that to decide. Is that OK?

R/supply_use.R
eduaguilera: Ok! This is not urgent, we can revisit it later on

eduaguilera (approved):

PR #65, opened 2025-12-15
Gapfilling and LMDI decomposition

jinfama:


R/gapfilling.R
```diff
@@ -178,39 +171,1041 @@ proxy_fill <- function(df, var, proxy_var, time_index, ...) {
...
+fill_growth <- function(
+  data,
+  value_col,
+  proxy_col,
+  time_col = "year",
+  group_by = NULL,
+  max_gap = Inf,
+  max_gap_linear = 3,
+  fill_scope = NULL,
+  smooth_window = NULL,
```
eduaguilera: This is very nice. As I understand it, it refers to smoothing the proxy variable for the calculation of growth rates. I wonder if it would also be possible to add a similar smoothing to the gapfilled variable, as I've done now for `fill_linear()` (or maybe it exists and I've missed it?). Then, instead of `smooth_window`, we may have something like `value_smooth_window` and `proxy_smooth_window`

R/calculate_lmdi.R
```diff
@@ -0,0 +1,1056 @@
...
+#' )
+calculate_lmdi <- function(
+  data,
+  identity,
+  identity_labels = NULL,
+  time_var = year,
+  periods = NULL,
+  periods_2 = NULL,
+  .by = NULL,
+  rolling_mean = 1,
```
lbm364dl: Using `rolling_mean` removes data from endpoints, e.g. if it's 2, it removes first and last years because not enough space to calculate a 3 year rolling mean. Is this intended? Should we keep them, calculating mean for the years actually available inside the window?

PR #66, opened 2026-01-26
Harmonization functions

justin-morgan-csic:
Harmonization function which takes data (dataframe with (at least) columns year, value, item_code_harm, and type) already joined with a harmonization table. The function first harmonizes simple data and checks if there are any complex harmonizations. If there aren't, it returns the simple data, and if there are, it checks if all of the 1:N groups are complete and finds their range of years. If the groups are incomplete, it returns an error. If complete, it takes the shares of the N products for the benchmark years and interpolates the shares between them. Then, it left_joins the df with shares to the 1:N data, calculates the estimated values, then joins the simple dataframe and does one last summarize(.by) harmonization.


R/harmonize.R
```diff
@@ -0,0 +1,301 @@
...
+#'   - `year`: Numeric, year of observation.
+#'   - `value`: Numeric, summed value of observation.
+#'   - and any additional grouping columns.
+#' @export
+#' @examples
+#' # TODO: add example
+#' # harmonize_interpolate(data, country)
+harmonize_interpolate <- function(data, ...) {
+  grouping_cols <- rlang::enquos(...)
+  grouping_names <- .group_names(grouping_cols)
```
lbm364dl:
```suggestion
```
Unused

R/harmonize.R
```diff
@@ -0,0 +1,301 @@
...
+    dplyr::left_join(
+      year_bounds,
+      by = c("items", grouping_names),
+      relationship = "many-to-many"
+    ) |>
+    dplyr::mutate(
+      year = purrr::map2(start_year, end_year, seq)
+    ) |>
+    tidyr::unnest(year) |>
+    dplyr::select(-end_year, -start_year) |>
```
lbm364dl: Useful selects when debugging, but unnecessary once you have the function working. It's better to remove these two selects and leave only the last one.
```suggestion
```

R/harmonize.R
```diff
@@ -0,0 +1,301 @@
...
+#'   the full year range and applies them to split values.
+#' @param data A data frame containing at least columns:
+#'   - `items`: String, original item name.
+#'   - `item_code_harm`: Numeric, code for harmonized item.
+#'   - `year`: Numeric, year of observation.
+#'   - `value`: Numeric, value of observation.
+#'   - `type`: String, harmonization type (only "Simple" rows are used).
+#' @param ... Additional grouping columns provided as bare names.
+#' @return A tibble with columns:
+#'   - `item_code_harm`: Numeric, code for harmonized item.
```
lbm364dl: I see this name is different
```suggestion
#'   - `item_code`: Numeric, code for harmonized item.
```

lbm364dl (changes requested):
Great first PR! You'll notice I added some commits myself, you're also welcome to check those (probably enough with [this last one](https://github.com/eduaguilera/whep/pull/66/commits/951e404cb2beca85df149b424088e4b45167f5e7). I tried to add here some more general comments.

R/harmonize.R
```diff
@@ -0,0 +1,301 @@
...
+    )
+  group_year_presence <- .find_group_year_presence(
+    df_simple,
+    data_groups,
+    harm_groups,
+    grouping_cols
+  )
+  .check_group_year_presence(harm_groups, group_year_presence, grouping_cols)
+  year_bounds <- .find_year_bounds(group_year_presence, data, grouping_cols)
+  print(year_bounds)
```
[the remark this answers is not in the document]
justin-morgan-csic: oke

Issue #114, opened 2026-06-29
Wire the labour footprint extension (Global Labour Database)


lbm364dl:
Split out of #107 (follow-up to #100, which delivered the **GHG** extension and the generic `build_footprint()` / `align_extension()` + co-product `method = "value"` allocation machinery).

Same pattern as the land/GHG extensions: a `build_labour_extension()` returning `(year, area_code, item_cbs_code, impact_u, method_labour)`, fed to `build_footprint()`, validated with the #98 trustworthiness layer (conservation, uncertainty, method comparison). PR-sized.

### Labour (socio-economic)

- **Global Labour Database** (Juan Infante-Amate): hours·ha⁻¹ by crop × country, disaggregated (total / poverty tiers / child / forced), per `~/Edu/Global` `R/footprints.r:64–84` (`labour_synthesis_items.csv`). Output labour-hours per `(area_code, item_cbs_code)`.
- **Needs external data** (likely obtainable in-house).

---

Suggested order across the split-out pressures (value × tractability): **biodiversity** → **water** (#107) → **nitrogen** → **labour**.

lbm364dl:
### Triage 2026-08-27 — STILL VALID, unstarted, no PR

Checked in contrast to #107, which does have a (stale) PR:

- No `build_labour_extension()` anywhere on `main`; the extension set is
  `crop_land`, `grassland_land`, `livestock_ghg`, `energy_co2`, `crop_soil_n2o`,
  `n_exceedance`.
- **No open PR** targeting this issue, and no remote branch matching `labour`
  (`repos/eduaguilera/whep/branches` has none). So unlike #107 this is not
  blocked on reviving stale work — it is simply not begun.
- No duplicate: searching `labour` returns only #855 (a `regions_full` data
  defect in `region_labour_mech`, unrelated to the extension) and the sibling
  extension issues.

**Still blocked on data.** The Global Labour Database synthesis CSV
(`labour_synthesis_items.csv`) is not in the repo and there is no
`gld-labour-*` entry in `inst/extdata/whep_inputs.csv`; the data has to be
obtained and pinned before any code is worth writing. That is the first
deliverable, not the extension.

Worth knowing before starting: whatever is wired will inherit the trace gap in
#119 (measured at -15% blue / -42% green water on the same machinery), so a
conservation number for labour will not close until that is resolved.

**Labels**: `enhancement, priority:medium, area:footprint, footprint-extension,
[...]

Issue #113, opened 2026-06-29
Wire the nitrogen footprint extension (BNF / N surplus)


lbm364dl:
Split out of #107 (follow-up to #100, which delivered the **GHG** extension and the generic `build_footprint()` / `align_extension()` + co-product `method = "value"` allocation machinery).

Same pattern as the land/GHG extensions: a `build_nitrogen_extension()` returning `(year, area_code, item_cbs_code, impact_u, method_nitrogen)`, fed to `build_footprint()`, validated with the #98 trustworthiness layer (conservation, uncertainty, method comparison). PR-sized.

### Nitrogen

- Two candidate metrics: (a) **biological N fixation** as a pressure (`calculate_crop_bnf()` exists; needs `item_prod_code`→`item_cbs_code` mapping + the NPP inputs); (b) **N surplus** (inputs − removal). Both can reuse the soil-N input assembly already built for `build_crop_soil_n2o_extension()` (synthetic F_SN, manure F_ON, residue F_CR, plus BNF).
- Mostly in-repo; surplus needs crop N-removal (production × N content).

### Related / further candidates (lower priority)

Moved here from the original #107 backlog (phosphorus naturally mirrors nitrogen):

- **Energy/fertiliser CO₂** — tracked separately in #105.
- **Phosphorus** application/surplus (mirrors nitrogen; FABIO `12_3`/`12_4`).
- **Land-use-change CO₂** (deforestation; BLUE model / LUH2) — major for beef/soy/palm; heavy gridded data.
- **Eutrophication** characterization (N/P surplus → PO₄/NO₃-eq) — neither peer codebase does this yet.

---

Suggested order across the split-out pressures (value × tractability): **biodiversity** → **water** (#107) → **nitrogen** → **labour**.

lbm364dl:
Closing: this issue has been overtaken by the SJOS-N work on `main`, and its remaining half is split out to #664.

**The N surplus half is delivered.** `main` now carries a complete, tested, per-crop nitrogen footprint path:

```
build_nitrogen_balance()                          # gridded N balance
  -> calculate_n_surplus()                        # per-crop soil-surface surplus (R/n_surplus.R)
  -> build_n_boundary_exceedance(resolution = "country")
  -> build_n_exceedance_extension()               # year, area_code, item_cbs_code, impact_u, method_n_exceedance
  -> build_footprint()                            # the FABIO IO model
  -> build_sjos_n_footprint()                     # consumption side, domestic vs traded
```

[...]

Issue #112, opened 2026-06-29
Wire the biodiversity footprint extension (potential species loss / PDF)


lbm364dl:
Split out of #107 (follow-up to #100, which delivered the **GHG** extension and the generic `build_footprint()` / `align_extension()` + co-product `method = "value"` allocation machinery).

Same pattern as the land/GHG extensions: a `build_biodiversity_extension()` returning `(year, area_code, item_cbs_code, impact_u, method_biodiversity)`, fed to `build_footprint()`, validated with the #98 trustworthiness layer (conservation, uncertainty, method comparison). PR-sized.

### Biodiversity (potential species loss / PDF)

- Lowest marginal effort because it **layers on the existing land footprint**: apply **Chaudhary et al. (2015)** characterization factors to cropland/pastureland land use, per FABIO `R/15_fabio_footprints.R` (global + regional variants).
- **Needs external data**: the published CF table (small) registered as a pin.

---

Suggested order across the split-out pressures (value × tractability): **biodiversity** → **water** (#107) → **nitrogen** → **labour**.

lbm364dl:
Too broad for WHEP at least for now

PR #116, opened 2026-06-29
feat(water): wire the water footprint extension (blue + green) (#107)

### LINKED ISSUE #107: Wire the water footprint extension (blue + green)
Follow-up to #100, which delivered the **GHG** extension (livestock CH₄/N₂O + crop/soil N₂O) and the generic `build_footprint()` / `align_extension()` + co-product `method = "value"` allocation machinery. The original #100 framed a broader set of pressures (GHG → water → nitrogen → biodiversity); this issue now tracks **water only**. The other pressures have been split into their own issues:

- Biodiversity → #112
- Nitrogen → #113
- Labour → #114

This is the same pattern as the land/GHG extensions: a `build_water_extension()` returning `(year, area_code, item_cbs_code, impact_u, method_water)`, fed to `build_footprint()`, validated with the #98 trustworthiness layer (conservation, uncertainty, method comparison). PR-sized.

### Water (blue + green)

- Port FABIO's approach (`/home/usuario/fabio/R/12_1_ext_land_mass_water.R`): **Water Footprint Network / Mekonnen & Hoekstra** m³·t⁻¹ per crop × country (4TU.nl), split into green + blue (irrigation + capillary rise); livestock blue water m³/head; grazing green water m³/ha.
- **Needs external data**: register the WFN coefficient table as a pin (`whep_inputs.csv`). Output m³ per `(area_code, item_cbs_code)`.
- One builder with a `component = c("blue", "green")` selector (default `"blue"`): blue = crop blue + livestock blue; green = crop green + grazing green.

Suggested order across the split-out pressures (value × tractability): **biodiversity** (#112) → **water** (this issue) → **nitrogen** (#113) → **labour** (#114).

lbm364dl:
Closes #107.

Wires the **water footprint extension** (blue + green), the water analogue of the GHG and soil-N2O extensions. `build_water_extension()` returns a tidy `(year, area_code, item_cbs_code, impact_u, method_water)` extension in m³, fed to `build_footprint()` the same way as the land/GHG ones.

### Design
- `build_water_extension(component = c("blue", "green"), data = list(), example = FALSE)`. The footprint engine traces one `impact_u` per run, so `component` (default `"blue"`) selects the water type and is recorded in `method_water` (`"WFN_blue"` / `"WFN_green"`).
- **blue** = crop irrigation (`wfb_i + wfb_cr` per t) × crop production + livestock blue water (m³/head) × head counts.
- **green** = crop rainfed (`wfg` per t) × production + grazing green water (m³/ha) × grazed grassland area (item_cbs 3000/3002).
- Crop intensities join whep's **own** production (`get_primary_production()` "tonnes" rows) by FAO crop and country-year, so the footprint tracks whep production. Crop coverage spans 1990–2019.

### Coefficient data (3 pins, all uploaded to the saco board and wired)
| Pin | Source |
|---|---|
| `wfn-water-crop` | Mialyk et al. (2024), 4TU dataset (doi:10.4121/7b45bcc6…); 244,058 rows, 175 crops × 178 countries |
| `water-livestock-blue` | Chapagain & Hoekstra (2003), Report 13 Tables 3.8–3.9 (drinking + service, adult, mixed system); 8 whep live-animal species |
| `water-grazing-green` | Green ET of managed grassland from the WHEP LPJmL run (band "rainfed grassland", 2000–2009 mean), country-aggregated via `country_grid.parquet` |

No coefficient values are hardcoded; all are read via `whep_read_file()`. `data-raw/water_footprints.R` documents the full reproducible derivation of each.

### Grazing basis
Grazing green water charges the **full** managed-grassland ET, consistent with `build_grassland_land_extension(grassland_metric = "occupation")` — the same occupation-vs-active-grazing choice already settled for the land extension. Validation: total grassland area 3,253 Mha ≈ FAO permanent pastures (~3,300 Mha); global total ~13,700 km³/yr, within the published 8,258–12,960 km³/yr range for total grazing-land ET.

### Verification
- 11 new tests in `tests/testthat/test_water_extension.R`; sibling extension/footprint tests still pass (no regression from the shared `utils.R`/`toy_examples.R` edits).
- `air format` clean · `rcmdcheck` 0 errors / 0 warnings / 0 notes · lint clean · `_pkgdown.yml` updated.
- End-to-end on the real pins: blue (crop + livestock) and green (crop + grazing) both reproduce hand-computed values.

### Known limitations (follow-ups, not blockers)
- Livestock blue water covers 8 species with C&H 2003 data; buffalo/camels/etc. get no direct blue water (would need proxies).
- Crop coefficients span 1990–2019 (no extrapolation yet).

The sibling pressures from the original #107 were split into #112 (biodiversity), #113 (nitrogen), #114 (labour).


lbm364dl:
## Real-data validation (year 2010)

Ran the extension on the full `get_primary_production()` and traced it through `build_footprint()` (2010 IO model).

**Production-side extension totals** — consistent with published global water footprints:

| Component | Total | Split | Benchmark |
|---|---|---|---|
| Blue | **775 km³** | crop irrigation 726 · livestock 49 | M&H 2011 global crop blue ≈ 900 km³ |
| Green | **18,169 km³** | grazing 13,160 · crop rainfed 5,008 | M&H 2011 global crop green ≈ 5,800 km³; grazing within the 8,258–12,960 km³/yr range for total grazing-land ET |

Coverage: 187 (blue) / 169 (green) countries.

**Footprint trace** runs cleanly (14.8M blue / 13.8M green non-zero flows). Top consumption-based countries are sensible:
- blue: India, China, Pakistan, USA, Iran, Egypt (irrigation-heavy)
- green: Brazil, China, USA, Colombia, India, Kazakhstan

**Caveat — consumption-side conservation gap.** The trace conserves only ~85% of blue and ~58% of green pressure (cons 658 vs prod 775 km³; 10,568 vs 18,169 km³). `conserve_extensions = TRUE` does not change this. The IO build logs imbalances (negative outputs, capped `A` columns, near-zero-output sectors dropped). This is **generic footprint-machinery behaviour** — the IO trace is extension-agnostic — but grassland-routed pressure (grazing, which dominates green) is hit hardest, because grassland reaches final demand only via the livestock chain. Worth a separate look at IO balance / the `conserve_extensions` path; not specific to this extension and not a blocker for wiring it.

lbm364dl:
## Rebased onto main and brought up to the current contract

Was 784 commits behind; the four original commits replay cleanly and three new
ones follow. Full suite `FAIL 0 | PASS 6646`, lint clean, `rcmdcheck` 0/0/0,
`_pkgdown.yml` complete.

### 1. Keyed on `polity_area_code` (the reason it could not merge)

Output now carries the four reporting-polity columns and is keyed on
`polity_area_code`, matching `build_grassland_land_extension()` and
`build_livestock_ghg_extension()`. As written in June it emitted legacy FAOSTAT
codes, so joining it against the sibling extensions or the IO model would have
dropped rows silently.

Coefficients are still joined on the **legacy** `area_code` that both they and
`get_primary_production()` carry; only the resulting cubic metres are collapsed
onto the polity. The ordering matters: the coefficients are intensities
(m³/t, m³/head, m³/ha), which cannot be averaged across legacy areas sharing a
polity without a weight — and silently picking one would be wrong. Volumes add,
so collapsing after the multiplication needs no weight. There is a test with two
legacy areas on one polity and deliberately unequal intensities: 104,000 m³,
against the 252,500 a mean intensity would have given.

### 2. The grazing pin derivation goes through the reader

`data-raw` now builds `water-grazing-green` via
`read_lpjml_hydrology("cft_consump_water_g")` filtered on `band_name`, instead
of hand-rolled `ncvar_get` slicing. That removes a hardcoded band index 14, a
hardcoded absolute run path (which `CLAUDE.md` forbids), and by-hand annual
time-index arithmetic. Verified as an equivalence, not by eye: against the same
5.9.7 run the pin was built from, it reproduces the documented totals exactly
(13,683 vs ~13,700 km³/yr; 3,253 Mha grassland). This builds on #681.

### 3. The pin stays 5.9.7-derived — 6.1.1 is broken (#710)

I regenerated from the 6.1.1 production run as planned. The global total drops
13,683 → 3,923 km³/yr, and that is a bug, not physics. Band 14 is **rainfed**
grassland, which cannot be irrigated:

| | 5.9.7 | 6.1.1 |
|---|---|---|
| green | 434.9 mm | 134.2 mm |
| **blue** (must be 0) | **0.0 mm** | **382.2 mm** |
| green + blue | 435.0 | 516.4 |
| transpiration | 344.5 | 371.1 |
| cells with green < transp | 0 % | **99.1 %** |

The total is fine and tracks the run's ET increase; only the green/blue
partition is wrong. Traced to `stand->frac_g` in `src/soil/waterbalance.c`,
which must stay 1.0 on a rainfed stand and has drifted to ~0.26. Filed as
**#710** — it also affects `build_water_balance(blue_green = "cft_native")`, the
default method, on any 6.1.1 run. Regenerating is one command once a run with a
correct split exists.

### Still open for @eduaguilera

The **occupation basis** question from #681 is unchanged: grazing is charged the
full managed-grassland ET, consistent with
`build_grassland_land_extension(grassland_metric = "occupation")`. The Schyns
et al. (2019) alternative is several times smaller and drives ~72% of the green
total. Proceeding on the occupation basis for consistency with land; your call
to confirm.

### Not yet done

End-to-end re-verification against June's real-data totals (blue 775, green
18,169 km³/yr for 2010). World totals should hold and per-country values shift
with the re-keying; that run needs the full `get_primary_production()` (~26 min)
and I will post the comparison separately.

lbm364dl:
## All 10 checks green — and the last blocker is cleared

### The grazing pin is now annual, and its 6.1.1 split is corrected

Regenerated from the 6.1.1 production run: **1901–2023, 21,365 rows, 175
countries**, version `20260810T142528Z-b4612`, replacing one 2000–2009 mean that
was being applied to every year since 1850.

Corrected green is `green + blue`, because band 14 is *rainfed* grassland and
cannot be irrigated — its blue is the #710 misassignment. That substitution is
valid because the fix (lbm364dl/LPJmL#3, merged) is a **pure repartition**:
rebuilding the parent commit and running both binaries on one config moves total
consumptive water by **+0.0011%**, with `pre(green+blue) == post(green)` cell by
cell. **No 12-hour rerun was needed.**

| basis | global grazing green water |
|---|---|
| 6.1.1 as-is (the bug) | 3,923 km³/yr |
| **6.1.1 corrected, annual** | **15,735 km³/yr** |
| 5.9.7 (previous pin) | 13,683 km³/yr |

+15%, against the +13.5% rise in global ET that run's README documents. Two
independent numbers agreeing is what made the existing run trustworthy here.

### The join_audit gate is cleared by *removing* an entry

Going annual made the coefficient join key on `(year, area_code)`, so
`.grazing_green_water` **leaves** the territorial-join baseline. The cap stays at
58 and the "never raise it for a new year-blind read" rule is intact — the list
shrank rather than grew, which is what #669 asks for.

### Cross-check worth reading

172 of 175 countries match the decade computation to 0.01%. The three that don't
have grassland in only **1, 4 and 10** of those years, and the old decade mean
diluted their intensity by averaging zero-grassland years into the millimetres —
understating them by 10× and 2.5×. The annual values don't.

### Full state

Rebased onto current main, 7 commits, `FAIL 0 | PASS 6916`, lint clean,
`rcmdcheck` 0/0/0, all 10 CI checks green.

---

## Not merging — one decision is yours, @eduaguilera

At **15,735 km³/yr** the occupation-basis total sits **above** the published
8,258–12,960 km³/yr range for total grazing-land ET. The 5.9.7 value (13,683)
already exceeded it, and this PR's original text claiming it was "within" that
range was **not accurate** — corrected here.

The question is unchanged from #681: grazing is charged the **full**
managed-grassland ET, consistent with
`build_grassland_land_extension(grassland_metric = "occupation")`. Schyns et al.
(2019) instead restrict to the grazed area at the necessary livestock density and
get several times less. That choice drives ~72% of the green footprint.

It is a methodology call, not an engineering one, so this sits ready rather than
merged.

### Remaining follow-ups (neither blocking)

- A global rerun with the LPJmL fix compiled in, to confirm whole-run effects and
  pick up the other bands. Off the critical path now.
- End-to-end footprint re-verification against June's totals (blue 775, green
  18,169 km³/yr for 2010). World totals should hold; per-country values shift
  with the polity re-keying, and green rises with the corrected pin.

lbm364dl:
## Real-data verification, and a cross-check against an independent model

Two things finished since the last comment: the end-to-end run on real data, and
a test of whether LPJmL could supply the crop water instead of Mialyk.

### 1. End-to-end, year 2010, against the June baseline

| | June | now | change |
|---|---|---|---|
| **blue** | 775 | **779.8** km³/yr | +0.6% |
| **green** | 18,169 | **20,927.6** km³/yr | +15.2% |
| — crop rainfed | 5,008 | **5,008.9** | **+0.02%** |
| — grazing | 13,160 | **15,918.7** | +21% |
| countries (blue / green) | 187 / 169 | 195 / 175 | — |

**The crop line is the one that matters.** Crop water never touches the grazing
pin, so it isolates the polity re-keying: the global total is preserved to 0.02%
while the country partition changed underneath. That is the signature of a
correct re-keying, and it is the failure CI cannot see.

Blue is effectively unchanged, as it should be — it does not read the grazing
pin. Top consumers are the same set as June (India, China, USA, Pakistan, Egypt,
Iran).

Green rises entirely through grazing, and the +21% decomposes:
- **+15%** from the corrected pin (#710)
- **~+5%** because 2010 was wetter than the 2000s mean (per-country coefficient
  ratio 2010 / 2000–2009 = **1.068**) — information the old climatological pin
  discarded by construction, and the reason for going annual

**Independent cross-check:** the extension's grazing total (15,918.7 km³) uses
**FAOSTAT/LUH2** grassland area, while the pin's own global figure (15,735 km³)
came from **LPJmL CFTfrac** area. Two unrelated land-area sources agreeing to
**1.2%**.

### 2. Could LPJmL supply the crop water too? Tested — no, but it earns its keep

WHEP's LPJmL run also simulates per-crop water, so we asked whether the crop
coefficients could come in-house instead of from Mialyk et al., which would give
one coherent water cycle across the whole extension. The crosswalk to do it
already exists (`inst/extdata/cft_mapping.csv`, used forward by
`.aggregate_to_cft()` to build LPJmL's own input).

Applying LPJmL **intensities** to **WHEP** crop areas, 2010:

| | LPJmL | Mialyk | ratio |
|---|---|---|---|
| all crops | 5,960 km³ | 4,961 km³ | 1.20 |
| **the 12 crops LPJmL models** | **3,259 km³** | **3,597 km³** | **0.91** |
| generic `others` band (103 crops) | 2,701 km³ | 1,364 km³ | **1.98** |

**Decision: keep Mialyk.** Two reasons.

- The generic band is badly wrong. `others` carries 103 crops and 27% of crop
  water, and comes out at ~2×: coffee 1.84, cotton 1.75, oil palm 1.71, rubber
  1.61, cocoa 1.60, vegetables **3.14**. LPJmL grows one generic annual crop
  there, which does not represent perennials or short-season vegetables.
- Global agreement hides local disagreement. Even for the 12 modelled crops,
  only **39.7%** of country-crop pairs land within ±25% of Mialyk (median ratio
  1.10, outliers to 70×). The totals agree because errors cancel.

**But the result is worth keeping.** For the crops LPJmL does model, two models
sharing no assumptions agree within **9%** globally with a country-crop rank
correlation of **0.972** — per group: temperate cereals 0.96, sugarcane 1.03,
pulses 1.05, groundnut 1.00, temperate roots 1.03, rapeseed 1.10, rice 1.14.
That is independent corroboration of the Mialyk-based crop footprint we did not
have before, and the country-crops where the two diverge are a ready-made list
of things worth a human look. Better suited to the validation harness than to
the extension pipeline; happy to file it separately if wanted.

Consequence worth stating plainly: the green footprint mixes two hydrologies —
grazing (15,919 km³) from WHEP's LPJmL run, crops (5,009 km³) from Mialyk's crop
model. About three-quarters / one-quarter. Normal practice in footprint work, but
now measured rather than assumed.

---

## Over to you, @eduaguilera — one decision

Everything verifiable is verified: all 10 checks green, full suite
`FAIL 0 | PASS 6916`, `rcmdcheck` 0/0/0, real-data totals reproduced.

The open question is **methodological, not technical**, which is why this is not
merged:

**Grazing is charged the full managed-grassland ET**, consistent with
`build_grassland_land_extension(grassland_metric = "occupation")`. The
alternative — Schyns et al. (2019), restricting to the grazed area at the
necessary livestock density — is several times smaller.

That choice drives **~76% of the green footprint** (15,919 of 20,928 km³ in
2010). And at 15,735 km³/yr the occupation-basis global total sits **above** the
published 8,258–12,960 km³/yr range for total grazing-land ET. The previous
5.9.7-based value (13,683) already exceeded it — this PR's original text claiming
it was "within" that range was **not accurate**, and is corrected here.

The argument for the current choice is consistency: the land extension already
made the same call, so land and water charge grazing on the same basis. The
argument against is that the resulting number is outside the published range for
the quantity it claims to estimate.

Related open items, neither blocking this PR: **#710** (LPJmL green/blue defect —
fixed upstream in lbm364dl/LPJmL#3, stays open until a production run exists with
the fix) and **#737** (`build_water_balance`'s default `cft_native` split is
wrong on 6.1.1 runs; latent, nothing downstream reads it).

lbm364dl (commented):
*Review by Claude.* This PR is `CONFLICTING` and last touched 2026-08-10, so it needs a rebase before anything else — but **I independently verified its most load-bearing claim on the LPJmL runs on this machine, and it holds.** That claim is also worth more than the PR gives it credit for.

## The "pure repartition" justification is correct, and I measured it

`data-raw/water_footprints.R` derives `water-grazing-green` from `cft_consump_water_g`, band `"rainfed grassland"`, and corrects for the LPJmL 6.x defect by taking **green + blue**, on the grounds that the upstream fix is a pure repartition and therefore no rerun is needed. That is the kind of claim that quietly bakes a wrong coefficient into a pin if it is off, so I tested it against the two 6.1.1 runs directly, band `"rainfed grassland"`, every cell, 2000–2009 (the pin's own window):

```
                 green         blue    green+blue
defective   6,197,905   16,277,853    22,475,758      GIT_hash 61119f2d
fixed      22,475,395          416    22,475,811      GIT_hash 552f1d68

defective(green+blue) vs fixed(green):   +0.0016%
total consumptive water, run vs run:     -0.0002%
```

So `pre(green + blue) == post(green)` holds to 0.0016% over the whole band and decade, and total consumptive water is unchanged to 0.0002% — consistent with the +0.0011% quoted in the comment. `git merge-base --is-ancestor` also confirms the fix commit `bb25d2a4` is in `552f1d68` and not in `61119f2d`, so these really are the two sides. The reasoning is sound and the pin does not need regenerating from a rerun.

**What the PR undersells: the correction is worth a factor of 3.63.** Naively taking `green` alone from the production run gives 6,197,905 against a corrected 22,475,395 — grazing green water would have been **3.6× too small**, silently, in a shipped coefficient pin. That belongs in the PR description and in NEWS, not only in a `data-raw` comment, because it is the number that tells a future reader why the `green + blue` line must not be "simplified".

## Cross-reference #857, which is about the same defect

#857 adds a runtime guard for exactly this: it warns when `cft_native` receives rainfed bands carrying blue water. Its remedies are "use a run built with the fix" or "use `irrig_share`". This PR establishes a third and cheaper one — `green + blue` from the existing run — and, importantly, **that remedy is only valid where the bands are rainfed-only**, as grazing is. `build_water_balance()` cannot use it, because summing green and blue destroys the split it exists to compute. Worth saying so in both places: right now the verification that makes the repartition trustworthy lives only in this file's comment, and #857 does not know about it.

## Rebase: regenerate, do not merge

`git merge-tree` gives three conflicts:

```
NEWS.md
inst/extdata/whep_inputs.csv
data/whep_inputs.rda
```

The last two are the input registry and its serialisation. **Do not resolve `data/whep_inputs.rda` as a merge** — it is binary, git will happily produce a corrupt-but-plausible blob, and the `.rda` and the `.csv` will disagree in a way nothing checks. Resolve `whep_inputs.csv` by hand (both sides add rows; take both), then re-run the `whep_inputs` `data-raw` script to rebuild the `.rda` from the merged CSV, and commit the regenerated artefact. NEWS.md is an ordinary text conflict.

## What I could not verify

- **The three pins.** `wfn-water-crop`, `water-livestock-blue` and `water-grazing-green` live on the saco board and I have no credentials for it, so the 244,058-row Mialyk table and the C&H 2003 livestock coefficients are unchecked. Only the grazing derivation is reproducible from local data, which is why that is the one I measured.
- **The 3,253 Mha grassland total and the ~13,700 km³/yr global figure.** Both need the cftfrac × cell-area weighting and the country grid, which I did not rebuild. The band-level mm above is upstream of both, so a discrepancy there would be in the weighting, not in the ET.

## On the grazing basis

Charging the full managed-grassland ET, consistent with `grassland_metric = "occupation"`, is the right call for coherence with the land extension — and citing Schyns' 2,191 km³/yr as the *restricted* comparison rather than as a contradiction is the honest way to present a global total at the top of the published range. Keep that paragraph; it is what stops the number reading as an error.

eduaguilera:
Overlap notice from PR #1058 — correct the gridded soil carbon-nitrogen chain. We both change `R/lpjml_hydrology.R`, and there are two collisions worth settling before either merges.

**1. A fifth LPJmL pin, from a different run.** WHEP currently carries four LPJmL-derived pins — `lpjml-grass-availability`, `lpjml-grass-productivity`, `lpjml-grass-natural-net-c`, `lpjml-soc-hydrology` — and CLAUDE.md is explicit that they must be regenerated **together, from one run**, because refreshing some leaves WHEP mixing two LPJmL versions across its feed, soil-carbon and water chains at once, invisibly, since every pin still loads with the right schema. #1058 repoints all four at the 1750-2023 `socn_diag` run. If this PR adds a fifth from a different run, that invariant breaks in exactly the way the rule was written to prevent. Worth confirming which run yours comes from, and whether it belongs in `regenerate_whep_lpjml_pins()` rather than beside it.

**2. `stand_frac` versus `cftfrac`.** #1058 names the per-CFT stand fraction `stand_frac` throughout the hydrology reader; this PR uses `cftfrac`. Same quantity, two names, and whichever merges second will silently read a column that is not there or, worse, join on a name that exists for a different reason. Mine also weights per-CFT cubes by stand area before summing them to a cell (`fix(water): weight per-CFT cubes by stand area before summing them to a cell`) — if the water extension sums the same cubes unweighted, the two will disagree on any cell with more than one crop.

**3. Smaller, but real.** #1058 adds `read_lpjml_hydrology()` variables for per-crop monthly applied irrigation (`cft_airrig_month`) and refuses a cube whose bands do not match the stands, and it reads each run's start year from the file rather than assuming 1901. If the water extension assumes 1901 anywhere, that is worth checking against a run that does not start there.

Files in common: `R/lpjml_hydrology.R`, `R/utils.R`, `R/toy_examples.R`.

No action needed from me right now — I would rather agree the column name and the pin provenance with you than have the second merge discover them.

PR #121, opened 2026-07-02
Defer FABIO Rest-of-World aggregation out of the base datasets (per-country grain)

### LINKED ISSUE #120: Defer FABIO Rest-of-World aggregation out of the base production dataset (keep per-country grain)
## Summary

The base production dataset (`build_primary_production()`, and the raw-input path generally) **bakes FABIO's "Rest of World" aggregation into the atomic data**. Identifiable, real countries that FABIO does not enumerate are silently summed into a single `area_code = 999` "Rest of World" row instead of being kept individually. This is a FABIO **matrix-compatibility** decision applied at the wrong layer — the base table — where it destroys per-country information that cannot be recovered downstream.

Proposal: keep the base dataset at per-country (FAOSTAT/ISO3) grain and move the FABIO RoW collapse into an explicit downstream step at the IO/matrix boundary.

## Evidence

Inspecting a real build (`build_primary_production()`, 2015–2016):

- Real sovereign countries are **absent as individual rows** — folded into `999`: **Syria (SYR)**, **North Macedonia (MKD)**, **Palestine (PSE)**, **Equatorial Guinea (GNQ)**, **Eswatini (SWZ)**, plus micro-states (Andorra, Monaco, San Marino, Liechtenstein, Greenland, Faroe Islands, …).
- The `999` "Rest of World" row carries substantial summed production (for the 2015–2016 slice alone: ≈44.7 M tonnes, ≈97.9 M heads, ≈252 M slaughtered_heads, ≈9.2 M ha). A meaningful share is identifiable real countries (Syria and North Macedonia are non-trivial producers), not just uninhabited territories.
- Sudan / South Sudan (`276`/`277`) are likewise collapsed into `206` "Sudan"; neither appears individually.
- In the output, `area_code == polity_area_code` for every row — because the raw FAOSTAT area code has already been replaced by the polity code.

The discriminator for the RoW bucket is a single column: **every** RoW-mapped code has **`fabio_code == 999`** (i.e. "not in FABIO's fixed region list"). It is unrelated to identifiability — each of these countries has an ISO3 and a geometry.

## Root cause

`.aggregate_to_polities()` (`R/read_raw_inputs.R:310`):

1. maps `area_code` → polity via `.add_polity_columns_dt()` (year-aware),
2. drops rows with `NA` polity (this is the correct China‑351 double‑count guard — see below),
3. **sums `value` grouped by `(year, polity_area_code, …)`** — this is the collapse (small countries → `999`, Sudan → `206`),
4. renames `polity_area_code` → `area_code`.

It runs at raw-input ingestion and is called broadly: `R/build_production.R` (×4, incl. `:390`), `R/build_cbs.R:721`, `R/prices.R:363`, `R/read_raw_inputs.R:412`. So the polity collapse is pervasive, not local to production.

## What is correct and should NOT change

- **Dropping FAOSTAT China aggregate `351`.** That is a *redundant total* of components (41/96/128/214) that are themselves present; dropping avoids double-counting. This is separate from the RoW collapse and must stay.
- Genuine **non-countries** (Antarctica, Neutral Zone, "Others (adjustment)", "Unspecified", uninhabited/dependent islands) can legitimately map to RoW or be dropped — they have no per-country home.

## Proposed refactor

- Keep the **base** production (and ideally the shared raw-input path) at **per-country grain** — every identifiable FAOSTAT country/ISO3 retained as its own row (Syria, North Macedonia, Andorra, …).
- Introduce an explicit **`collapse_to_fabio_regions()`** applied **only** at the FABIO/IO-matrix boundary, where a closed region list + RoW catch-all is actually required. Aggregation is one-way; do it at the point that needs it.
- Genuine non-countries still map to RoW (or are dropped) in that downstream step.
- This likely means splitting `.aggregate_to_polities()` into (a) a *mapping/relabel* concern (attach polity columns, drop redundant aggregates like 351) and (b) an *aggregation* concern (sum to FABIO regions), and moving (b) downstream.

## Impact / scope

Non-trivial. `.aggregate_to_polities()` is used across `build_production`, `build_cbs`, `prices`, and raw-input reads; every downstream consumer that assumes polity grain (CBS, trade, footprint, spatialize, IO) would need the explicit aggregation step inserted at the right boundary. This is orthogonal to, and larger than, the in-flight dataset-validation work.

## Validation angle

Once the base is per-country, this becomes checkable: add a WARN-level integrity check that **no real ISO3 country maps into the RoW polity in the base output**, and a conservation check that `sum(individual countries) == RoW aggregate` at the collapse step. (Surfaced while building the `validation/integrity.R` harness — the current design makes both impossible because the sum is baked in.)

## Suggested steps

- [ ] Quantify what real production currently disappears into `999` (per-country, so the stakes are explicit — e.g. Syria, North Macedonia).
- [ ] Separate "map/relabel to polity + drop redundant aggregates (351)" from "sum to FABIO regions" in `.aggregate_to_polities()`.
- [ ] Keep the base output per-country; add `collapse_to_fabio_regions()` at the IO/matrix boundary and wire the downstream consumers to call it.
- [ ] Add the two validation checks above.

lbm364dl:
Closes #120.

## What this does

The base datasets now keep **per-country grain**: every identifiable FAOSTAT reporting area keeps its own `area_code` row, and the FABIO matrix-compatibility collapse happens in one explicit, exported step at the boundary that actually needs it.

- **`.aggregate_to_polities()` is split into its two concerns** (as proposed in the issue):
  - `.harmonize_areas()` (raw-input path): attaches the periodized polity mapping, drops redundant statistical aggregates (FAOSTAT 351 "China" — the existing double-count guard is preserved), canonicalises the area name, and keeps the raw per-country `area_code`.
  - **`collapse_to_fabio_regions()` (new export)**: year-aware sum to the closed FABIO region list (`999` Rest of World, Sudan + South Sudan → `206`). Conservation-preserving and idempotent; drops unmapped aggregates with a warning.
- **`build_primary_production()` / `get_primary_production()` are per-country** and trimmed to identity codes only (`year, area_code, item_prod_code, item_cbs_code, live_anim_code, unit, value, source`). The derived polity columns are pure `(area_code, year)` lookups, so they moved out of the base table into the new export **`add_reporting_polity_columns()`** (same trim for `get_primary_residues()`).
- **The collapse is wired at every FABIO-grain boundary**: CBS input assembly (`.cbs_read_inputs()`), prices, `build_io_model()` / `build_supply_use()`, livestock CBS, feed intake, and all footprint extensions (idempotent, so they work with both old pinned data and the new grain).

## Quantified (2015–2016 build, cached pins)

20 real countries were previously summed into `999` at ingestion: ≈ **44.7 M tonnes, 97.8 M heads, 252 M slaughtered heads** per the 2015–16 slice — led by Syria, Eswatini, North Macedonia, Palestine, Equatorial Guinea. They are now individual rows in the base output (no `999`, no `351`, Sudan `276`/South Sudan `277` kept separately).

Collapsing the new per-country build back to FABIO regions reproduces the old production build for **99.7 % of keys**; the differences are recovered data the old design destroyed:
- **RoW grassland (~8.9 Mha)**: LUH2 grassland for RoW members was silently dropped by a name-keyed join in the old pipeline.
- **Sudan**: the old ingestion-time 276/277→206 merge produced duplicate name-split series that `.dedup_production()` then mangled (e.g. sorghum area 61.6 k ha vs the correct ~3.1 M ha).

## Bonus bug fix: CBS name-split double count

Comparing old vs new `build_commodity_balances()` exposed a real bug on main: **CBS source selection groups by `area` name + `area_code`**, and inputs carried the same `area_code` under two different names — the crosswalk *polity* name from `.aggregate_to_polities()` ("China (PRC)") vs the FAOSTAT *area* name from production (`"China, mainland"`). Both series survived `.select_best_source()` and were **summed** in `.format_cbs_output()`.

Exact reproduction: old CBS China rice production = 354,277,100 = 212,142,000 (FBS series, `"China (PRC)"`) + 142,135,100 (production series, `"China, mainland"`). This inflated CBS values (~2×) for every area whose polity name differs from its FAOSTAT name — China, Türkiye, Iran, Bolivia, Taiwan, Venezuela, … ≈ 11 % of total CBS value in 2015–16.

All CBS inputs now share the canonical crosswalk `area_name`, so source selection merges them correctly.

## Follow-ups

- The WARN-level integrity checks proposed in the issue (no real ISO3 in RoW in base output; conservation at the collapse) belong in the `validation/` integrity harness, which lives on the in-flight dataset-validation branch — to be added there once merged. The same invariants are covered here by unit tests (`test_fabio_regions.R`, `test_read_raw_inputs.R`).
- Pins built from the old grain remain compatible: `collapse_to_fabio_regions()` is a no-op on already collapsed data.

## Checks

- `devtools::test()`: pass (no failures).
- `rcmdcheck` (no tests/vignettes): 0 errors, 0 warnings, 0 notes.
- `lintr`: no lints. `air format .` applied. pkgdown reference coverage verified.


lbm364dl:
Closing: **stale beyond recovery, and the substance is now tracked elsewhere.**

This branch is 530 commits behind `main`. Its diff against current `main` is **+10,684 / −65,795 across 472 files** — merging it would delete most of what has landed since (the SOC models, water balance, soil carbon inputs, spatialize work, the `validation/` harness). There is no useful rebase of a branch whose diff is mostly other people's deletions.

The idea is not lost, and it did not need this branch to survive:

- **Per-country grain in the base datasets** is how `main` works today. `.aggregate_to_polities()` keys on `polity_area_code` and every builder downstream of it carries the reporting-polity columns — see the "Already polity-keyed on main" section of epic #458.
- **The FABIO Rest-of-World question this PR wanted to defer** is now owned by [#419](https://github.com/eduaguilera/whep/issues/419), which measured the thing this PR could only assume: promoting the 16 folded areas to their own `polity_area_code` inflates global feed **13.7x**, export 13.2x, production 1.8x. That is why the fold cannot simply be deferred, and it is why #419 and #414 are the epic's critical path.

Issue #120 stays open. Anyone picking it up should start from #419's measurement rather than from this branch.

lbm364dl:
**Correction to my closing comment.** I cited #419's measurement that promoting the 16 folded areas inflates global feed **13.7x**, and gave that as the reason this PR's deferral of the Rest-of-World aggregation could not simply be picked up.

That figure has since been disproved. PR #555 ran two complete `get_wide_cbs()` builds over 1850-2023 — current `main` against all 61 Rest-of-World members promoted — and global feed comes out at **1.0000**, with the largest column move 1.2% on `stock_addition`. The 13.7x was measured against a baseline corrupted by the `dcast()` duplicate-key fallback (#425, fixed by #429).

This does not reopen the PR: it is still 530 commits behind with a −65,795-line diff, and that is the reason it is closed. But it does change the outlook for **#120**, which stays open. Deferring the Rest-of-World aggregation now looks arithmetically far cheaper than #419 suggested, and the real obstacles are the ones #555 names instead: FABIO matrix comparability for output carrying `polity_area_code` values FABIO's 192-country layout has no column for, and the promotion's unmeasured effect on the footprint, trade and nitrogen chains, which key on the same code. #555 posted its measurement on #120 as well.

PR #125, opened 2026-07-11
feat(landuse): fallow-inclusive arable/permanent land base from FAO RL

eduaguilera:
## What

Adds `get_arable_permanent_land()`: a physically-correct annual/perennial cropland split (`arable_ha` / `permanent_ha` per `area_code` x `year`) sourced from the FAOSTAT RL land-use survey (Arable land / Permanent crops / Cropland), harmonised to polity area codes, 1961+, with a LUH2 CFT backcast pre-1961 spliced onto the 1961 FAO anchor and a `source`/tier column.

Also adds `build_fao_arable_fallow_extension()` (`method_land = "fao_arable_fallow"`): an additive per-crop physical-land method that attributes fallow (FAO Arable − harvested arable) to arable crops via the existing `attribute_fallow_to_crops()` + `gridded_fallow_weights()`, so arable-crop footprints include their rotational/set-aside fallow.

## Why

Every existing crop-area path (`cropland_apportion`, `intensity_divide`, `cropgrids_fallow`) is harvested-area derived and cannot recover FAO's fallow-inclusive physical arable land for rain-fed, fallow-prone economies. Tunisia 2020 perennial share is 0.73 from all three methods vs FAO physical 0.43: cereal harvested area collapses in drought years while physical arable land stays ~constant (it counts fallow). FAO RL is the only physical land-base source, and whep did not previously ingest it. This corrects the annual/perennial denominator used downstream (crop-management/intercropping) and enables fallow-inclusive arable footprints.

## Validation

- Permanent share matches FAO physical to 3 dp: TUN 0.40/0.43/0.46/0.43 (1990-2020, was 0.73), PRT 0.25 -> 0.47 rising smoothly (no 2020 step); FRA/USA low, MYS 0.86 / CIV 0.6 preserved.
- `arable_ha + permanent_ha == cropland_ha` exact except 4 pre-existing raw-FAO rows where FAO itself reports Arable > Cropland (Peru 1961/62, Lesotho 2017/18), handled by `permanent = max(0, cropland - arable)`.
- `sum(arable physical) == FAO Arable land` (max rel err ~1e-16); perennials receive zero fallow.

## Scope / safety

Additive only. Existing crop-land methods, their defaults, and `footprint_balance.R` are unchanged (byte-identical to HEAD). Adds 8 tests; full suite green (`FAIL 0 | PASS 2879 | SKIP 1` pre-existing).

## Notes for review

- Whether to switch the footprint balance's `cropgrids_fallow` to the new FAO-arable fallow (changes published arable footprints portfolio-wide) is deliberately left as a follow-up decision, not part of this PR.
- Surfaced upstream data quirks (not addressed here): FAO Arable > Cropland for a few country-years; fodder harvested-area inflation in `primary_all` affecting only the per-crop fallow distribution.


lbm364dl (changes requested):
Potential double-counting of temporary grassland

Before merging, please clarify the intended accounting invariant here and adjust the implementation if needed.

This extension reconciles crop occupation to the full FAOSTAT arable-land total. Elsewhere, build_grassland_land_extension(source = "luh2") can also return CBS 3002, Temporary grassland, from primary production. If consumers combine both extensions, temporary grassland appears to be included once inside the FAOSTAT arable target and then added again as CBS 3002.

Could you confirm which invariant is intended? I would expect:

  ordinary crop occupation, including allocated fallow, + CBS 3002 = FAOSTAT arable land

rather than:

  ordinary crop occupation = FAOSTAT arable land, followed by adding CBS 3002

There is also a provenance ambiguity that affects the fix. The grassland-extension documentation says CBS 3000 and 3002 are LUH2 pasture and rangeland, but .build_grassland() maps both LUH2 pastr and range to CBS 3000. CBS 3002 is instead mapped from production item 996, and its hectares pass through the fodder reconstruction using FAOSTAT production, dry-matter yields, EuropeAgriDB data, and gap filling.

Please clarify whether modelled CBS 3002 or the direct FAOSTAT temporary-meadows-and-pastures series should be authoritative, and add a reconciliation test/check so the combined crop plus temporary-grassland occupation cannot silently exceed the intended arable-land target.

lbm364dl:
Resolving the temporary-grassland double-count thread.

Confirmed with the domain author: **temporary grassland is a type of crop — part of cropland/arable land, not grassland** (FAO's `Arable land` definition includes *temporary meadows and pastures*). So this extension, which reconciles ordinary crops to the full FAO Arable land total, already absorbs the temporary-grassland slice. The intended invariant is:

```
ordinary crop occupation (incl. fallow) + CBS 3002 = FAO Arable land
```

**Decision for this PR:** merge as-is. It is purely additive — `fao_arable_fallow` is not the footprint-balance default, so nothing currently sums both extensions and the double-count is latent, not live. I've added a `@section` to `build_fao_arable_fallow_extension()` documenting the invariant and warning consumers not to add CBS 3002 on top.

**Provenance:** whep's CBS 3002 is modelled (item 996 → fodder reconstruction) rather than the patchy direct FAO series, so we treat modelled CBS 3002 as authoritative.

The real work — net modelled CBS 3002 out of the arable target, switch `.land_balance_extension()` to this method as the crop-side default, and add a reconciliation test — is tracked in #342 as the follow-up (a PR for it is starting now).

lbm364dl (approved):
Approving. Temporary-grassland double-count resolved for this PR by documenting the invariant; the fix (net out CBS 3002 + switch footprint default + reconciliation test) is tracked in #342. Additive-only, suite green, lint/format clean.

PR #127, opened 2026-07-22
feat(gapfilling): log-space interpolation for fill_linear

eduaguilera:
## What

Adds a back-compatible `log_space = FALSE` argument to `fill_linear()`. When `TRUE`, each interior gap segment whose **two bracketing anchors are both finite and strictly positive** is interpolated in log space (constant compound growth rate); any segment with a non-positive or non-finite anchor **falls back to linear interpolation**. Carry-forward / carry-backward at the series ends are unchanged, and no extrapolation beyond the anchor range is introduced.

Log-space fills are labelled `"Log-linear interpolation"` in the `source_<value>` column, distinct from `"Linear interpolation"`, so the choice survives downstream provenance.

## Why

Roughly exponential series (energy use, and many economic/biophysical growth series) are systematically **under-filled at mid-gap** by linear interpolation: the geometric mean, not the arithmetic mean, is the constant-growth midpoint. Concretely, the geometric midpoint of `1` and `1024` is `32`, whereas linear interpolation returns `512.5`.

This is promoted from the **energy-hist** long-term-energy pipeline, whose `loglinear_approx()` (landed in commit `dada470` of that repo) is the reference implementation. There, linear interpolation over an exponential gas gap around 1950 over-filled the world-energy series by roughly 30%; the log-space rule removed that bias while leaving non-exponential series (e.g. EROI ratios) on the linear path.

## Design

Extends `fill_linear()` **at origin** with a boolean toggle rather than a sibling `fill_loglinear()`:

- `fill_linear()`'s argument surface is entirely boolean toggles (`interpolate`, `fill_forward`, `fill_backward`); a boolean `log_space` reads as "interpolate in log space" and fits that local style. It is placed right after `interpolate` (all in-repo callers pass every argument past `value_col` by name, so no positional break).
- Log-linear is **not universally more rigorous** than linear (it is undefined for non-positive anchors and inappropriate for non-exponential series), so the package's `method = ` convention — whose default is "the most rigorous method" — does not cleanly apply here; the safe, established general-purpose default is linear. Back-compat also requires the default to reproduce current behaviour exactly.
- All three interpolation code paths (ungrouped `.fill_linear_vec`, grouped-vectorized, grouped-smoothing `.fill_linear_by_group`) route through a single private `.loglinear_interp()` primitive, so the semantics are identical everywhere.

## Back-compatibility

Hard guarantee. With `log_space = FALSE` (the default) every code path is byte-identical to before: the log branches are skipped and the source label is unconditionally `"Linear interpolation"`. A regression-lock test asserts that the default call is identical to an explicit `log_space = FALSE` call **and** reproduces the established linear interpolation output.

## Tests

New tests in `tests/testthat/test_gapfilling.R`:

- geometric midpoint of `(0, 1)` and `(10, 1024)` at `t = 5` is `32`, explicitly `!=` the linear `512.5`;
- zero and negative anchors fall back to linear (`"Linear interpolation"`);
- mixed log/linear segments within one series (distinct labels coexist);
- `.by` grouping (per-group geometric midpoints, no cross-group bleed);
- the grouped-smoothing path with `value_smooth_window`;
- default-argument regression lock.

Verification (worktree, global R 4.5.2 library):

```
devtools::document()   -> only man/fill_linear.Rd regenerated; NAMESPACE unchanged
devtools::test(filter = "gapfilling")  -> FAIL 0 | WARN 0 | SKIP 0 | PASS 128
devtools::test()  (full suite)         -> FAIL 2 | WARN 0 | SKIP 0 | PASS 699
lintr (CI config) on both changed files -> 0 lints
air format . -> idempotent (no changes)
pkgdown coverage check -> empty (fill_linear already listed)
```

The 2 full-suite failures are the pre-existing `test_commodity_balance_sheet.R` failures documented in `CLAUDE.md` (unrelated CBS module; not touched here).

### DIFF
--- a/R/gapfilling.R
+++ b/R/gapfilling.R
@@ -14,6 +14,14 @@
 #' @param time_col The column containing time values. Default: `year`.
 #' @param interpolate Logical. If `TRUE` (default),
 #'   performs linear interpolation.
+#' @param log_space Logical. If `TRUE`, interior interpolation is performed in
+#'   log space (constant compound growth rate) for each gap segment whose two
+#'   bracketing anchors are both finite and strictly positive; any segment with
+#'   a non-positive or non-finite anchor falls back to linear interpolation.
+#'   Default: `FALSE`, i.e. linear interpolation everywhere. Log-space fills are
+#'   labelled `"Log-linear interpolation"` in the source column, distinct from
+#'   `"Linear interpolation"`, so the choice survives downstream. Carrying
+#'   forward or backward is unaffected.
 #' @param fill_forward Logical. If `TRUE` (default),
 #'   carries last value forward.
 #' @param fill_backward Logical. If `TRUE` (default),
@@ -47,11 +55,13 @@
 #'   interpolate = FALSE,
 #'   .by = c("category"),
 #' )
+#' fill_linear(sample_tibble, value, log_space = TRUE, .by = c("category"))
 fill_linear <- function(
   data,
   value_col,
   time_col = year,
   interpolate = TRUE,
+  log_space = FALSE,
   fill_forward = TRUE,
   fill_backward = TRUE,
   value_smooth_window = NULL,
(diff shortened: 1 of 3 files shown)


lbm364dl (approved):
Reviewed via code inspection: log-space path traced across all three code paths, no correctness issues; back-compat locked by regression test. Style nits waived (see #131).

PR #128, opened 2026-07-22
feat(checks): check_series_jumps() — first AFE validation-library detector

eduaguilera:
## Summary

Adds `check_series_jumps()`, the **first landed detector of the reusable
`check_*` validation library** described in the AFE data-and-code validation
framework (`afse-wiki/wiki/decisions/afse-data-validation-framework.md`). It is
the framework's **level-2 (within-series) detector**: it scans each series for
consecutive-observation ratios that fall outside a plausible band, catching
unjustified single-year steps, spikes, and the level shifts that appear where
two sources are spliced.

## Origin

Generalised from the energy-hist jump scan, where a consecutive-year
world-total ratio scan against a break-year allowlist locked the fix for the
1820 source-splice seam. That regression test (`tests/test_jumpscan.R`, guarding
commits `2fe5d33` / `4a33958`) is a hard-coded, single-panel special case; this
promotes it into a grouped, band-parameterised WHEP function usable both as an
inline pipeline guard and as a test backend, per the framework's two-deliverable
plan (reusable library + generated tests).

## API

```r
check_series_jumps(
  data, value_col, time_col = year, .by = NULL,
  ratio_bounds = c(0.55, 1.6), bands = NULL, min_value = 0,
  consecutive_only = TRUE, allowlist = NULL, verbose = TRUE
)
```

Per series (`.by`), it takes the ratio of consecutive observations (only single
time steps when `consecutive_only`) and flags ratios outside `ratio_bounds`
where both values exceed `min_value`. Two pieces of the framework's metadata
model are exposed as arguments:

- `bands`: an optional per-group override table (group columns + `lo`/`hi`)
  implementing the framework's **per-variable jump bands** — area is tight, yield
  is wide, so a single global band is wrong for a mixed panel.
- `allowlist`: an optional table matched on the group columns + `time_col`
  marking **documented break years**; matching flags are returned with
  `allowlisted = TRUE` rather than dropped, so nothing is silently swept away.

It returns a tidy tibble of flags (group columns, time, `prev_value`, `value`,
`ratio`, `allowlisted`) with a typed zero-row shape when nothing flags, and
reports counts `cli`-style like the existing `check_*` functions. The robust
MAD-of-log-ratios variant (Hampel/Leys anchors in the framework) is left as a
documented future extension.

## Tests

`tests/testthat/test_check_series_jumps.R` (30 assertions): a known step flags at
the right year and ratio; a documented break returns `allowlisted = TRUE`;
sub-`min_value` onsets are not flagged; non-consecutive gaps are skipped unless
`consecutive_only = FALSE`; per-group `bands` override the global band; a flat
series returns a typed zero-row tibble; ungrouped use reports `cli` counts; and
input validation aborts on missing columns and inverted bounds.


lbm364dl (approved):
Reviewed via code inspection; the duplicate value-column issue found in review was fixed on-branch (source column now dropped after coercion, with a regression test). CI green.

PR #129, opened 2026-07-22
feat(consolidate): add consolidate_sources() multi-source panel winner-consolidation

eduaguilera:
## Summary

Generalises the priority-based panel deduplication that builds the long-term historical energy panel into a reusable, tidy-style WHEP function, `consolidate_sources()`. Given a long multi-source panel, it selects one winning row per `(.by, time_col)` cell and returns it with provenance columns.

It operationalises the AFE decision *Consolidate multi-source panels measure-consistently* (afse-wiki: wiki/decisions/measure-consistent-panel-consolidation.md): measure identity is part of the dedup key's semantics, and priority alone cannot arbitrate cells whose sources report different measures.

## Two silent-corruption classes this prevents

Both were found in the 2026-07 energy-hist audit; both are invisible to schema tests and world-total benchmarks, so only measure-aware, drop-aware consolidation removes them:

1. **Measure-mixing re-basing (2,402 cells).** A production-basis series won contested cells against a consumption-basis series on ordinary priority, silently re-basing "consumption" onto production: oil exporters overstated (SAU 1960 oil 2.70 â†’ 0.036 EJ) and importers understated (GBR 1950 oil 0.002 â†’ 0.400 EJ). Fixed in energy-hist by a scoped effective-priority penalty (`git 83f30fc`). Here: `measure_basis` + `measure_penalty`, with `measure_exempt` for world-level cells where production == consumption, and lone-reporter survival.

2. **Spurious observed zeros (9,596 cells).** WorldBank electricity-generation-share rows carried `value = NA`, which an `na.rm` sum coerced to 0; the unlisted source's fallback priority then won every cell no primary-energy source reported, injecting 9,596 `"Observed 0 EJ"` panel cells (Nuclear 4,962; Coal 1,827; Renewables 1,770; Hydropower 985; Oil 21). Fixed in energy-hist by pinning the source out of consolidation (`git ed2e51d`). Here: the pre-consolidation hard drop (`drop_at`) removes a pinned source before any cell is contested, so it never wins even an uncontested cell.

## API

```r
consolidate_sources(
  data, value_col, source_col, priority, .by = NULL, time_col = year,
  drop_at = 100L, measure_basis = NULL, measure_penalty = 1000L,
  measure_exempt = NULL, tie_break_coverage = TRUE, quality_col = NULL,
  quality_levels = NULL, continuity_override = TRUE, verbose = TRUE
)
```

- `priority`: named integer vector (or two-column data frame) mapping source â†’ rank; lower wins. Sources absent from the map take the documented fallback rank `drop_at - 1L` (kept, but below every listed source).
- **Hard drop**: rows whose source ranks `>= drop_at` are removed before consolidation, so a pinned source never wins even an uncontested cell.
- **Measure demotion**: rows flagged by `measure_basis` (a data frame keyed on the source column, optionally scoped by a category column) get `+measure_penalty` on effective rank, losing any contested cell a measure-consistent source also reports; a lone flagged reporter still wins its own cell. `measure_exempt` is a filter expression (e.g. `region == "WLD"`) exempting rows from the penalty.
- **Winner** per `(.by, time_col)` cell = lowest effective rank; ties broken by broader within-series coverage (when `tie_break_coverage`), then `quality_col` ordered per `quality_levels`, then ascending source name (logged when `verbose`).
- **Continuity override** reverts an isolated single-period winner flip when both flanking periods share a different winner that also reports the middle period.
- **Returns** the winning rows plus `n_sources` (distinct sources contesting the cell after the drop), `source_rank` (base rank), `effective_rank` (base + any penalty), and `measure_demoted` (winner carried the penalty, true only for a lone flagged source).

## Test evidence

New `tests/testthat/test_consolidate_sources.R` mirrors the energy-hist `tests/test_dedup.R` synthetic cases: priority wins; coverage tie-break beats alphabetical; quality preference; `>= drop_at` hard-drop including an uncontested pinned source (which vanishes); measure penalty flips a contested cell while a lone demoted source still wins; exempt keys unaffected; continuity override reverts `A, B, A` but leaves a genuine two-year run; deterministic output ordering; global (`.by = NULL`) consolidation; duplicate-row abort.

Verified in the branch worktree:

- `devtools::test()` â€” full suite `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 2887 ]` (36 of them new).
- `rcmdcheck::rcmdcheck(build_args="--no-build-vignettes", args=c("--no-tests","--ignore-vignettes"))` â€” `Status: OK`, 0 errors / 0 warnings / 0 notes (no undeclared-NSE-global note; example runs).
- `lintr::lint` with the CI linter config â€” no lints.
- `air format .` â€” clean.
- `devtools::document()` â€” `man/consolidate_sources.Rd` + `NAMESPACE` export; every `man/*.Rd` registered in `_pkgdown.yml`.

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)


lbm364dl:
Pushed review fixes in the latest commit:

1. **Continuity override × measure demotion — decision documented.** The continuity pass no longer reverts an isolated cell won by a measure-consistent source back to a measure-demoted flanking source. Rationale: a single-period source switch is cosmetic, while a measure switch (e.g. mixing production into a consumption series) corrupts the data — exactly what the penalty exists to prevent. The reversion is still allowed when both rows carry the penalty. Documented in stage 4 of the roxygen details; regression test added.

2. **Deprecated `filter(across(...))` → `if_all()`** in the tie logging, plus a `verbose = TRUE` test so the default verbose path is now exercised by the suite.

3. **Argument grouping per CLAUDE.md**: `measure_basis`/`measure_penalty`/`measure_exempt` → `measure = list(basis, penalty, exempt)` and `tie_break_coverage`/`quality_col`/`quality_levels` → `tie_break = list(coverage, quality_col, quality_levels)`. Note two call-syntax changes: `measure$exempt` is a one-sided formula (`~ region == "WLD"`) and `tie_break$quality_col` is a string. Done pre-merge since it would be a breaking API change later.

The full test file passes locally (13 tests, 41 expectations, 0 warnings).

lbm364dl (approved):
Review items addressed on-branch: continuity override no longer undoes the measure penalty (with regression test), deprecated filter(across()) replaced by if_all() with a verbose-path test, and the 15 args grouped into measure/tie_break lists. Local tests pass (13 tests, 41 expectations, 0 warnings); CI green.

PR #302, opened 2026-07-23
fix(polities): retain aggregate recent years, drop dead tiebreak, symmetric partner code

### LINKED ISSUE #218: polities: .current_area_lookup hardcodes is_current against the literal year 2025
## What

`.current_area_lookup()` (`R/polities.R:17`) defines "current" via `polity_end_year >= 2025` and treats NA end-year as *not* current (1 such row exists in the crosswalk). This feeds `add_area_name()`, `.energy_area_iso3()`, etc.

## Why it matters

When the crosswalk is regenerated with end years beyond 2025, or open-ended polities are encoded as NA, the wrong (historical) row can win the per-area dedup with no warning. Time-bomb; correct on today's data (max end year is exactly 2025).

## Suggested fix / check

Derive "current" from the max end year in the data (or treat NA end-year as open-ended/current), rather than a literal constant.

---
_From deep static audit (2026-07-22), parallel subsystem review pass. Deduplicated against then-open issues. Not verified by running unless stated._

### LINKED ISSUE #295: polities: aggregate recent-year drop, dead exact_start tiebreak, asymmetric partner polity_area_code
Three application-layer defects in `R/polities.R`, all in the year-aware polity mapping path (`.add_polity_columns_dt`) and its role wrappers.

### 1. Aggregate reporting areas silently drop their most-recent years (HIGH)

The primary non-equi period join (`join_start_year <= year <= join_end_year`) *includes* aggregate polities, but the out-of-period fallback then hard-excludes them (`lookup_polity_type != "aggregate"`). A pure-aggregate area whose only polity period ends early therefore gets `polity_code = NA` for recent years, and callers that filter unmapped rows (e.g. `build_cbs.R`, `read_raw_inputs.R`) drop that data entirely.

Concrete:
- area 904 "Latin America Other" (only `RLAM-…-2013`) loses 2014-2023.
- areas 901/902/903/905/906 lose 2022-2023.

Fix: let the nearest-period fallback apply to aggregate areas that have no non-aggregate period covering the year, so recent years map to the aggregate polity instead of being dropped — while still preferring a real (non-aggregate) polity over an aggregate one where both are candidates for the same row.

### 2. `exact_start` tiebreak is dead code

`exact_start := polity_start_year == join_start_year` runs *after* data.table's non-equi join, which overwrites `join_start_year` with the query `year`. So the expression degrades to `polity_start_year == year` and never changes the pick (checked across 1961-2023). Fix: remove the misleading machinery, keeping the `polity_start_year DESC` tiebreak that actually decides.

### 3. Partner side loses `polity_area_code`

`.add_reporting_polity_columns` promotes `reporting_polity_area_code -> polity_area_code`, but `.add_partner_polity_columns` just drops `partner_polity_area_code` with no equivalent, so FABIO-collapsed partners are not canonicalized symmetrically. Fix: emit a symmetric `partner_polity_area_code` output column.

Finding 4 (hardcoded `is_current = polity_end_year >= 2025`) is tracked in #218 and fixed alongside these.

lbm364dl:
Application-layer fixes in `R/polities.R`, verified against the actual mapping path.

## Changes

1. **Aggregate reporting areas no longer drop their most-recent years.** The primary non-equi period join includes aggregate polities, but the out-of-period fallback hard-excluded them (`lookup_polity_type != "aggregate"`), so pure-aggregate areas whose period ends early got `polity_code = NA` and were dropped by callers. The fallback now extends aggregate areas to their nearest period too (e.g. area 904 "Latin America Other" / `RLAM-…-2013` now keeps 2014-2023; areas 901/902/903/905/906 keep 2022-2023), while still preferring a real non-aggregate polity over an aggregate one where both are candidates for the same row. China area 351 stays unmapped, so no double-counting is reintroduced.

2. **Removed the dead `exact_start` tiebreak.** data.table's non-equi join overwrites `join_start_year` with the query `year`, so `polity_start_year == join_start_year` degraded to `== year` and never changed the pick. The `polity_start_year DESC` order that actually decides is kept, with a clarifying comment. Also removed the now-unused `exact_start` from `globalVariables`.

3. **Symmetric partner `polity_area_code`.** `.add_partner_polity_columns` previously dropped `partner_polity_area_code` with no equivalent, unlike the reporting side which promotes `reporting_polity_area_code -> polity_area_code`. It now emits `partner_polity_area_code`. `build_trade.R`'s selection uses `any_of()` + `everything()`, so this is backward compatible.

4. **Open-period sentinel derived from data** (#218). `is_current` no longer keys off the hardcoded literal `2025`; it uses the crosswalk's latest end year.

## Tests (`tests/testthat/test_polities.R`)

- aggregate area 904 keeps 2013-2023 mapped to `RLAM-1850-2013`;
- out-of-period rows extend to their nearest period (incl. aggregates 15/151/904);
- partner path emits `partner_polity_area_code`;
- current mapping selects the latest-ending (open) period, tied to the data-derived max rather than a literal.

All `test_polities.R` tests pass; `air format .` clean; `lintr` clean.

Closes #295
Closes #218


lbm364dl:
Brought up to date with `main` (517 commits). Full suite **5427 pass, 0 fail, 8 skip**, lintr and air clean.

**Change 1 of the four is withdrawn.** Three survive untouched.

## Withdrawn: extending aggregate reporting areas to their nearest period

The symptom you reported is real and still reproduces — area 904 "Latin America Other" carries only `RLAM-1850-2013` in the embedded vintage, so its post-2013 rows resolve to `NA` and callers drop them. But two measurements say this is the wrong layer.

**Upstream already fixed it at source.** All seven reporting buckets now run to 2025: `RLAM-1850-2013` → `RLAM-1850-2025`, the six `-1850-2021` buckets and `ROW-1850-2023` likewise. That is one of the ten retirements in #530, and PR #551 is doing that re-sync now. Under the epic's own rule a territorial validity span is upstream's fact.

**And nearest-distance extension is not safely asymmetric.** It back-fills years *before* an aggregate's start as readily as after its end, so an 1830 Guadeloupe figure gets booked to `ROW-1850-2023` — a bucket that did not exist. `main` has since added a test asserting exactly those rows are dropped, and it names the cost: **64 rows / 1,722,000 t** of the historical trade feed. That test failed against this branch, and I think it is right.

## Kept, all three independent of the above

| change | status |
|---|---|
| dead `exact_start` tiebreak removed | kept |
| symmetric `partner_polity_area_code` | kept |
| open-period sentinel derived from data (#218) | kept |

## Test updates

Two tests here asserted the withdrawn behaviour. They now pin what replaces it, with the reasoning inline and the expectation that **#530 flips them back**: areas 15, 151 and 904 are all typed `aggregate` and correctly resolve to `NA` past their period, while area 2 Afghanistan still extends because it is `national`. `test_build_trade.R`'s two `expect_named` lists gain `partner_polity_area_code`, which is this branch's own column.

**Closes #218 as before. #295 is now only partly addressed** — its aggregate-recent-years half is answered by #530/#551 rather than here, so you may want to re-scope or close it against that instead.

lbm364dl:
Made both aggregate guards **vintage-independent**, because PR #551 (the #530 re-sync) would otherwise have inverted them.

They asked about area 904 in 2021 — outside `RLAM-1850-2013`, but **inside** the `RLAM-1850-2025` that #551 lands. A hardcoded calendar year silently starts asking the opposite question when the bucket moves, which is how a guard becomes its own negation with nobody editing it. Both now read the bucket's declared end year off the crosswalk and test `end - 1` and `end + 1`.

The rule under test is unchanged. Verified against both vintages: 44 pass / 0 fail on main's 740-row data, and the same two tests pass with `polity/530`'s regenerated data overlaid. **The two branches can now land in either order** — only a textual merge on `test_polities.R` remains, which both touch.

PR #305, opened 2026-07-23
feat(polities): build area→polity crosswalk from curated in-repo alias table

lbm364dl:
## What

Builds the FAOSTAT/FABIO `polity_area_crosswalk` from a curated alias
table `faostat_aliases.csv` instead of the legacy prefix-guess. The alias
table maps each numeric reporting `area_code` to a periodized WHEP
`target_polity_code` chain (e.g. Sudan-former `206 → SUD` pre-2011 vs.
`276 → SDN` / `277 → SSD` after secession), which the old iso3c/reporting
coalesce could not disambiguate. Areas the table does not cover keep the
legacy coalesce fallback (including the FABIO `fabio_code == 999 → Rest of
World` bucket), so no routed area silently becomes `NA`.

## Hardening fixes applied

1. **Reproducibility (HIGH).** The alias table is now **vendored** into the
   repo at `inst/extdata/harmonization/faostat_aliases.csv` and read by
   default; `WHEP_POLITIES_ALIASES` still overrides for a fresh
   whep-polities export. Previously an absent external CSV only `warning()`ed,
   fell back to the degraded legacy guess, then unconditionally clobbered the
   committed `.rda`. The build now **aborts** if the resolved path is missing
   rather than producing and saving a degraded crosswalk. A rebuild is now
   reproducible from the checkout alone.
2. **`mapping_status` conflation (MEDIUM).** The two redundant `matched`
   branches (the first was dead code) collapsed curated matches, legacy
   iso3c/reporting guesses and the RoW bucket into one label. Split into
   `matched_curated` / `matched_fallback` / `row` / `unmapped`; removed the
   dead branch; updated `R/table_mappings.R` `@format` docs.
3. **Junk NA row (LOW).** The blank-code "Africa Other" row produced a
   contradictory `NA` `area_code` yet carrying a polity and a
   `not_a_reporting_area` status. `is.na(area_code)` rows are now dropped
   before the crosswalk is assembled.
4. **Test gaps.** Tests now assert **value** correctness (specific
   `area_code → polity_code` for periodized-name countries: Sudan
   206/276/277, Czechoslovakia 51, USSR 228, Yugoslav SFR 248, French Guiana
   69, Palestine 299, China 41/351), plus a closed `mapping_status`
   vocabulary and class-count guard so a no-alias rebuild divergence is caught
   — not just presence/non-NA.

## Regeneration

Rebuilt `data/polity_area_crosswalk.rda` (and the downstream
`regions_full` / `polities_cats` which carry `reporting_polity_code` from
the crosswalk): **615 rows**, **China 351 unmapped**, statuses
**577 matched_curated / 6 matched_fallback / 31 row / 1 unmapped**. Locking
the `.rda` to the in-repo CSV also surfaced that the previously committed
`.rda` no longer matched the current external table — exactly the drift the
vendoring fix removes.

## Known follow-ups (need a design decision — intentionally not fixed here)

- **`reporting_polity_code` bare-vs-periodized convention mismatch** against
  the rest of the package. Needs a convention decision before changing.
- **`polity_area_code = coalesce(fabio_code, area_code)`** merges South Sudan
  into Sudan-206 and Ethiopia 62+238 at the FABIO grain. This is intended for
  FABIO-region aggregation but is flagged so it is not mistaken for a bug.


lbm364dl:
Closing: **superseded by `main`.**

The crosswalk is now built from upstream's published FAOSTAT map rather than a prefix guess, which is what this PR set out to achieve. Commit `f5a2037e` ("Take the FAOSTAT area-to-polity mapping from upstream's published map") put `data-raw/table_mappings.R` on `WHEP_POLITIES_FAOSTAT_MAP`, and — the specific hardening this PR argued for — it **aborts** rather than degrading:

```
"x" = "This aborts rather than falling back to prefix matching, because a
       [degraded crosswalk] ..."
```

So both of this PR's central claims are satisfied by a different implementation: a curated map is the authority, and a missing map is fatal instead of silently producing a degraded crosswalk.

What is *not* carried over is the vendored `inst/extdata/harmonization/faostat_aliases.csv`. That is deliberate on `main` — the map lives upstream in `whep-polities` and is read through an env var, so there is one authority rather than a second in-repo copy. That is the same reasoning as [#457](https://github.com/eduaguilera/whep/issues/457) (a pin adds an uncheckable second copy).

Mechanically it is also unlandable as-is: 530 commits behind `main`, 9 conflicts including `R/table_mappings.R`, `data-raw/table_mappings.R` and all four `data/*.rda`.

The `mapping_status` split this PR proposed (`matched_curated` / `matched_fallback` / `row` / `unmapped`) is worth keeping — `main` still has the coarser vocabulary. Filing that separately rather than losing it here.

PR #313, opened 2026-07-23
fix(scrape_faostat): keep China aggregate distinct from mainland

### LINKED ISSUE #158: scrape_faostat: "China" and "China, mainland" both map to CHN (double-counts the largest producer)
## What

FAOSTAT bulk data contains both "China" (aggregate area 351, incl. Taiwan/HK/Macao) and "China, mainland" (area 41). `fillCountryCode` maps "China" → CHN, and the manual fix at `scrape_faostat.R:87` *also* sets "China, mainland" → CHN.

## Why it matters

Any downstream `ISO3_CODE` grouping/filter double-counts China — roughly 2× for the world's largest agricultural producer.

## Suggested fix / check

Map only one of the two to CHN (drop the aggregate "China" 351, keep "China, mainland" 41, per FAOSTAT convention), or assert one ISO3 per source area. Add a test: no ISO3 code appears for two distinct source area codes after harmonization.

Minor, same file: `require`/`detach` without `on.exit` (`:28, :64`) can leave `FAOSTAT` attached / detach a user's pre-attached copy on error.

---
_From a static read-only code audit (2026-07-22). Line references from source reading, not verified by running — confirm with a test before fixing. Related to #131._

lbm364dl:
## What

FAOSTAT bulk data contains both the aggregate **"China"** (area 351, which bundles mainland, Hong Kong, Macao and Taiwan) and **"China, mainland"** (area 41). In `.populate_iso3_code()`, the manual fix forced `"China, mainland" → CHN` while `FAOSTAT::fillCountryCode()` could also assign `CHN` to the aggregate `"China"`. Any downstream `ISO3_CODE` grouping/filter then double-counted the world's largest agricultural producer (~2x).

## Fix

Leave the aggregate `"China"` (351) unmapped (`NA`), matching the polity crosswalk convention that deliberately leaves 351 unmapped, so only `"China, mainland"` (41) resolves to `CHN`. This is robust regardless of the `fillCountryCode` version's China behavior.

## Test

Added a test asserting:
- `"China, mainland"` → `CHN`
- the aggregate `"China"` stays `NA` (never `CHN`)
- no ISO3 code is shared by two distinct source areas

Closes #158


lbm364dl:
Closing as **superseded** — the defect is already fixed on `main`, by a
different change.

Verified against `main` @ 2f3102aa:

```
China           -> NA     (the aggregate, area 351)
China, mainland -> CHN
```

`.match_fao_area_to_iso3()` / `.fao_area_iso3_lookup()` were rewritten to
reproduce `FAOSTAT::fillCountryCode()`'s matching rule — an area name resolves
only when all its matches fall in a single profile row — so the "China"
aggregate, which names three profile rows, stays unmatched. The code comment
calls that case out by name, and there is already a regression test at
`tests/testthat/test_scrape_faostat.R:126-133` asserting exactly what this PR
proposed to assert.

That also explains this branch's 6 failing checks: it re-adds a fix that has
landed, against a tree that has moved on.

Nothing here is lost. If #158 is still open it can be closed against the current
implementation rather than this branch.

PR #325, opened 2026-07-23
fix(cbs): wire historical trade into pre-1961 CBS extension

### LINKED ISSUE #141: build_cbs: historical trade is read but never used (pre-1961 extension is silently dead)
## What

`.read_historical_trade()` is called (`build_cbs.R:621`) and stored as `inputs$trade_hist` (`:644`), but `.assemble_cbs_sources()` binds only `fbs_new, fbs_old, cbs, primary, trade`. `trade_hist` is never bound in, so no row ever has `source == "trade_hist"` and the exclusion filter at `:1564` filters nothing.

## Why it matters

The docstring for `.read_cbs()` (`:170`) promises the series is extended back to 1850 "using historical trade data." In fact, for `start_year < 1962` the pre-1961 import/export values come only from interpolation/proxy fills of FAOSTAT-era data — the historical trade evidence is downloaded (two pins) and discarded.

Note: `.read_historical_trade()` also assigns the exports pin `element := "import"` and the imports pin `element := "export"` (`:749-760`) — a possible deliberate mirror-flow convention or a direction swap. It's currently masked because the data is unused; whoever wires `trade_hist` in must resolve this first.

## Suggested fix / check

Bind `trade_hist` into `.assemble_cbs_sources()` (or remove the read + dead filter if the extension is intentionally interpolation-only). Add a provenance test asserting every input read in `.cbs_read_inputs()` is reachable in the assembled output.

Related: #102 (pre-1962 back-cast coverage).

---
_From a static read-only code audit (2026-07-22). Line references are from source reading and were **not** verified by running the pipeline — confirm with a test before fixing. Part of the validation-over-style effort in #131._

lbm364dl:
Closes #141.

## Problem

`.read_historical_trade()` was read into `inputs$trade_hist` but `.assemble_cbs_sources()` only bound `fbs_new, fbs_old, cbs, primary, trade`. No row ever carried `source == "trade_hist"`, so the exclusion guard at `.assemble_cbs_sources()` never matched and the pre-1961 import/export extension was **silently dead**: for `start_year < 1962`, historical import/export evidence was downloaded (two pins) and discarded.

## What this does

- **Wires it in.** `trade_hist` is now bound into the assembled sources under the `trade_hist` source label (via a small `.prepare_trade_hist_source()` helper), restricted to years `< 1961` so the FAOSTAT era (>= 1961) is untouched. This is the only import/export source feeding the pre-1961 extension, and it now flows through `.select_best_source()` into the historical destiny/domestic-supply fill.
- **Resolves the flow-direction question flagged in the issue.** The read function previously labelled the exports pin `element := "import"` and the imports pin `element := "export"`. Checked against the 1961 FAOSTAT overlap: the exports pin gives USA wheat ≈ 19.2 Mt (a known ~17-19 Mt exporter) vs ≈ 0.2 Mt on the imports pin, and Egypt (importer) ≈ 0.4 vs ≈ 661. So it is a genuine direction swap, not a mirror-flow convention — the flows are now labelled directly.
- **Avoids the naming-convention trap.** `trade_hist` carried an `area_name` alongside its polity `area_code`, and one polity code could receive several ISO3-derived names. Historical trade is now aggregated by integer `area_code` only, and the area name is derived from the code via `add_area_name()` — the same convention primary production uses, which is what the extension joins production against.

## Verification

Exercised against the real cached pins and with a fixture through `assemble -> select -> extend`: with historical trade a 1950 wheat row gains `import`, `export`, and `domestic_supply = production + import - export`; without it those elements are absent (pre-1961 previously had no import/export evidence). New tests in `test_build_cbs.R` assert both.

`devtools::test()` on `test_build_cbs.R`: all pass, no warnings. `lintr` clean on both changed files. `air format .` applied.

## Scope / remaining

Kept scoped to wiring the input in (open PRs #292 and #284 touch other `build_cbs.R` functions). For edge polities where the legacy production `area_code` differs from the polity `area_code`, historical trade may not align with production and simply won't contribute — no double-count, no regression vs. the previous behaviour (where it never contributed at all).

### DIFF
--- a/R/build_cbs.R
+++ b/R/build_cbs.R
@@ -739,25 +739,32 @@ build_processing_coefs <- function(
     .(item_cbs, item_cbs_code)
   ]
   cbs_trade <- data.table::as.data.table(whep::cbs_trade_codes)
+  # Map the ISO3 reporter to the polity area code only. The human-readable
+  # area name is derived from the code afterwards (see below) so it matches
+  # the primary-production naming convention exactly, rather than trusting a
+  # potentially multi-valued name carried alongside the code.
   area_bridge <- .current_area_lookup(include_unmapped = FALSE)[
     !is.na(area_iso3c),
-    .(iso3c = area_iso3c, area = area_name, area_code = polity_area_code)
+    .(iso3c = area_iso3c, area_code = polity_area_code)
   ]
   area_bridge <- unique(area_bridge, by = "iso3c")
 
+  # The exports pin holds reporter exports and the imports pin holds reporter
+  # imports (verified against FAOSTAT 1961 overlap values), so label them
+  # directly. An earlier version mirrored the flows, which reversed direction.
   exports <- .read_input(
     "historical-trade-exports",
     years = years,
     year_col = "year"
   )
-  exports[, element := "import"]
+  exports[, element := "export"]
 
   imports <- .read_input(
     "historical-trade-imports",
     years = years,
     year_col = "year"
   )
-  imports[, element := "export"]
+  imports[, element := "import"]
 
   dt <- data.table::rbindlist(
     list(exports, imports),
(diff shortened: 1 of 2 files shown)


lbm364dl:
Superseded by #519, which rebuilds this on top of #477.

The two defects this PR identified are real and #519 fixes both — the swapped `element` labels and the discarded feed. The direction-swap evidence here was the most valuable part and is carried over (extended: global import/export imbalance goes 2.58x -> 0.98x on the 1950-55 build).

What is dropped is the polity-resolution half. This PR resolved `iso3` through the year-insensitive `.current_area_lookup()`; #477 has since established that these pins report under the borders in force at the time and must resolve to the polity of their own reported year via `.resolve_hist_trade_polities()`. Keeping this PR's version would have written 1840s British trade to `GBR-1921-2025` and 1830 Guadeloupe to a `ROW-1850-2023` bucket that did not exist then — the anti-pattern #458 names and #469 flags.

PR #349, opened 2026-07-24
feat(landuse): net temporary grassland out of arable footprint default (#342)

### LINKED ISSUE #342: Promote FAO arable-fallow land to footprint default; net out temporary grassland (CBS 3002) double-count
## Context

PR #125 added `build_fao_arable_fallow_extension()` (`method_land = "fao_arable_fallow"`): a fallow-inclusive per-crop physical-land method that reconciles arable-crop occupation to FAO's physical **Arable land** total. It was merged as a purely additive step — the footprint balance default was deliberately left unchanged.

This issue is the follow-up that makes the method actually count: promote it to the footprint-balance default. That is the whole point of the FAO-arable land base, but it cannot be done safely until a double-count is resolved.

## The double-count

FAO's **Arable land** definition includes *temporary meadows and pastures* — i.e. temporary grassland is part of arable land / cropland, not grassland (confirmed with the domain author). So `build_fao_arable_fallow_extension()` already absorbs the temporary-grassland slice into its arable-crop total.

Meanwhile `.land_balance_extension()` in `R/footprint_balance.R` builds the grass side from `build_grassland_land_extension(grassland_metric = "occupation")`, which returns **CBS 3002, Temporary grassland**, alongside CBS 3000. Today the crop side uses `cropgrids_fallow`, so there is no overlap. The moment the crop side is switched to `fao_arable_fallow`, temporary grassland is counted **twice**: once inside the FAO arable target, once as CBS 3002.

Intended invariant:

```
ordinary crop occupation (incl. fallow) + CBS 3002  =  FAOSTAT Arable land
```

## Provenance decision

whep's CBS 3002 is **modelled** (production item 996 → fodder reconstruction via FAOSTAT production, dry-matter yields, EuropeAgriDB, gap-filling), not the direct FAO "temporary meadows and pastures" RL series. The direct FAO series is patchy/inconsistent country-to-country, which is why it was modelled. **Decision: treat modelled CBS 3002 as authoritative** for temporary grassland.

## Work items

1. **Net CBS 3002 out of the arable target** in `build_fao_arable_fallow_extension()` — subtract modelled CBS 3002 hectares from `arable_ha` before reconciling ordinary arable crops, so `ordinary crops (incl. fallow) + CBS 3002 = FAO Arable land`.
2. **Switch the footprint-balance default** — replace `build_cropgrids_land_extension(source = "cropgrids_fallow")` with `build_fao_arable_fallow_extension()` on the crop side of `.land_balance_extension()` (`R/footprint_balance.R`).
3. **Reconciliation test** — assert that combined crop-plus-CBS-3002 occupation cannot exceed FAO Arable land for any `(area_code, year)`, so the double-count cannot silently reappear.

## References

- Follows up PR #125.
- Original review thread on #125 (double-counting of temporary grassland).

lbm364dl:
## What

Realises the point of #125: promotes `build_fao_arable_fallow_extension()` to the **crop-side default of the land-balance footprint** (`.land_balance_extension()`), replacing `cropgrids_fallow`, and closes the temporary-grassland double-count that switch would otherwise open.

Adds a `temporary_grassland` argument to `build_fao_arable_fallow_extension()`. When supplied (grassland-extension schema), its **CBS 3002** rows are netted out of the FAO arable target before ordinary crops are reconciled, enforcing per `(area_code, year)`:

```
ordinary crop occupation (incl. fallow) + CBS 3002 = FAO Arable land
```

`.land_balance_extension()` builds the grassland occupation once and passes it in, so the exact CBS 3002 it adds back on the grass side is the CBS 3002 netted out on the crop side — the two reconcile to FAO Arable land with no overlap.

## Why

FAO **Arable land** includes *temporary meadows and pastures* — temporary grassland is part of cropland, not grassland (confirmed with the domain author). Since the FAO-arable method reconciles all crops to that total, it already contains temporary grassland; the grassland extension reports the same land again as CBS 3002. Before this PR the crop side used `cropgrids_fallow` so there was no overlap; switching to the FAO-arable method (the whole point) made the double-count live. Netting fixes it.

## Provenance

whep's CBS 3002 is **modelled** (item 996 → fodder reconstruction), not the patchy direct FAO temporary-meadows series, so modelled CBS 3002 is treated as authoritative. Where modelled CBS 3002 exceeds FAO Arable land for a country-year (survey vs. reconstruction mismatch) the arable target is clamped at 0 and a warning is emitted.

## Design note for review

The `@AskUserQuestion` decision was to add a plain 6th argument. On `NULL` I made it a **no-op (no netting)** rather than auto-building the grassland extension from pins, because:
- it keeps `build_fao_arable_fallow_extension()` correct standalone (full arable target when CBS 3002 is not being added), and
- it keeps every pre-existing reconciliation unit test pin-free and unchanged.

The invariant is enforced where it matters — the footprint balance passes the grassland occupation explicitly. Happy to flip `NULL` to auto-build-and-net if you'd rather the extension be self-protecting by default.

## Validation

- 3 new reconciliation tests: netting reconciles crop + CBS 3002 to FAO arable; non-3002 grassland rows ignored; overshoot clamps at 0 and warns.
- `test_crop_land_extension.R`, `test_footprint_balance.R`, `test_footprint.R`, `test_footprint_paths.R`, `test_footprint_grazing.R` all green (FAIL 0 / SKIP 0).
- `air format`, `lintr`, `roxygen2`, pkgdown reference coverage all clean.

Closes #342.

### DIFF
--- a/R/arable_permanent_land.R
+++ b/R/arable_permanent_land.R
@@ -433,19 +433,23 @@ get_arable_permanent_land <- function(
 #' perennial base area is reported as an error because it cannot be reconciled
 #' without inventing a crop allocation.
 #'
-#' The default of [build_cropgrids_land_extension()] and the footprint balance
-#' are unchanged; this is an additive method.
+#' This is the crop-side default of the land-balance footprint
+#' ([build_land_balance_footprint()]).
 #'
-#' @section Temporary grassland (do not double-count):
+#' @section Temporary grassland (no double-count):
 #' FAO's **Arable land** total includes *temporary meadows and pastures* —
-#' temporary grassland is part of cropland, not grassland. This method therefore
-#' already absorbs the temporary-grassland slice into its arable-crop total. Do
-#' **not** combine it with CBS 3002 (`Temporary grassland`) from
-#' [build_grassland_land_extension()] on top, or that land is counted twice. The
-#' intended invariant is
-#' `ordinary crop occupation (incl. fallow) + CBS 3002 = FAO Arable land`.
-#' Netting modelled CBS 3002 out of the arable target and promoting this method
-#' to the footprint-balance default is tracked in issue #342.
+#' temporary grassland is part of cropland, not grassland. That land is also
+#' reported separately as CBS 3002 (`Temporary grassland`) by
+#' [build_grassland_land_extension()], so summing both extensions naively would
+#' count it twice. Pass that grassland occupation as `temporary_grassland` and
+#' its CBS 3002 is netted out of the arable target before reconciling ordinary
+#' crops, enforcing the invariant per `(area_code, year)`
+#' `ordinary crop occupation (incl. fallow) + CBS 3002 = FAO Arable land`. The
+#' land-balance footprint ([build_land_balance_footprint()]) does exactly this. Where
+#' modelled CBS 3002 exceeds FAO Arable land (survey vs. fodder-reconstruction
+#' mismatch) the arable target is clamped at 0 and a warning is emitted. Called
+#' standalone (`temporary_grassland = NULL`) no netting is applied and the arable
+#' crops reconcile to the full FAO Arable land.
 #'
 #' @param harvested Tibble of harvested area with columns `year`, `area_code`,
 #'   `item_cbs_code`, `harvested_ha`. If `NULL`, built from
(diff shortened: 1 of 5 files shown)


lbm364dl:
### Provenance verified empirically — modelled CBS 3002 confirmed as the right netting basis

The open question was whether to net using whep's **modelled** CBS 3002 (item 996 fodder reconstruction) or FAO's **direct** "Temporary meadows and pastures" series (FAOSTAT RL item **6633**). Rather than decide by assumption, I compared the two head-to-head (modelled grassland occupation vs FAO 6633, harmonised to polity `area_code`):

| | Modelled CBS 3002 | FAO direct 6633 |
|---|---|---|
| Year coverage | **1850–2019** | **2001–2023 only** |
| Country-years > 0 | 1,973 | 3,339 |
| …pre-2001 | **1,612** | **0** |
| Overshoots FAO arable | 27 / 863 (3.1%) | 0 / 4,237 |

Agreement in the 2001–2019 overlap (n = 361): **Spearman r = 0.86**, **median ratio modelled/FAO = 0.97**.

**Conclusion:** modelled CBS 3002 is the correct basis. FAO's direct series does not exist before 2001 and so cannot net temporary grassland across ~1,600 pre-2001 country-years (let alone the pre-1961 back-cast); where the two overlap the model reproduces FAO within ~3% at the median. So this PR keeps modelled CBS 3002 as authoritative, and the `NULL = no netting` / clamp-and-warn behaviour stays as is.

**Two reconstruction outliers surfaced** (tracked separately in #354, not blocking this PR): all 3.1% of arable-ceiling overshoots are **Ireland** (up to 1.7× arable — hence the clamp), and the aggregate overlap gap (sum ratio 0.58) is driven by **Germany**, where the model captures only ~10% of FAO's reported temporary meadows.

PR #364, opened 2026-07-27
Reconcile the GRAFS typologies and plot branches (#37 + #63)

lbm364dl:
Reconciles the two open GRAFS PRs — **#63** (GRAFS plot dataset) and **#37** (Typologies Spain) — onto current `main`, and fixes the defects that surfaced once the code could actually be run.

## ⚠️ Results changes

Two published quantities change. Both are corrections; the previous numbers were wrong.

**1. Manure N drops ~46% (80.1M → 54.8M MgN), and 2022–2023 stop being zero.**

`.calculate_n_soil_inputs()` overwrote the current `n_balance_ygpit_all` pin's `Excreta` with values from `n_balance_ygpit_all_old`, via `coalesce(Excreta_old, 0)`. That graft was *correct* when added on 2026-04-22 (`a7fbfc08`) — at that moment the `_old` alias pointed at the **newer** upload, so it was a hot-fix pulling corrected Excreta into a stale alias. On 2026-05-08 (`737c267e`) the pin was bumped but `n_soil_inputs_nue.R` was never touched, so the graft has been misapplied since.

Evidence the pin as shipped is right:

- Both pins satisfy their own identity `Synthetic + Excreta + Solid + Liquid + Urban + BNF + Deposition == N_input_std` to 1e-16.
- Against the independent `n_excretion_ygs`, the grafted values are physically impossible: 1860 Excreta of 345,363 MgN vs **74,054 MgN of total grazer excretion (4.7×)** and 126,854 MgN of total livestock excretion (2.7×); manure/excretion = 3.99. The shipped series tracks the independent data as livestock move indoors (1.38 → 0.90 → 0.47).
- Per-hectare grazing deposition on Dehesa: 6.7 kgN/ha shipped vs **17.0** grafted — unsustainable on Spanish rangeland.
- The graft broke the pin's own arithmetic by **+11.4%**, and since the old pin ends in 2021 it zeroed all Excreta for 2022–2023 (248,870 and 217,385 MgN).

Added `validation/n_balance_invariants.R` with three guards, any of which would have caught this: the pin identity, `manure N ≤ livestock excretion N` per year, and no year with zero Excreta.

*Known remaining issue, not fixed here:* even the shipped pin has manure/excretion > 1 before ~1965 (1.65 in 1860) — a real back-cast inconsistency, but 2.4× better than the graft.

**2. Feed N rises by 120,839 MgN (81,631,866 → 81,752,704) as aquaculture stops leaking.**

`alice/typologies` had dropped the aquaculture handling that `main` carries in `.add_feed()`. This was a lost merge, not a modelling decision: `8e2d601b` added it, that commit *is* an ancestor of `alice/typologies`, and `6e98f4fd` ("Updates from grafs_plot changes") is where it disappears. Restored, keeping alice's own correct addition of `Other_birds` to the monogastric list.

Restoring the denominator alone is inert (`livestock_rum = local_consumption × (feed/total) × (ruminant/feed)` — `feed` cancels), so this PR also merges **#288**, which adds the `livestock_aqua` destiny that carries the N. #288 was written against `main`'s functions; this branch has alice's rewrite of all three, so it is ported rather than applied — the residual is taken off her *scaled* `local_feed` and *gap-derived* `import_feed`, and threaded through her national logic as `feed_aqua` / `share_aqua` / `share_feed_aqua`.

## Reconciliation decisions

- **Convention: snake_case** (project style rule). The data layer is **alice's** `n_prov_destiny.R` / `n_soil_inputs_nue.R`, because her newer pins carry schema changes (`item_cbs`, `intake_MgFM`) that `main`'s frozen copy cannot read. `main`'s `example = TRUE` offline mode is grafted back on, so tests and roxygen examples still run without network.
- **Plot layer** is `main`'s `grafs_plot_df.R` / `input_output_plots.R` (tested, offline-capable), with alice's features ported in: per-ha normalization, N accumulation, livestock feed-origin and rum/mono split.
- Git's auto-merge produced an internally inconsistent mixed-case `n_prov_destiny.R`; it was not trusted.

## Portability fixes

These functions previously only ran on one machine:

- **Natural Earth provinces shapefile** was a hardcoded `C:/PhD/...` path in four functions. Now `.provinces_shapefile()` resolves an explicit `shapefile_path` → `options(whep.provinces_shapefile)` → local cache → on-demand download from the **official Natural Earth CDN** (the same source `download_naturalearth.R` already uses for admin-0). Not redistributed, so attribution stays upstream.
- **GRAFS Spain inputs** were a hardcoded `inputs_dir`. Now an argument → `options(whep.grafs_inputs_dir)` → an abort listing the required files. Also fixed a case bug that only breaks on Linux/macOS: the code asked for `n_inputs_combined.csv`, the file is `N_Inputs_combined.csv`.
- **10 `ggsave()` calls** wrote to hardcoded absolute paths. Saving is now opt-in via `out_dir = NULL`; the plot is returned as before and a file is written only when asked.
- **`codes_coefs`**: three reads of `Codes_coefs.xlsx` replaced with the pin, verified identical (233 rows, same 4 columns, all 230 `Name_biomass` values in common). `codes_coefs_items_full` is a *different* table and would have been the wrong substitute.

## Bugs found by running the pipeline for real

Fixtures and mocks could not catch these — they only appear against the live pins:

- `create_grafs_plot_df()` errored: `n_excretion_ygs` reports `N_excr_MgN` directly, not the `Gross_Prod_GgN − Net_Prod_GgN` pair.
- `calculate_nue_livestock()` errored twice: `intake_MgN` not `n_mgn`; `item_cbs` not `item`.
- `plot_input_output_livestock()` errored: same `item_cbs` vs `Item` mismatch.
- `circ_nue_traj_plot()` errored on a stray PascalCase `LandUse` in a file that lowercases `npp_ygpit` everywhere else.
- `.livestock_prod_type_map()` **silently discarded 0.59% of livestock production N** — `Other_birds` is in the pin but was in neither the ruminant nor monogastric list, so it fell through to `NA` and was filtered out. Classified as monogastric, consistent with `.add_feed()`.
- `whep_typologies_spain.R` called `create_n_prov_destiny()$prod_destiny`, which is `NULL` since that function returns a tibble; and `alfredo_typologies.R` referenced a `box_destiny` column the data layer does not produce. Both fixed; the `box_destiny` categories were reconstructed from the authoritative historical definition found in `e6bfd2b2`.
- `Typologies_Julia.R` could never complete: `.calculate_imported_feed_share()` selected `LU_total` while **both** joined frames carried it, so dplyr suffixed to `.x`/`.y`.
- `whep_read_file()` failed with a bare `invalid 'path' argument` when a pin lacked the requested format. It now names the missing format and what is available.

**All 13 typology entry points and the full N pipeline now run on real data**, including `create_typologies_grafs_spain()` (50 provinces) and `create_typologies_of_josette()`, which previously could not run at all.

## Housekeeping

- Dropped the legacy `grafs_prod_destiny` pin — superseded by `create_n_prov_destiny()` and read by nothing. Also removed its vestigial `globalVariables` entries, which named elements of a list return that no longer exists.
- Removed the last renv remnants that survived #133 (`RENV_CONFIG_PAK_ENABLED`, and `USE_RENV` in four workflows where nothing reads it).
- Fixed a non-ASCII en dash in a plot title (a portability WARNING) and `sf` being declared in both Imports and Suggests.
- Registered the 25 undefined globals R CMD check reports on `main`, so this branch's check is clean rather than merely no worse.

## Verification

| Gate | Result |
|---|---|
| `R CMD check` | 0 errors, 0 warnings, 0 notes |
| Test suite | 3,444 pass / 0 fail / 0 error |
| `lintr` | 0 |
| `air format` | clean |
| pkgdown | every `man/*.Rd` present in `_pkgdown.yml` |

`R CMD check` was run on a clean `git archive` export — an in-place build fails because it copies the gitignored 37 GB `LPJmL_inputs/`.

## Still needs Alice

One open question:

- Whether `Model inputs/GRAFS_spain` is the data vintage her thesis numbers assume. `Typologies_Josette` / `Typologies_Julia` now run against it, having previously been unable to run at all.

## Note for reviewers: imported feed N is now read, not estimated

`whep_typologies_spain.R` and `alfredo_typologies.R` used to *estimate* how much of an item's imports went to feed: take the item's total imported N, then multiply by the share of that item's consumption that goes to feed.

That estimate is no longer possible, and no longer needed. It requires rows with `destiny == "import"`, and the data layer has **zero** of those — imports are represented as `origin == "Outside"` with the destiny already resolved (`population_food`, `population_other_uses`, `livestock_rum`, `livestock_mono`, `livestock_aqua`). The old path therefore silently returned 0. Both functions now sum the pre-split feed rows directly.

The resulting numbers differ from the old estimate by design: the data layer allocates imports by unmet demand gaps rather than by average consumption mix, so imports are pushed toward feed where local production already covers food demand. For scale, 59.3% of imported N (33,560 of 56,558 Gg N) goes to feed.


AliceBeckmann:
Hi @lbm364dl , thank you for reviewing this!
Regarding the question about the GRAFS_spain input: I think this dataset as a source for Typologies_Josette/Typologies_Julia should not be necessary to use. Instead the function "create_n_soil_inputs()" should be called, since it is creating all the necessary N soil inputs, which are found in GRAFS_spain. Can we change that by calling the function, instead of reading this file from inputs_dir?

R/n_soil_inputs_nue.R
```diff
@@ -111,7 +111,7 @@ create_n_soil_inputs <- function(example = FALSE) {
       deposition = sum(Deposition, na.rm = TRUE),
       fixation = sum(BNF, na.rm = TRUE),
       synthetic = sum(Synthetic, na.rm = TRUE),
-      manure = sum(Solid + Liquid, na.rm = TRUE),
+      manure = sum(Excreta + Solid + Liquid, na.rm = TRUE),
```
AliceBeckmann, line 114: Yes that’s the right fix, I already removed the same n_balance_ygpit_all_old / Excreta_old before seeing this PR, so that should be correct.

lbm364dl:
> Hi @lbm364dl , thank you for reviewing this! Regarding the question about the GRAFS_spain input: I think this dataset as a source for Typologies_Josette/Typologies_Julia should not be necessary to use. Instead the function "create_n_soil_inputs()" should be called, since it is creating all the necessary N soil inputs, which are found in GRAFS_spain. Can we change that by calling the function, instead of reading this file from inputs_dir?

Thanks! I'm fixing it now. Just wondering, is the GRAFS_spain folder then something old that can be removed from SACO? I'm worried about having these confusions again in the future

lbm364dl:
The idea is you confirm numbers look correct, or otherwise you can fix yourself and push to this PR. If something is not urgent, we create a separate issue for that. Also check if you had some more uncommitted work in your old branches that you might want to add here.
Below text is generated by Claude:

---

Following your comment I looked at what `Model inputs/GRAFS_spain` actually is, and it is not a sanctioned data source: it has no `_pins.yaml`, so it is not a pins board, it carries no versioning or integrity metadata, and nothing in the package has ever referenced it. `whep_read_file()` only reads the two real boards, `nitrogen_spain` (18 pins) and `world` (46). So the folder is an unmanaged parallel copy, which is why its contents diverge from the boards inconsistently — some files newer, some only renamed, some identical. Nothing was keeping it in sync. My `inputs_dir` argument was formalising that side-channel, which was the wrong call on my part.

So rather than only swapping the N inputs, **both typologies now take every input from a pin or a package function**, and the `inputs_dir` machinery is deleted (`R/grafs_spain_inputs.R` removed, argument dropped from both exported functions).

### What each input became

| was | now | equivalence |
|---|---|---|
| `N_Inputs_combined.csv` | `create_n_soil_inputs()` | your suggestion — and it matters: the CSV is a frozen snapshot still carrying the manure numbers from the stale `Excreta` graft this PR removes, so reading it left `Manure_share` on pre-correction data |
| `Biomass_coefs.xlsx[Coefs]` | `whep_read_file("biomass_coefs")` | identical (421×41) |
| `Codes_coefs.xlsx[items_full]` | `whep_read_file("codes_coefs_items_full")` | identical (199×15) |
| `Codes_coefs.xlsx[Liv_LU_coefs]` | `whep_read_file("livestock_units")` | identical LU coefficients — `LU_total` is 1,842,258,700 before and after |
| `Livestock_Prod_ygps.csv` | `whep_read_file("livestock_prod_ygps")` | equivalent for the columns actually used |
| `NPP_ygpit.csv.gz` | `whep_read_file("npp_ygpit")` | **not identical, see below** |
| `PIE_FullDestinies_FM.csv` | `whep_read_file("pie_full_destinies_fm")` | **not identical, see below** |
| `GRAFS_Prod_Destiny_git.csv` | `create_n_prov_destiny()` | recomputed rather than frozen |

Because `livestock_units` already carries the `Livestock_cat` mapping, the hardcoded `.create_livestockcat_mapping()` tribble in `Typologies_Julia.R` was redundant and is deleted — verified by `LU_total` being unchanged to float noise.

The `GRAFS_Prod_Destiny_git.csv` translation lives in one place, `.grafs_prod_destiny_legacy()`, which maps the new vocabulary back to the legacy one (`population_food → Food`, `livestock_rum|mono|aqua → Feed`, and so on) so none of the downstream helpers needed touching. Imports are emitted both under their resolved destiny and as a parallel `Destiny == "Import"` row, because that is the legacy convention the code depends on: `Production = Food + Feed + Other_uses + Export − Import` only holds if the use rows are gross.

### Numbers change — please sanity-check these

Two of the pins are genuinely newer data, not copies. Over the **common 1860–2021 window** (so this is revision, not just the two extra years the pins add):

- `npp_ygpit`: cropland `Prod_MgN` **−6.21%**
- `pie_full_destinies_fm`: feed imports **+19.09%**

Downstream in Julia: crop productivity **−10.8%**, `Semi_nat_share` **+13.2%**, `Imported_feed_share` **+26.2%**, `Domestic_feed_MgN` **−28.8%**.

Despite that, **all 50 provinces keep their 1980 typology** (37 `Extensive cropping system`, 13 `Specialized livestock-farming system`). That is margin luck rather than evidence of equivalence — the shifts happen not to cross the 0.4 LU/ha, 60 kgN/ha, 0.6 and 0.5 thresholds. Worth knowing that the classification is closer to those boundaries than it looks.

**Josette does change**: 38 of 50 provinces unchanged in 1980; the 12 that move are Alicante, Araba, Bizkaia, Ciudad_Real, Cordoba, Guadalajara, Illes_Balears, Jaen, Leon, Soria, Valencia, Zaragoza. Across all year×province cells, 48.7% keep the same typology set. It also stops emitting a phantom `"Sea"` province (51 → 50) and now runs to 2023.

### One pre-existing bug, not from this PR

`.calculate_feed_domestic_share()` returns several rows per `Year` × `Province_name`, so `distinct(Year, Province_name, Typology)` can yield **two different typologies for the same province-year**: 6,349 of 8,100 cells before this change, 2,723 after. Something to fix separately — happy to open an issue if you'd like.

AliceBeckmann:
Thanks for the changes @lbm364dl!

Numbers check: I think the changes in numbers, do make sense because many datasets were updated since the last use here.

The .calculate_feed_domestic_share() bug: Claude fixed this directly, so no need to open an issue. feed_summary was never aggregated over Item before being joined to province-level LU shares by Year alone, so every province-year got fanned out once per item, leading to the double counting. This got fixed. 

Aquaculture: I don't think livestock_aqua should be a destiny here. Fish/seafood items (Box == "Fish") are always Origin == "Outside" in this dataset (so imported from the sea), so a domestic aquaculture feed-consumption category doesn't have anything to pair with. Since the numbers are small, I got rid of this category for now, but I will open an issue.

SACO: I think Model inputs/GRAFS_spain can be removed, since nothing in the package reads it anymore and every pin the code now depends on is already registered in nitrogen_spain.

Other branches: I do have some uncommitted work (LMDI decomposition, circularity index, national trade validation, sensitivity analysis) in alice/typologies. I'll bring it in as a separate PR once this merges.

AliceBeckmann (approved):

PR #368, opened 2026-07-27
fix(lpjml): extend forcing to 2023 and derive climate in-pipeline (#340)

### LINKED ISSUE #340: LPJmL hydrology data effectively ends 2009; the 1901-2018 run's seepage is corrupt (do not use it for drainage)
## Summary

Any LPJmL-derived term for year **2010 or later** (drainage via `build_water_balance()`, NPP via `read_lpjml_npp()`, SOC climate via `get_soc_climate_drivers()`) silently reads **all-NA** for those years, because every available LPJmL run's output effectively ends in **2009**. One run *looks* like it covers 2010+ but its data there is missing, and its deep-seepage output is physically corrupt. Callers should treat **2009 as the latest usable LPJmL year** and use it as a proxy for 2010, and must **not** switch `WHEP_LPJML_RUN_DIR` to the 1901-2018 run for drainage.

## Evidence (real 2009/2010 reads via `read_lpjml_hydrology("drainage", ...)`)

Two runs under the LPJmL runs directory (`WHEP_LPJML_RUN_DIR`, `.../output/scenario_1`):

| run | seepage 2009 (median / p90 / p99 / max, mm) | year 2010 |
|---|---|---|
| `..._grassland_livestock_npp_vegc_fix` (1901-2009, the NCW-validated default) | 37.9 / 612.8 / 1564.7 / 4832.7 | not present |
| `global_1901-2018_spinup_200_our_inputs` | **0 / 0 / 3215 / 87347.3** | **all NA (0 finite of ~68k cells)** |

- The `npp_vegc_fix` run's seepage is clean and physically plausible (deep drainage a few hundred mm, max ~4.8 m over a full year).
- The `1901-2018` run's seepage is **corrupt**: median 0, and a max of **87,347 mm (≈ 87 metres of drainage in one year)**, which is impossible. Its `.nc.json` metadata claims `firstyear 1901 / lastyear 2018 / nyear 118`, but the actual values for 2010 onward are all NA (verified by direct read; `read_lpjml_hydrology("runoff", years=2010)` is likewise all-NA).

`build_water_balance()` built with the 1901-2018 run therefore returns `drainage_mm` = NaN / a few thousand garbage cells; built with the `npp_vegc_fix` run (year 2009) it returns a clean field (mean ≈ 209 mm, ~40k of ~68k cells draining).

## Why it is a trap

- The metadata (`lastyear 2018`) makes the 1901-2018 run look like the fix for the "no 2010 data" wall. It is not — the data there is empty and, where present, the seepage variable is corrupt.
- The failure is **silent**: the reader indexes month `(year - first_year) * 12 + month` and returns NA for out-of-range/empty steps, so a full-year aggregation yields NaN drainage rather than an error. Downstream leaching (`calculate_n_leaching(method="meisinger_drainage")`) then gets NaN/garbage drivers.

## Suggested actions

1. **Document** that LPJmL-derived terms are only valid through 2009, and that 2009 is the intended proxy for a 2010 analysis (drainage, NPP, SOC climate).
2. **Do not** point `WHEP_LPJML_RUN_DIR` at `global_1901-2018_spinup_200_our_inputs` for hydrology; prefer `..._grassland_livestock_npp_vegc_fix`.
3. Consider a **sanity check on load** in `read_lpjml_hydrology()` / `build_water_balance()`: warn (or abort) when a requested year's field is all-NA, or when annual seepage exceeds a physical ceiling (e.g. > precipitation, or a hard cap like ~10,000 mm), instead of propagating NaN silently downstream.
4. If a genuine post-2009 run is needed, the 1901-2018 run must be **re-checked / re-generated** — its seepage output is unusable as delivered.

lbm364dl:
Closes #340 by removing the 2009 ceiling on LPJmL-derived terms (drainage, NPP, SOC climate) rather than proxying around it, and by removing the mechanism that would have silently reintroduced it.

## Root cause

Not corrupt output. Both global runs were forced with **CRU TS 3.10, which ends in 2009**, while `global_1901-2018` set `lastyear = 2018`. Climate is a *hard* LPJmL input — `readclimate()` aborts with `ERROR130` rather than holding the last year constant — so 2010+ could never exist, and that run's impossible `mseepage` (max 87 347 mm) is an artefact of the aborted write, not a bad variable.

Verified in the LPJmL source that inputs split two ways, which is what makes 2023 reachable at all:

| class | inputs | behaviour past end |
|---|---|---|
| **hard** | temp, prec, cloud, wetdays, wind, co2 | `ERROR130`/`ERROR131`, no fallback |
| **clamped** | landuse, sdate, crop_phu, fertilizer, manure, tillage, residues, popdens, ndep | `checkyear()` → `WARNING024`, holds last year |

Only the hard inputs bind. `tamp`, `wateruse` and `burntarea` are never read in this configuration (`fire = "fire"`, `wateruse = "no"`), which is why a dead `/p/projects/...` path never broke anything.

## What changed

| forcing | before | after |
|---|---|---|
| CRU TS | 3.10 → **2009** | **4.09 → 2024** |
| CO₂ | 1765–2018 | **1765–2025** (NOAA tail, offset-corrected onto the series) |
| wind | gswp3-w5e5 → 2016 | **1901–2023** (ERA5 tail, bias-corrected per cell and calendar month against the 2017–2019 overlap) |
| popdens | HYDE3 → 2011 | **ISIMIP3a → 2021** (per-cell counts → people/km², per-latitude spherical cell areas) |

`simulation_end_year` is now **2023**.

⚠️ **popdens is wired but inert under the current config.** `initinput.c` reads it only `if(config->ispopulation)`, and these runs set `population = FALSE`. The conversion is correct and validated, but it changes nothing until human ignition is switched on. Reviewers should not expect it to affect results.

### Climate is derived in-pipeline, not symlinked

Section 10a used to **symlink** prepared climate files under hardcoded 3.10 filenames. That was the live trap: a pipeline re-run would have relinked 2009 forcing under a config expecting 4.09. Climate is now **Section 9d** (`prepare_climate_inputs()`), deriving real files from raw downloads like every other input. Together with Section 9c (popdens) and `download_climate()`/`download_population()`, the chain is automatic — `download_all()` then `prepare_spatialize_all()`, no manual steps.

### Two pinned wind artefacts

Neither is rebuildable from this repo, so both are pinned and fetched by `download_climate()`:

- `lpjml-wind-isimip-1901-2019` — monthly base, assembled by a script that was never committed
- `lpjml-wind-era5-2017-2023` — ERA5 monthly means (~85 GB of streaming to reproduce)

`whep_read_file()` gains an `nc`/`nc4` type returning the path, since these grids are read lazily.

### CRU land-mask gap

CRU misses ~20 of LPJmL's 58 795 grid cells (Arctic and small islands, Ladoga, the IJsselmeer) and LPJmL aborts on the first with `ERROR423`. The 3.10 files hid this by being filled globally with ocean set to 0; filling land with 0 °C is wrong, so these are gap-filled from the nearest valid CRU cell by great-circle distance.

## Validation — three runs, 8 MPI tasks

**1901–1903** against the `lakefix` baseline:

| | mine (CRU 4.09) | baseline | diff |
|---|---|---|---|
| GPP | 106.702 | 106.509 | +0.18% |
| NPP | 57.7959 | 57.751 | +0.08% |
| RH | 58.3419 | 58.2618 | +0.14% |
| transp | 38.2171 | 38.0344 | +0.5% |
| **firec** | 0.240539 | 0.268771 | **−10.5%** |

The `firec` gap is the CRU revision propagating through the **Nesterov fire danger index**, which keys on temperature and precipitation — not popdens, which is not read here. Underlying data agree: July 1901 precipitation over 62 415 shared cells is 78.53 vs 78.65 mm.

**2005–2012** produced the years no existing run can:

| year | finite cells | seepage median | max |
|---|---|---|---|
| 2010 | 58 795 / 58 795 | 255.7 mm | 5071 mm |
| 2011 | 58 795 / 58 795 | 236.2 mm | 5722 mm |
| 2012 | 58 795 / 58 795 | 244.9 mm | 4303 mm |

against the corrupt run's **0 finite cells, median 0, max 87 347 mm**.

**2016–2023** exercises what the first two could not — the ERA5 wind tail, the CRU 2013+ tail, and the clamp path (the first two smokes both used the pre-ERA5 wind file and stopped at 2012). Result: exit 0, 58 795 cells, no `ERROR130`/`ERROR131`; ndep warns `WARNING024` for 2022–2023 and continues. No discontinuity at the 2019/2020 wind joint:

| year | seepage median | draining cells |
|---|---|---|
| 2019 | 261.6 mm | 35 069 |
| **2020** (first ERA5 year) | 276.1 mm | 35 407 |
| 2023 | 231.2 mm | 35 918 |

`.check_climate_coverage()` passes for 1901–2023, and Section 9d regenerates the validated forcing **bit-identically**.

## Caveat for any write-up

A 2023 run is **native climate with management held constant** — tillage after 2010, residues after 2015, ndep after 2021. Those are the last years the source datasets cover at all (PIK ships the identical tillage and residue files; ISIMIP3a ndep also stops at 2021), so there is nothing newer to adopt. This qualifies the "removes the proxy entirely" framing in the issue.

## Known debt

Pinning the 1901–2019 wind base papers over the fact that no code rebuilds it from the ISIMIP2a chunks. Worth a follow-up issue.


lbm364dl:
New run should be fine now and can be found in [SACO](https://saco.csic.es/s/nrJ3JGPZyZeQMW8?dir=/Sources/datasets/unclassified_datasets/LPJmL/LPJmL_runs/global_1901-2023_spinup_200_our_inputs).

Issue #207, opened 2026-07-22
energy_co2_extension: countries missing from GLEAM crosswalk get NA intensity and are silently dropped; NA emission factors silently become zero


lbm364dl:
## What

`.energy_co2e_by_group()` (`R/energy_co2_extension.R:268-271`) left-joins intensity by iso3; an `area_code` whose iso3 is absent from `gleam_geographic_hierarchy` yields `ef_total = NA` → `co2e_kg = NA` → NaN/0 after allocation → removed by the `impact_u > 0` filter with no message. Separately, `.energy_mean_factor()` (`:232`) uses `mean(emission_factor)` without `na.rm`; a single NA factor makes the grouping's ef NA, which `coalesce(..., 0)` (`:211`) converts to a silent zero — understated, not missing.

## Suggested fix / check

Warn on countries with no GLEAM crosswalk match; use `na.rm = TRUE` (or validate) in `.energy_mean_factor`, and distinguish genuine zero from missing before coalescing.

---
_From deep static audit (2026-07-22), parallel subsystem review pass. Deduplicated against then-open issues. Not verified by running unless stated._

lbm364dl:
## Adopted as a child of the polity migration epic #458

A subsystem audit re-derived the work this issue describes. Its spec is below so this issue can be worked without re-deriving it. Keeping this issue rather than filing a duplicate.

---

### Port #382's GLEAM-classification-gap warning; tidy `.energy_area_iso3()`

**Problem**
`gleam_geographic_hierarchy` has no row for 5 live present-day polities — Bermuda, Guam, Nauru,
Palau, Tuvalu — so `.energy_country_grouping()` silently drops them (`ef_total = NA`, filtered out
with zero visibility). Tuvalu is explicitly asserted as LDC (`R/energy_co2_extension.R:146`) yet is
unclassifiable — an internal contradiction currently invisible. Separately,
`.energy_area_iso3()` (`R/energy_co2_extension.R:330-336`) reuses `.current_area_lookup()` — a
function built to pick one "best current" polity per raw area_code for a different purpose — purely
to extract `area_iso3c`, riding on an undocumented invariant (verified: iso3 is invariant across
polities sharing an area_code for all 99 multi-polity codes).

**Evidence**
- `R/energy_co2_extension.R:146, 155-193, 237-242, 268` — silent-drop chain.
- `R/energy_co2_extension.R:330-336`, `R/polities.R:11-35`.

**Target state**
[...]

PR #382, opened 2026-07-28
polities: link the package to whep-polities and stop resolving to dead polities

lbm364dl:
## Start here

**What this does.** Links the package to `whep-polities` as a live source of truth rather than a stale embedded snapshot, and routes every area/country/region identity through it instead of through hand-maintained copies. 165 files, ~14.5k lines, most of it tests and the reasoning behind them.

**What is verified.** 14 local gates (5,188+ tests, 0 fail; 133 upstream contract tests), R CMD check green on five platforms, and the committed `data/` **regenerates byte-for-byte** from the published contract — so a reviewer can rebuild rather than trust. Current `main` is merged in, including #429 and #427.

**RETRACTED: "it changes no published numbers".** That claim stood here until 2026-07-31 and is false. It was measured *before* #425 was fixed, when every commodity-balance value was a row count — and row counts compare group **sizes**, not quantities, so the measurement was structurally incapable of detecting a value change. Re-measured on the fixed build against `main` at `3f479f7a`: **`feed` +7.0%, `other_uses` +3.1%, `production` +2.0%, `domestic_supply` +1.9%**, four columns outside the 1% band. Isolated as roughly one third code and two thirds data, and localised to FABIO's rest-of-world bucket (`feed` there goes up 28.7x, which is 78% of the total move). Cause **not yet identified**; five hypotheses tested and rejected. **Tracked in #444, which is the thing to read before merging this.**

What DOES hold, re-measured on the fixed build: **195 areas** in both, the numeric aggregation key **identical for all 267 area codes**, and the supply-use imbalance count **47,514 — identical to the row across three independent builds** (main, branch, and branch-code-with-main-data). So the structural integration is neutral; the arithmetic is not.

The likely benign reading is that the branch's `area` vocabulary lets rows survive joins that `main` silently drops — the branch has *fewer* rows but *more* value, which is what including previously-dropped members looks like. If that is what it is, it is an improvement and should be documented as one. It is not neutrality, and the original framing — that this can merge ahead of #414/#419/#426 *because* it moves nothing — does not survive.

**A reviewer's shortest path**, if the diff is too large to read linearly:

| to check | look at |
|---|---|
| the resolver and its guards | `R/polities.R`, `tests/testthat/test_resolve_polity_label_name_route.R` |
| what is guaranteed and by which test | the "What is guaranteed" table in `vignettes/polity-integration.Rmd` |
| what is *not* fixed, and why | the "Where the gaps are" section of the same vignette |
| that the data is regenerable | `Rscript inst/scripts/check_like_ci.R` — gate "data regenerates from the contract" |
| **that no polity is named in logic** | the same script's gate "no polity code in package logic" — 0 literals across 122 files, which is this PR's central claim made executable |
| that a magnitude cannot drift unnoticed | `inst/scripts/compare_cbs_totals.R` |

**Not in scope, and filed rather than fixed.** The work surfaced one live correctness defect and several modelling questions, none of which this PR resolves:

- **#425 — FIXED and merged (#429).** Every commodity-balance cast value was a row count. Re-recording this branch's totals baseline on the fixed build reproduced the predicted magnitudes to three digits: **food 259.00x, seed 182.69x, feed 11.89x, import 7.09x**, rows 2,768,578 → 1,999,609. The 28% row drop is the fix working — counts are never zero, so the `value != 0` filter had nothing to remove until quantities came back. This branch now carries the fix, and the duplicate-key warning it added has been repurposed rather than kept: with `fun.aggregate` present the `length()` fallback cannot fire, so the guard now warns only when duplicates appear in a bucket that folds no other area.
- **#426 — destiny shares divide by zero. My attribution here is RETRACTED.** I wrote that it was largely downstream of #425. Re-measured after the fix: the contamination is gone from that frame (`production == 1` fell from 37.4% to 0.000%, 5 rows of 1,893,136) and the symptom grew tenfold — **45 `Inf`, 9,167 `NaN`, 689 shares above 1**, against 6, 926 and 80. #425 was *masking* it: row counts are never zero, so `x/0` was unreachable while every value was a count. It is independent and larger than filed.
- **#420 — the build is not reproducible.** Re-measured on the fixed frame, and two things changed. It is **intermittent, not alternating** (one build in four diverged, by 4.8e-7 on `production`, alongside a verified-equal pair), and it occurs **within a session**, which I had previously reported as impossible — the integer counts were quantising the wobble away, so the build looked deterministic because its output had no resolution. Row counts are now stable within a session, so the residual variation is in values, not in which rows survive. Localisation to `.fill_historical_destinies()` is unaffected.
- **#414 / #419** — the label side of #425: which polity a folded bucket should carry, and whether to fold at all. #414 comes with #425's fix, because correcting the values makes the labels wrong.
- **#414 — now LIVE rather than latent.** With #425 fixed, bucket 206 post-2011 publishes a genuine Sudan + South Sudan sum under the Sudan-only code `SDN-2011-2025`. Exactly three buckets fold at all (derived from the crosswalk, not assumed): 206, 238 and 999. 238 is benign — Ethiopia PDR and Ethiopia never report in the same year — and 999 names its aggregate honestly. 206 is the only fold where summing and labelling disagree.
- **#444 — the value movement above.** New, and the blocker for this PR's merge argument.
- **#405, #421, #424, lbm364dl/whep-polities#55** — smaller, each costed in the issue.

**One caveat about the narrative below.** Two changes in this PR were withdrawn after measurement, and my first explanation of one of them was wrong — I attributed a 266× inflation to summing folded areas when it belonged to a bundled relabelling of a join key. The text below has been corrected, but it is long and was written incrementally; where a figure matters, the issues carry the isolated measurement.

---

> [!NOTE]
> **The row-count blocker is resolved, and the resolution reversed two of this PR's own changes.**
>
> The 702,166-row gap in the shared areas was mine: relabelling the aggregator's output with the
> bucket's stable `area_name` created two vocabularies for `area`, which is a JOIN KEY -- four inner
> joins use `c("area", "area_code")` -- while `.read_crop_residues()` kept emitting the polity name.
> The rows did not match and were dropped. Both halves of that change are withdrawn:
>
> 1. **Summing the areas that fold into one bucket.** Withdrawn because it was bundled with a
>    relabelling of the `area` join key that alone dropped 702,166 rows -- **not** because summing
>    is wrong. I originally reported it as double-counting (`food` **266x**) and that attribution
>    was mine and incorrect: the 266x belongs to the relabelling. Isolated, summing removes the
>    cast's duplicate keys outright and matches the cast-side fix to within a few percent, so it is
>    one of two candidate fixes in **#425**. The PR keeps the previous grouping pending that
>    decision, and the `dcast` guard added here WARNS rather than aborting.
> 2. **Promoting FABIO's rest-of-world members to their own numeric key.** `feed` **13.7x**,
>    `export` 13.2x, `production` 1.8x, with the whole `feed` increase landing on ONE area -- 212
>    Syria at 1.049e12, twelve times the world total. #419.
>
> Both were found by diffing every column TOTAL against main, which nothing in the 5,179-test suite did. Row
> counts barely moved while values exploded, so the suite was silent. Isolated by swapping datasets
> one at a time: **branch code + main data reproduces main's totals exactly**, so no R change here
> moves a number -- one expression in `data-raw/table_mappings.R` did all of it.
>
> **Where the PR now stands:** the numeric aggregation key is **identical to main for all 267 area
> codes**, the area set is main's **195**, every column total is **within 1%** of main, and the
> supply-use imbalance count is **46,424 -- identical to main to the row**, so that open caveat is
> resolved rather than hedged. Full range: ~2.766M rows, zero `NA` in all four polity columns. The count is given loosely because the build is not reproducible -- it alternates between 2,768,578 and 2,764,471 on an unchanged tree, #420 -- so any claim resting on a few thousand rows is inside its own noise.
>
> The PR is therefore **value-neutral by construction**, and the two modelling questions it uncovered
> are costed on both sides in #418 and #419 rather than decided here. Folding costs 0.21% of livestock
> feed demand (no Bouwman region on RoW); promoting costs a 13.7x feed error. That trade is the
> maintainer's call.

Links the package to `whep-polities` as a real source of truth rather than a stale snapshot, and stops area resolution landing on polities that no longer exist.

Paired with **lbm364dl/whep-polities#39** — that side publishes a contract, this side consumes it. The two are
interconnected by construction and several changes here required a matching change there:

| finding | published side (whep-polities#39) | consuming side (this PR) |
|---|---|---|
| generic short forms unreachable | four aliases rescoped `source=faostat` → blanket | `add_area_code()` routes unmatched names through the database |
| a bound neither matcher read | `matchlib` treats a missing bound as unbounded on that side | `resolve_polity_label()` does the same |
| deliberately-unmapped areas | 351 published in the manifest | the name route refuses names of such areas |
| `TAN-1922-1964` code/column disagreement | baselined by `validate_code_year_agreement.py` | its cost measured: the last 200 unmatched residue rows |

Merge order: whep-polities#39 first, since this side's tests assert against the contract it publishes.

## The problem, in numbers

`data/polities.rda` had drifted from `~/whep-polities/data/final/polities_database.csv` with nothing detecting it:

| | |
|---|---|
| rows in the package | **603** |
| rows upstream | **740** |
| missing from the package | **144** (`ALK-1867-1959`, `AOI-1936-1941`, the `ARG-1899-1902`/`ARG-1902-2025` split, `ADE-1839-1963`, …) |
| in the package but gone upstream | **7** (`CAN-1866-1948`, `PAK-1800-1947`, `ZAF-1828-2025`, `HUN-1919-1920`, …) |
| **retired or superseded upstream** | **24** |

That last row is not staleness, it is **wrong output**. 24 crosswalk rows routed real FAOSTAT area codes to withdrawn polities:

- **area 21 Brazil** → `BRA-1800-2025`, a row superseded upstream when the 1903 Treaty of Petrópolis (Acre, ~125,568 km²) forced a three-way split
- **area 100 India** → `IND-1800-1893`, superseded by a split at the 1886 Upper Burma annexation
- **area 101 Indonesia** → both `IDN-1800-1889` and `IDN-1889-1945`, merged upstream because the 1889 boundary had no territorial basis

So the package both attributed data to rows that no longer exist **and hid the period splits that replaced them**. Upstream territorial work was not reaching the analysis at all.

## Changes

**1. Exclude dead polities from resolution.** `wiki_status` of `retired` or `superseded` means the row must never receive data — `whep-polities` enforces exactly this in its own matcher (`matchlib.Matcher.DEAD_STATUS`). The rows stay in `polities` for provenance, since a reader may need to know what a historical code meant, but are dropped from `polity_attrs` and therefore from the crosswalk. 26 excluded, reported at build time via `cli`.

**2. Map FAOSTAT Djibouti (72) to the `FRS` prefix.** A consequence of change 1 that had to be chased: `DJI-1886-2025` was retired upstream as a duplicate of the FRS chain, and it was area 72's *only* mapping — so excluding it left the area unmapped. It now resolves to `FRS-1884-1977` (French Somaliland) and `FRS-1977-2025` (Djibouti).

## After

```
polities.rda 740 = 740 upstream, 0 missing
crosswalk rows targeting a dead polity: 0  (was 24)
unmapped areas: 1  (351 China — no iso3c, no WHEP polity; pre-existing)
```

and the splits now flow through:

| FAOSTAT area | resolves to |
|---|---|
| 21 Brazil | `BRA-1800-1903`, `BRA-1903-1909`, `BRA-1909-2025` |
| 100 India | the full seven-period chain, incl. `IND-1800-1886`, `IND-1886-1893` |
| 101 Indonesia | `IDN-1800-1945` (merged), `IDN-1945-1949`, `IDN-1949-1969`, the three 1949-51 island groups, … |

## Still to do on this branch

- **A staleness gate.** Nothing yet *prevents* the 603-vs-740 drift recurring. Intended: a test asserting the embedded table matches the upstream manifest the paired PR will publish, so drift fails loudly instead of being discovered by inspection.
- **`inst/extdata/harmonization/regions_full.csv`'s `polity_code` column is misnamed.** It holds bare ISO3 strings (`AFG`, `AGO`, …), so all 206 values are unresolvable *as polity codes*; the R side correctly treats them as **prefixes** (`polity_prefix = sub("-.*", "", polity_code)`, `known_polity_prefixes`). Renaming it to `polity_prefix` removes a real trap — a prefix cannot identify a territory, which is the entire point of a periodized polity database.
- **Adoption survey.** 50 files under `R/` handle `area_code`; 13 resolve polities. Blanket adoption would be wrong, since most legitimately pass through an already-resolved table — the work is separating genuine *entry points* (where area identity enters from an external dataset) and *region aggregations* (which should be defined over polities, not ad-hoc iso3c) from harmless pass-throughs.

---

# Update: what the branch grew into

The description above covers the first two commits. The branch is now 14, and the later work came from cross-checking this package against `whep-polities` rather than from inspection. Summary for review, with the decisions I did **not** take.

## Correctness fixes

**Pre-independence years resolved to the modern polity** (`325e751e`). The crosswalk mapped each area to exactly one polity prefix, and upstream gives colonial and modern polities *different* prefixes. So the colonial polity was unreachable and resolution fell through:

| | before | after |
|---|---|---|
| 1965 Angola | `AGO-1975-2025` | `ANG-1905-1975` |
| 1970 Sudan | `SDN-2011-2025` | `SUD-1956-2011` |
| 1962 Zimbabwe | `ZWE-1964-1980` | `SRH-1953-1964` |

**118 area-years change attribution, and all 118 previously had no span-covering polity at all** — the crosswalk lacked coverage and `add_polity_code()`'s fallback masked it by assigning the nearest modern polity. Sudan is 50 of those years, attributed to a polity that by definition excludes the South Sudanese territory those reports covered.

Fixed by listing both prefixes per area. No year bounds needed: the join is already many-to-many and resolution already year-aware. Verified first that all seven chains are contiguous and overlap-free, including the two that interleave (`ZWE/ZWE/SRH/ZWE/ZWE`, `VNM/F237/VNM`).

**`regions_full$polity_code` never held polity codes** (`7a0d7155`). All 271 values were bare ISO3-shaped prefixes, so the join the name invites returned nothing, silently. Renamed to `polity_prefix`; the real code was already in `reporting_polity_code`. The same rebuild also refreshed the two datasets, which were stale: 12 area codes changed reporting polity and 8 old targets were retired or superseded upstream.

**Ambiguous ISO3 broke on row order** (`04bf01e6`). `ETH` names both FAOSTAT 62 ("Ethiopia PDR", dissolved 1993) and 238, and `unique(by = "iso3c")` kept whichever came first — 62. Now broken on the polities database instead. Latent: the helper has no callers, so this fixes a trap, not live output.

**Dissolved-country filter was magic integers** (`327397db`). Replaced with a table that separates the two bounds it conflated — polity existence (in the database) from FAOSTAT reporting convention (not in it, and narrower). That distinction is why the filter *cannot* be replaced by a join: deriving it from spans would readmit Belgium 1831-1999 on top of the Belgium-Luxembourg aggregate that already carries it. The polity spans now bound every window, which found two missing lower bounds.

## Cross-repo verification

`test_upstream_faostat_agreement.R` compares this package's area→polity mapping against the one `whep-polities` publishes. Over 281 mappings:

- **0 true conflicts** — nowhere do the two name different polities for the same area and overlapping years
- 31 areas folded into `ROW-1850-2023` here while a real polity exists upstream
- 7 era gaps — **now closed** by `325e751e`

Both gaps are baselined **bidirectionally**: a new gap fails, and a *closed* gap fails until removed, so the baseline shrinks as work lands rather than licensing regressions.

## Decisions I left to you

1. **31 areas folded into Rest of World** while real polities exist upstream — not only small islands: **Syria** (113k layer-B rows), **Swaziland** (181k), **North Macedonia** (115k), **New Caledonia** (161k). Moving an area out of `ROW` changes published aggregates, so it is a deliberate call. Baselined, untouched.
2. **Prefix unification** (#387, left open). This branch compensates for the colonial/modern prefix split; unifying the prefixes upstream would fix the cause. `ANG` is not even Angola's ISO3. Changes `polity_code` identity, so it invalidates the manifest hash.
3. **14 of 17 region-classification columns have no consumer** (#386), including one named `region_test`. Breaking to remove from a public dataset.
4. **Three datasets carry unresolvable area identifiers** (#389, 779 rows, all latent). Right fix is upstream aliases, not a lookup here.

## What I deliberately did not change

`region_krausmann` and the other classification taxonomies, and `.energy_country_grouping()`'s GLEAM buckets. These are **source-owned** geographies; rebuilding them from polity continents would change which coefficient applies to whom. I verified instead that they never *contradict* the polity structure — all 17 columns agree within every non-aggregate polity, the only conflict being `ROW-1850-2023`, which spans continents by construction.

## CI caveat, stated plainly

For several iterations I ran the suite locally with `WHEP_POLITIES_MANIFEST` and `WHEP_POLITIES_FAOSTAT_MAP` pointed at the sibling worktree — conditions CI does not have — and reported it green while this PR's CI was red. `524ad7af` fixes both causes (a polygon-gap test that was *stricter* on CI because it fell back to a stale hardcoded copy of upstream data, and unformatted R files). Verified since under CI conditions: **808 passing, 0 failures, 6 skips**, the skips being the two upstream-contract tests correctly declining to compare against an absent second opinion.

Note `main` itself is currently red on `ubuntu-latest (release)` — `test_n_prov_destiny.R:1015`, a missing `pie_full_destinies_fm` parquet. Unrelated to polities and inherited here, not caused by this branch.

---

# Update 2: the last twelve commits

The addendum above covers the branch through the colonial-era fix. Since then the
work split into three strands: one new capability, several defects found by
cross-checking, and three CI failures that were mine.

## New capability: labels resolve to polities

`add_polity_code()` handles numeric FAOSTAT area codes. **Nothing handled a country
label**, so datasets carrying one had no supported path to a polity —
`mueller_synthetic_n$iso3c` holds FAO-style legacy codes (`"BZE"` for Belize,
`"ROM"` for Romania, `"ZAR"` for Zaire, renamed COD in 1997) and
`lassaletta_grassland_share$Country` holds name variants (`"Cape Verde"`,
`"Swaziland"`, unaccented `"Reunion"`).

`resolve_polity_label()` fills that gap, and `polity_label_aliases` **consumes** the
map whep-polities publishes rather than building one here — a second lookup would be
a second authority for the same question, which is what produced the 118-area-year
misattribution. Resolution is source- **and** year-aware: `"Cape Verde"` is
`CPV-1886-1975` in 1970 and `CPV-1975-2025` in 1990. `NA` is a real answer —
`"FSU"` and `"Belgium-Luxemburg"` are aggregates the source reports past the end of
the territory they name, so their tail years resolve to nothing rather than to a
polity that had ceased to exist.

The whep#389 baselines shrank from 10 and 9 unresolved labels to **0 and 2**, and
became a sharper claim: previously "does not resolve against polity *names*" (string
matching), now "does not resolve through the published map" (what a consumer
actually asks).

## Defects found by comparing two implementations

**My own resolver was wrong for five commits.** Its comment claimed it mirrored
`matchlib.Matcher.match_alias`. The tie-break ordering did; the **normalisation did
not**, and that is the half deciding which rules are even considered. A cross-check
over **6,627 probes found 25 disagreements**: upstream also folds accents, **drops
parenthesised qualifiers**, and strips a leading `"the"`. Two of the 25 resolved to
a *different polity* — dropping the parenthetical merges `"Sudan (former)"` into the
`sudan` rule set, so 2011 goes to pre-secession Sudan upstream and post-secession
Sudan in my version. Now 0 disagreements.

**Missing `iso3_code` was the literal string `"NA"`.** The GeoPackage round-trip
writes missing text that way, so `is.na(iso3_code)` found **3** missing codes when
**82** were missing, and every `!is.na(iso3)` guard treated 79 rows as carrying a
valid ISO3 — which the ISO3-keyed bridges in `read_raw_inputs.R` do. Fixed at the
read. This also explains a workaround I made earlier in this branch and never traced
to its cause.

**The contract test had a blind spot.** When `GCO-1884-2025` was retired upstream,
every existing assertion still passed — the row count was unchanged (a retirement
moves a code between lists), the code set unchanged for the same reason, and nothing
routed to it. It now compares the dead-code *set* and every identity field row by
row. The hash is deliberately not replicated: it digests Python's `json.dumps`, and
reproducing that from R gave a different digest from identical data.

**`add_polity_code()` and `matchlib.pick_by_year` disagree on 39 of 16,960
area-years** — all Belgium-Luxembourg 1961-1999, the one family with two genuinely
overlapping live polities. Neither is wrong: matchlib prefers `national` because it
maps a *label*, this prefers the aggregate because it maps a FAOSTAT reporting
*area*, and area 15 **is** an aggregate area. I nearly "aligned" them, which would
have moved 39 years of data onto a different polity to make a cross-check green.
Documented as deliberate instead, pointing at whep-polities#40 where the overlap is
tracked.

## Also

- **4 of 31 `ROW`-folded areas moved out — at the POLITY level only.** All 31 carry
  `fabio_code == 999`. Only **4** are flagged `cbs` — New Caledonia, North Macedonia,
  Eswatini, **Syria** (~113k layer-B rows) — so only those had their own data landing in
  Rest of World, and those four get their own `polity_code`. The other 27 carry no CBS
  data and are correctly folded.
  **Their `polity_area_code` still folds to 999, deliberately**, so their VALUES are still
  aggregated into Rest of World. Making the numeric key follow the polity inflates global
  `feed` 13.7× (#419). So `reporting_polity_code` and `polity_area_code` disagree for these
  areas by design, and that is the single most important thing to know before joining on
  one and assuming the other follows.

## Three CI failures, all mine

Worth stating plainly, since they say something about the method:

1. **I reported the suite green while this PR's CI was red**, because I ran it with
   `WHEP_POLITIES_MANIFEST` pointed at the sibling worktree — conditions CI does not
   have. One test was actually *stricter* on CI, since it fell back to a stale
   hardcoded copy of upstream's polygon-gap list.
2. **Unformatted files** and **UPPER_CASE constants** against a snake_case lint rule.
   The repo already had `inst/scripts/check_lint.R` with CI's exact linter set; I
   wrote my own default `lint()` call, which reports ~40 lints CI ignores and would
   have buried the 8 that mattered.
3. **pkgdown** errors on an exported topic missing from `_pkgdown.yml`.

`inst/scripts/check_like_ci.R` now runs formatting, lint (delegating to the existing
helper), the suite **with the upstream contracts made unreadable**, and the pkgdown
index — and states what it does *not* cover, so a green local result is not mistaken
for a green PR. I then pushed through a failing lint by chaining the check and the
commit in one command, which defeated the guard by construction; that is fixed too.

## Verification

`check_like_ci.R`: formatting clean, **0 lints**, **4,517 assertions passing, 0
failures**, 269 topics indexed. CI green across all five workflows.

---

## Smoke run against the real pins

The example fixtures were never enough. With the pin cache available I built
`get_primary_production()` on real data, and it found two things nothing else could.

**A regression this branch introduced.** `.read_fodder_euadb()` selected `polity_code`
from `whep::regions_full` — the column renamed to `polity_prefix` here, precisely because
it held prefixes rather than codes. The build aborted with *"Column `polity_code` doesn't
exist"*. It was invisible because `example = TRUE` never reaches that function, the real
path is one of the 12 `[external]` skips, and lintr's `object_usage` checks are off in the
CI config — so a stale column name in a `dplyr::select()` is not a lint either.

**The fold was still in force at the numeric level — and it still is, deliberately.** I had
reported the eleven data-reporting areas as pulled out of the FABIO rest-of-world fold. True
of `polity_code`, false of `polity_area_code` — which takes `fabio_code` (999 for every
folded area) and is what the builds actually key on: `get_primary_production()` emits it *as*
`area_code`, and `build_trade.R` does `area_code := polity_area_code`. Raw production holds
2,458 rows for the Faroe Islands and 9,606 for Palestine; the built output has **zero** for
both.

I then "fixed" that by promoting them numerically, which is the change measured below and
**withdrawn**. The figures are kept because they are what a reviewer would otherwise try to
reproduce and fail to:

| | main | promoted (withdrawn) | shipped |
|---|---|---|---|
| rows | 6,163,136 | 6,295,555 | ≈6,163,000 |
| distinct area codes | 194 | 209 | **194** |
| global `feed` | baseline | **13.7×** | baseline |
| rows with no reporting polity | 0 | 0 | 0 |

The extra 15 areas were real coverage, and giving them their own key really does stop their
data being attributed to `ROW-1850-2023`. It also inflated global `feed` 13.7× with the whole
increase on one area (212 Syria, twelve times the world total), so the attribution problem is
left open in #419 rather than traded for a larger error. The cost of leaving it open is
measured too: 1,389,871 t of feed demand dropped for want of a Bouwman region on RoW.

Worth stating why my own earlier verification missed it: I reported **"rows with NO
`reporting_polity_code`: 0 (0.000%)"** as proof the integration held. That was true and
concealed this completely, because 999 resolves to `ROW-1850-2023` perfectly well. *A
coverage metric cannot detect data arriving at the wrong polity, only at none.*

## Consequences measured rather than assumed

- **Pre-1962 back-cast**: areas with no LUH2 land match went 5 → 9. The four added are
  exactly the French overseas departments — LUH2 folds *départements* into `FRA`, so no
  separate land series exists. Their pre-1962 production is no longer back-cast, which is
  the honest outcome but is a real change. Pinned in `test_luh2_land_coverage.R`.
- **LUH2 land dropped for unmatched codes**: measured twice. My first figure (0.0008%,
  six territories with no polity) conflated two causes; against the actual predicate the
  code uses, the only unmatched code is `-99` at **0.31%** — LUH2's own unattributed-land
  sentinel, upstream of this package. Both asserted separately.
- **`region_UN` vs the polity's own continent**: 199 of 253 agree. 53 disagreements are the
  ROW fold; the 54th is a defect *in the vendored table* — the USSR filed under Asia while
  its successor Russia is under Europe. `reporting_polity_continent` now carries the
  polity's answer.

## A guard for the failure mode that kept recurring

Two column renames each left a test reading the old name, and each test **kept passing** —
reading an absent column returns NULL, and `all(is.na(NULL))` is TRUE. Both times the only
signal was the harness's warning count rising with failures at zero. `check_like_ci.R` now
fails on that class and names the test.

## What is verified, and what is not

Verified by measurement, each with a number in the commit that made the change:

| path | evidence |
|---|---|
| `get_primary_production()` real build | 6,295,555 rows, 209 areas, **0** with no reporting polity |
| fodder bridge | **0** of 23,183 rows unresolved (was 13.3%) |
| gdp-population proxy join | **33 of 33** match; **0 of 33** without the fix |
| area-name conventions | crosswalk `area_name` == `FAOSTAT_name` on **260 of 260** |
| pinned inputs | production/production-old/international-yields/fishstat all **0** unresolved |
| IPCC region chain | 9 producible regions all present; `Global` covers **12 of 12** categories |
| area overrides | derived from upstream; **12** areas, agreeing except the documented area 206 |

**Not verified end-to-end: the full `get_wide_cbs()` real-data build.** Three attempts were made
and none completed — the build runs well past half an hour and the process did not survive to the
end in this environment. What that run would have exercised is `.fill_with_proxies()`, and that is
instead verified directly: the proxy join is measured both ways above, and the eleven previously
unmatched names now resolve to a population for every year tested.

Stated rather than left implied, because "a smoke run is pending" is not a status — it is either
evidence or it is not, and here it is not.

---

## Silent drops, and what a dropped row now tells you

Three sites discard rows whose area resolves to no polity. Dropping is correct at all three; being
silent about it was not — `read_raw_inputs.R` alone removed **1,109,466 rows of real FAOSTAT
production (26%)** and said nothing, so a genuinely unknown area code vanished as quietly as an
expected one.

A dropped row now reports which of three things happened, at the right severity:

| reason | example | severity |
|---|---|---|
| deliberately unmapped | 351 China, reported alongside its own components | informational |
| FAOSTAT regional group | codes ≥ 5000 — World, Africa, Eastern Africa | informational |
| **unknown code** | neither | **warning** |

Verified on a real slice: 244 area codes in, 205 out, `15 and 351` reported as deliberate, 34 reported
as regional groups, **no warning** — because the third bucket is empty on today's data. That emptiness
is the point, and it is asserted across all four area-keyed inputs (production 244 codes,
production-old 221, international-yields 179, fishstat-trade 189; none unaccounted).

The classification took four attempts, all recorded in `test_dropped_area_reporting.R`:

1. one message for everything
2. split via `mapping_status` — **wrong**: every call site passes `include_unmapped = FALSE`, which
   strips unmapped rows from the lookup, so the message called China *"an area code this project does
   not know"*
3. split via the crosswalk — right mechanism, wrong effect: all 34 absent codes are FAOSTAT groups, so
   34 expected codes became input errors on every read
4. three buckets, warning only for the third

Each wrong version looked correct in isolation; each was caught by measuring real data.

## Stop deciding upstream's questions

`whep-polities` publishes `faostat_unmapped_areas` (the 5000 threshold and the deliberate codes)
because its matcher already acts on both and this package was re-deriving them — measuring the
threshold against production to arrive at a number upstream already knew, and *inferring* "deliberate"
from crosswalk membership. Inference cannot tell a decision from an absence, which is what produced
the China defect above. Both values are now embedded from the manifest by `data-raw/constants.R`, with
a test that compares the copy against the manifest and says *"rerun data-raw/constants.R"* on drift.

Same move as replacing the eleven hand-written area overrides with a read of the published map.

## Documentation that described a different package

Four defects, each a change made correctly in code and incompletely in its description — none of which
any test exercising the code would fail on:

- `polygon_status` documented as `"missing"`/`"excluded"`, values the vocabulary no longer has
- `start_year`/`end_year` described as **inclusive**; `end_year` is exclusive, so the docs told readers
  to write a filter that double-counts every boundary year
- `polity_label_aliases`' exhaustive column list missing `observed_rows` — the column this branch added
- `area_iso3c` documented without noting that the neighbouring `iso3_code` means something else, and
  differs on 56 of 273 rows

`test_polities_doc_matches_data.R` now pins the facts the prose asserts.

---

## Every dataset that names a territory now reaches a polity

The earlier sections are about `area_code`. A full census of the package's data — the lazy-loaded
datasets **and** the internal ones — found 21 area-keyed tables with no polity column. That
distinction is load-bearing: an `ls()` over the namespace alone returns **one** such table, because
lazy-loaded data does not appear there, which is how `gleam_geographic_hierarchy` escaped two
earlier sweeps.

They split in two, and the split is the point.

**Keyed by a territory — should carry polities, and now do:**

| dataset | key | coverage |
|---|---|---|
| `crops_manure_n` (31,648 rows) | ISO3 | **184/184 codes, 100% of 112,887,685 Mg** |
| `mueller_synthetic_n` | ISO3 | 156/156 |
| `urban_n_reference` | area_code | complete |
| `expand_trade_sources()` (1,133 rows) | reporter + year | **100%, 31 polities** |
| `lassaletta_grassland_share` (6,909 rows) | country name + year | **14.5% → 98.1%** |

187 aliases were added upstream rather than name-matching here, because whep-polities owns
label-to-polity identity.

Lassaletta's remaining 128 country-years are one class: **source anachronisms**, where resolving
would be wrong. `South Sudan` is listed from 1961 though it existed from 2011; `FSU` and
`Yugoslav SFR` run past their dissolutions; `Czechoslovakia` reaches 2009 though it dissolved in
1993. Each resolves correctly for the years its entity did exist.

**Keyed by a region taxonomy — should NOT carry polities:** the GLEAM, IPCC, Bouwman, Krausmann and
HANPP coefficient tables. A GLEAM region is no more derivable from a polity than an IPCC region.
"Add a polity column" is the wrong fix; what they need is that the vocabulary a table is *keyed by*
matches the one it is *joined against*.

## Three defects that vocabulary comparison found

**Comoros' grid cells were labelled Mayotte.** `cow_to_lpjml.csv` had `45,"MYT","Mayotte",153`.
FAOSTAT area 45 is **Comoros**; Mayotte is 270, absent from the file, and no row carried `COM`. The
LPJmL country raster therefore labelled every Comoros cell with Mayotte's index. Which field was
wrong came from the file's own structure — rows are ordered by `iso3c` and `lpjml_code` tracks that
order (MRT 147, MUS 150, MWI 151, MYS 152, **MYT 153**), so the index was right and the `area_code`
wrong. Fixed 45 → 270; Comoros deliberately left unmapped rather than given an inferred index
(whep#404).

**A ninefold coefficient range collapsing to one number.** `residue_feed_fraction.csv` has a column
named `region_hanpp` that holds **UN M49 sub-regions**: 0 of its 17 values are in `region_HANPP`, 14
are in `region_UN_sub`. It is joined against `region_HANPP`, so every row takes the `Global` default
of 0.20 while the file spans 0.05–0.45. Confirmed by running the function, not reading it: six
regions, six identical fractions, while the *neighbouring* join's recovery rates varied correctly.
Baselined rather than fixed — which taxonomy each cited coefficient belongs to is a scientific
judgement (whep#405).

**One override applied to one of two copies.** `polities_cats` and `regions_full` have identical 40
columns and a strict subset relation, from separate vendored CSVs. 18 columns disagreed. One was a
real defect and mine: `ADB_Region` was `AT`/`GB` in one table and `NA` in the other because the
override was written inline on `regions_full` alone. Now a function applied to both.

## Guards, not just fixes

- the `regions_full` column a coefficient file best matches must be the one it is joined against —
  the invariant #405 violates, now asserted for all three region-keyed coefficient tables
- GLEAM continents must equal the polities' continents, with `Americas` → {North, South America}
  declared, so the 40th difference fails instead of hiding among 39 granularity differences
- every `cow_to_lpjml` row's `iso3c` must agree with the crosswalk for its `area_code`
- no alias for a locally-owned source may be inert (152 checked, 0 inert) — upstream cannot check
  these, because it never sees the data
- the gate script now runs the contract tests **with** contracts present, and refuses when its three
  upstream inputs do not resolve to one directory

---

## Decisions: what is settled and what is not

Eight questions were raised over this branch where a change would move data or published numbers. Rather than
leave them as prose, each was measured. **Every one turned out to have evidence available in the repository that
I had not consulted before raising it** — twice the evidence contradicted my own recommendation.

**Closed:**

| | how |
|---|---|
| #404 Comoros LPJmL index | `include/managepar.h` defines `Comoros 46`, confirming an inference I had declined to act on. Row added. Both LPJmL 6.0.5 and 6.1.1 agree on all 257 indices, and all 190 of ours exist in them |
| whep-polities#53 Ethiopia AOI/ETH | the registry's `basis` column already explained it, per source and by measurement: ">1.5x pre-1936 ETH series magnitude" for the AOI route, "no step consistent with the AOI aggregate" for the yearbook sources. I filed it claiming the basis was silent |

**Narrowed to a bounded choice:**

| | what measuring established |
|---|---|
| #403 ROW ≠ FABIO | a FABIO comparison **does** exist (`inst/scripts/compare_fabio_footprints.R`) and is **unaffected** — it targets CHN/USA/EU28 by iso3c, and the French overseas departments are GUF/GLP/MTQ/REU, not FRA. Now asserted, so a future unfold that *would* affect it fails |
| #405 residue feed fractions | the fix is **two lines** — `.sci_crop_regions()` already reads the table holding `region_UN_sub`. Effect: 0 → 230 of 260 areas get a region-specific value, 1 → 8 distinct fractions |
| #407 LUH2 dependencies | **no double-counting risk**: LUH2 partitions land (13,589 Mha for 2000 ≈ global ice-free land). Cost of attribution is 0.25% for the UK, 0.49% for the Netherlands, and Saint Barthélemy has no LUH2 land at all |
| #408 LUH2 back-cast | the spatial route **works**: a point-in-polygon test recovers all 15 Soviet republics, 23 of 24 successors overall, all present in LUH2. What remains is the within-federation weighting, which is a modelling choice |
| whep-polities#54 alias spans | four of six early-start ranges have reasoning recorded; Gold Coast resolves outright (its data spans 1937–1951, inside the polity). Only the convention question is open |

**Where measuring showed me wrong:**

- **#406/#395** — I suggested deriving `polities_cats` from `regions_full` "by its row filter". No filter exists: the best predicate is `cbs == TRUE` at 93.1%, with 18 exceptions that look like a curated set drifting as the CBS series changed. That is now the *worst* of the three options, and it was mine.

## Gates

Nine, and they prove they can fail. `inst/scripts/selftest_gates.R` injects a defect per gate and requires the
gate to report failure **and name what it found** — because exit status alone cannot distinguish a detection from
a crash, which cost two false passes in the sibling repository's harness. The self-test itself is
mutation-tested, since adding an unfailable gate about failability would be its own joke.

## The database's own names were the ones it could not resolve

`resolve_polity_label()` consulted the alias map alone, so a caller passing the database's OWN name for a polity
got `NA` — `resolve_polity_label("Netherlands")` found nothing while `polities` carries a polity named exactly
that. Upstream resolves "by alias, **then** by ISO/name family + year containment", so two implementations of one
question disagreed on some of the most common country names there are.

The cross-check gate could not have found this: it draws its probes **from** the alias map, so every label it tests
resolves by construction. It surfaced from the consumer side — `get_primary_residues()` is the one builder that
resolves areas by name, and it left **44,985 of 475,688 residue rows (9.5%)** with no area code across 14 labels,
every one a short form of an area the crosswalk holds under a FAOSTAT long form.

Two guards decide when a name may answer, both measured rather than assumed:

| guard | why | evidence |
|---|---|---|
| exactly one polity of that name live in the year | 52 pairs share a normalised name **and** overlap in years | `Peru` 1865, `Hungary` 1939 → `NA`, not a row-order guess |
| a rule covering that year outranks the name | curated routing must win | `djibouti` 1961 → `NA`; every alias routes to the FRS family |

Two cruder rules were measured and discarded first. Falling through on any alias miss gave **69** answers
contradicting every rule written for their label; refusing to fall through on a *source* mismatch killed those and
also killed the fix, because `Iran`, `Netherlands` and the other seven carry source-scoped `faostat` aliases. What
separates them is not source but **agreement**.

Sweeping 3,760 `(label, year)` pairs against the previous implementation: **1,509 filled in, 0 lost, 0 changed, 0
contradicting a year-covering rule.**

`add_area_code()` now routes names the exact join cannot match through the polities database. It **fills `NA`
only**, so the crosswalk stays authoritative wherever it has an answer, and it is year-aware because
`"Czech Republic"` is Czechia from 1993 and Czechoslovakia before it.

| residue rows with no area code | |
|---|---|
| before | 44,985 (9.5%) |
| after this fallback | 10,022 (2.1%) |
| after rescoping four aliases upstream | **200 (0.04%)** |

The 200 that remain are all Tanzania 1961-1964, blocked by the `TAN-1922-1964` code/column disagreement upstream
baselines — so that baseline now has a measured downstream cost.

**One guard exists because the route got a case wrong and a gate caught it.** `China` resolved to `CHN-1950-2025`
"China (PRC)", because normalisation drops parenthesised qualifiers. But that label is FAOSTAT area **351, the
aggregate** of mainland (41), Hong Kong (96), Macao (128) and Taiwan (214), each reporting separately — so
answering with the mainland polity double-counts against its own four components. The refusal is derived from the
deliberately-unmapped area list this package embeds from upstream, so it cannot drift from the contract.

## Group by the reporting bucket, not by the bucket and its label

Two aggregations reduced areas to `polity_area_code` — the reporting bucket the polities database defines — and
grouped by the area **name** alongside it. Several areas fold into one bucket, and when they carry different names
the label split the bucket, so the folded areas were never summed.

Measured at 1990-2023: **1,525 duplicate `(area_code, year, item_cbs, item_cbs_code, element, source)` keys**
reaching the cast, every one in bucket 206 — which is FABIO's Sudan region, folding FAOSTAT 206 "Sudan (former)",
276 "Sudan" and 277 "South Sudan". From 2014 the latter two both report under different polity names, so the split
fires for 2014-2023. Changing `.aggregate_to_polities()` took 1,525 → 12, and `.get_traded_residues()` had the
same shape, taking it to 0 — but that change is **withdrawn** (see the callout below), so the duplicates remain.
At full range there are **31,642** of them; 1,525 was only what the 1990-2023 window showed.

**What it cost is worse than a duplicate row.** `dcast()` with no `fun.aggregate` answers a duplicate key with
`length()`, so values become **row counts**. Whether anyone finds out is luck: counts are integer, and
`FBS_Old_scaled` is double only when `scale_new_old` applies, which depends on the build window. At 1990-2023 the
build died in `fcoalesce`; over the full range it completed with counts standing in for quantities. The run that
still had 12 duplicates reported `CBS OK rows: 2604262` — the same row count as the clean run, because `dcast`
collapses a duplicate either way. **Only the values differed, so no row-count check could have seen it.**

So `.select_best_source()` now **warns** when duplicate keys reach the cast, naming the area codes, sources and
years, and saying explicitly that the values become row counts. A warning rather than an abort because these
duplicates exist on `main` too — aborting refuses to build a pipeline that has always had them, which is what
killed the first bisect of this very problem.

> [!IMPORTANT]
> **"Summing into the bucket is the correct operation" was wrong, and this section's own italics said why.**
>
> *"Only the values differed, so no row-count check could have seen it"* — written here about `dcast`, and then
> true of my own fix. Summing the areas that fold into one bucket double-counts: rows for a member and rows that
> already aggregate that member both land there. Measured against `main`: `food` **266×** -- an attribution
> I later disproved. That change bundled a relabelling of the `area` join key which alone dropped 702,166
> rows; isolated, summing is harmless and in fact removes the cast's duplicates, making it a candidate fix
> in **#425** rather than an error. What follows describes the withdrawal as it stood at the time.
> `feed` 12×, `import` 8.3×, while row counts moved barely at all. 5151 tests passed throughout, because nothing
> compared a magnitude with anything.
>
> Grouping by `(polity_area_code, polity_name)` — which looked like the defect — is correct: the name
> distinguishes territory-periods. The duplicate count is also larger than reported above: **31,642** `(key,
> source)` combinations at full range, not 1,525, which was the 1990-2023 window.
>
> What remains is the real decision, and it is not mine: choose `sum` or `first` as `fun.aggregate` at the cast.
> Either changes published numbers, and `sum` is not obviously right for the same reason it was wrong upstream.
> **#418.** `inst/scripts/compare_cbs_totals.R` now pins column totals against a baseline so this class of error
> cannot pass silently again.

**Pre-existing, and not caused by routing more areas through the polities database**: `main` fails identically at
the same window, with the same 1,525 keys, bucket, sources and years. An earlier claim of mine that this was a
regression compared main's *full-range* success against this branch's *narrow-window* failure, which is not a
comparison.

Three tests, each shown to fail against the unfixed code. The first version of the first was **vacuous**: it picked
the two lowest area codes in a folding bucket, 206 and 276, which both resolve to the polity "Sudan", so nothing
split — the split needs 277. It now searches for the condition itself, so it survives upstream renaming or
unfolding Sudan.

## What "all area-related code" was taken to mean, and how it was enumerated

The claim in the title is only worth as much as the enumeration behind it, so here is the sweep,
by axis, with what each turned up.

| axis | universe | result |
|---|---|---|
| package datasets | **95** in the namespace's lazydata env, **23** carrying a territory column | all 23 documented and referenced by a test; censused in `test_area_dataset_census.R` so a new one cannot arrive unlinked |
| `inst/extdata` | every CSV with an area/country/iso/region column | `regions.csv` verified (206 of 209 iso3c are polity prefixes; the 3 dissolved federations resolve via `F51`/`F228`/`F248`), `conv_bouwman` already asserted against the crosswalk, `crops_manure_N$Continent` documented as not-a-continent |
| exported API | **7** functions with an area-ish argument | six take frames or paths; `get_polity_geometries()` silently dropped unknown codes and now names them |
| coefficient vocabularies | 22 coefficient tables + GLEAM + IPCC | region taxonomies are legitimately their own (a GLEAM region is not derivable from a polity); what is asserted is that joins do not cross vocabularies |
| `inst/scripts` | analysis scripts | `c("CHN", "USA", "EU28")` is FABIO's own vocabulary for a FABIO comparison — not an area definition |

**The enumeration itself was the hardest part to get right.** A sweep over *exported* objects returns a
confident zero, because package data is not an export — data objects live in the namespace's lazydata
environment and are reachable through `::` only via `LazyData`. That is precisely how
`gleam_geographic_hierarchy` — a 204-country table defining the country universe for the entire energy
extension — went unnoticed until it was swept by hand. The census test is built on `lazydata` for that
reason, and is guarded twice against passing on nothing.

## Measured outcomes

| | before | after |
|---|---|---|
| residue rows with no area code | 44,985 (9.5%) | **200 (0.04%)** |
| label resolutions across 3,760 probes | — | **+1,509**, 0 lost, 0 changed |
| duplicate keys corrupting CBS values | 1,525 | **0**, verified at full range |
| CBS rows with no polity (full range) | — | **0 of ≈2,764,000** |
| `regions_full` vs `polities_cats` disagreements | 17 columns | **4**, all one deliberate fold |

## Raised rather than decided

Three questions this work surfaced are modelling decisions, not patches, and are filed with
measurements instead of being resolved unilaterally: **#414** (a summed bucket attributed to a
single-territory polity — the aggregate-polity route this repo already uses for `BLX-1850-1999` is
recommended), **#415** (five live territories with no GLEAM row and therefore no energy grouping),
and upstream **whep-polities#54**, which now carries the measured cost of the `TAN-1922-1964`
decision: exactly the last 200 unmatched residue rows.

## Validation, and what each check could not see

| check | result |
|---|---|
| R CMD check, five platforms | **green** on the branch merged with main |
| test-coverage | **green** |
| lint, pkgdown, format-suggest | green |
| local gate (10 gates, incl. self-test) | green |
| full suite | 5063 pass, 0 fail |
| **each of the 26 files touching the area tables, run INDIVIDUALLY** | **1187 pass, 0 fail** |
| full-range `get_wide_cbs()` | ≈2.764M rows across **195** areas, **0 NA** in all four polity columns. The count varies by 4,107 between identical runs — the build is not reproducible, #420 |
| published alias map conformance | **851 of 851** ranged aliases resolve to their own target |

The last-but-one row is not redundant with the suite, and neither is the isolation run. Three
failures in this work were invisible to a green suite:

- **`.carry_forward_shares` only worked inside the suite.** `tidyr::nesting()` evaluates with
  tidy eval, so a missing column is looked up in the CALLING ENVIRONMENT — another test file
  leaves an `area` object in scope. Run alone, it errored. Found by running the 26 files
  one at a time, which is why that row exists.
- **A helper was defined below its own call sites.** Local runs reported 5063 passing tests;
  R CMD check failed on all five platforms with `could not find function`. Now caught
  statically before a push by a new gate, and swept: zero other instances in 166 files.
- **The matrix never reported for hours.** `R-CMD-check.yaml` sets
  `concurrency: cancel-in-progress`, so every push killed the running check — four
  consecutive runs died before speaking, and "pending" read as slow rather than doomed.

Two more of the same shape, both fixed: the suite reported a warning COUNT with no way to
learn which warning arrived (now enumerated), and a census built on *exported* objects
returned a confident zero because package data is not an export (now built on the lazydata
environment, which is how `gleam_geographic_hierarchy` was found at all).

## Silent failures found by asking what each check could not see

Three more, after the ones above, each found by asking a question the existing checks were not
shaped to answer.

**A diagnostic that could not report the case it existed for.** `.warn_unmapped_codes()`
classifies unmappable rows three ways — deliberate non-mapping, FAOSTAT regional group, unknown
code — and every one is a property *of a code*:

```r
codes <- sort(unique(original[is.na(mapped)]))
if (length(codes) == 0L) return(invisible(NULL))
```

`sort()` drops `NA`. So rows whose code is itself missing produced an empty set and the function
returned in silence — it could report a code that failed to map, never the absence of one. That
is exactly what a name-keyed source produces when a label does not resolve, and it is how **160
crop-residue rows were dropped without a word**. Added as a fourth outcome, so it reports at all
five call sites. My first attempt was *inert*: I added the call at the residues drop point, saw
zero warnings, and had to find out why rather than conclude the drop was empty.

Bounded honestly: **zero** rows lack a code in the FAOSTAT path over 1990-1993, because FAOSTAT
always supplies codes. The fourth outcome is a guard for name-keyed sources specifically.

**Documenting a polity column is not emitting one.** The existing gate verifies all eight
builders document the polity columns they attach. Asking whether they *emit* them populated found
two that do not — `build_supply_use` (1 row of 10) and `get_feed_intake` (2 of 11), both because
those rows carry `area_code = NA`. `build_supply_use`'s row has `value = 3.33e-14`, floating-point
residue rather than an observation. Pinned at current counts rather than asserted to zero, so the
three clean builders cannot start emitting `NA` quietly and fixing #417 forces the pin to be
updated. Filtering the rows away in the polity step would hide the join behaviour instead of
fixing it.

**Two published columns measuring one thing under transposed names.** `observed_rows` in the
alias map, `rows_observed` in the FAOSTAT map. Same registry column, same coercion, different
word order, neither documented. Documented rather than renamed — the names are part of a contract
this package already reads.

## My own figures drifted too

The clean-build section claimed 170,972 and 265,124 rows; measured now, **170,976 and 265,640**.
The bucket-aggregation fix collapses keys that used to be duplicated and `add_area_code()`'s
fallback admits rows that used to be dropped, so the counts *should* have moved. Re-measured
rather than adjusted by arithmetic, and the claim they support was re-checked at the same time:
**zero `NA` across all four polity columns in both frames**, verified column by column.

That is the third quoted number in this work to drift from the code that produced it, after the
alias counts and the LUH2 nine-versus-eight — which is why `test_alias_map_conformance.R` asserts
against the live map rather than a remembered list.


lbm364dl:
Paired PR is open: **lbm364dl/whep-polities#39** — it publishes `data/final/polities_manifest.json`, a contract carrying row counts, an `identity_sha256` over the fields a consumer resolves against, the live polity codes, and `dead_polity_codes` (the rows that must never receive data).

Two things about its design matter for this side:

- The **dead set is named explicitly** rather than left implicit in a status column, so this package cannot repeat the dead-polity mistake by omission. Its `dead_status` mirrors `matchlib.Matcher.DEAD_STATUS`, which is the same rule this PR's change 1 implements — so both sides of the boundary now agree on what "must never receive data" means.
- The **hash covers identity, not geometry** (`polity_code`, `polity_name`, `start_year`, `end_year`, `polity_type`, `iso3_code`, `cow_code`, `wiki_status`). A re-measured polygon will not invalidate this package's copy; a changed date or status will. Without that split the check would fire on every upstream polygon fix and get ignored.

Next commit here consumes it: a test asserting `polities.rda` matches the manifest's count and hash, and that no crosswalk row targets a `dead_polity_codes` entry. That converts "603 vs 740 discovered by inspection" into a failing test.

lbm364dl:
## Merge order

Verified that every upstream path this package reads by default matches exactly what **lbm364dl/whep-polities#39** publishes — five for five:

| file | read by |
|---|---|
| `polities_database.gpkg` | `data-raw/table_mappings.R` |
| `polities_database.csv` | the identity-field contract test |
| `polities_manifest.json` | the contract and coverage tests |
| `faostat_area_polity_map.csv` | the matcher-agreement test |
| `label_alias_map.csv` | `data-raw/table_mappings.R` |

**whep-polities#39 should merge first.** Its `main` already carries `polities_database.{csv,gpkg}`; the other three arrive with that PR. Of those, only `label_alias_map.csv` is needed to *build* rather than to test, so it is the one artifact a regeneration can be missing in between.

This is not a blocker for merging in either order:

- **Tests** skip cleanly when the manifest or map is absent — that is why this PR is green on CI, which has no sibling checkout. Verified by pointing the paths at `/nonexistent`.
- **Use** is unaffected: the committed `data/*.rda` already carry everything.
- Only **regeneration** of `data/` needs the sibling repo, and that already needed it for the GeoPackage.

`data-raw` now aborts with an explanation rather than readr's bare `/path/label_alias_map.csv does not exist.`, naming what publishes the file, the env var to override, and the merge order.

lbm364dl:
## What CI does and does not verify

Worth stating explicitly, because this PR has been green on CI throughout while **eight of its assertions never ran there.**

`inst/scripts/check_like_ci.R` now names every skipped test rather than reporting a count. Under CI conditions:

```
testthat suite   PASS 4539, FAIL 0, WARN 2, SKIP 20
  20 test(s) SKIPPED — 8 needing the upstream contracts, 12 needing external data
    [upstream] test_polities_upstream_contract.R: the embedded polity table matches the upstream manifest
    [upstream] test_polities_upstream_contract.R: the embedded copy agrees with upstream about WHICH rows are dead
    [upstream] test_polities_upstream_contract.R: no polity resolution targets a dead upstream polity
    [upstream] test_polities_upstream_contract.R: no polity lacking geometry claims one, beyond the upstream backlog
    [upstream] test_polities_upstream_contract.R: every identity field matches upstream, row by row
    [upstream] test_upstream_faostat_agreement.R: no area conflicts with upstream on overlapping years
    [upstream] test_upstream_faostat_agreement.R: every upstream-matched area is present in the crosswalk
    [upstream] test_upstream_faostat_agreement.R: the known completeness gaps do not grow
```

**Why:** `lbm364dl/whep-polities` is a **private** repository, so CI cannot clone it. Every one of those eight compares this package against a file only that repo publishes. Closing the gap needs a cross-repo token — a secrets decision for you, not something to arrange from a branch.

**What I did about it instead.** The dead-routing assertion — the defect that started this whole integration, 24 area codes resolving to withdrawn polities — turned out to be checkable *without* upstream, because `polity_area_crosswalk` carries `polity_code` and `polities` carries `wiki_status`, both embedded here. That is now `test_crosswalk_polity_agreement.R`, running unconditionally, along with the alias/polity agreement checks. I audited the remaining eight and none can be moved: each genuinely needs a file from the other repo.

So the split is now deliberate and documented: **in-package invariants are enforced on CI; cross-repo drift is enforced only where the sibling checkout exists.** If you want the second half enforced too, a `WHEP_POLITIES_TOKEN` secret plus a checkout step in `test-coverage.yaml` would do it — the tests already take their paths from environment variables.

lbm364dl:
## Superseded by epic #458 — but read this before closing

The polity migration is now tracked as **#458**, decomposed into 18 independently shippable children. This PR is intended to be closed in favour of those.

**Several children are straight ports of work already written here.** Closing this PR does not lose it — each child names the files. Recover with `git diff origin/main...polities-integration -- <files>`:

| child | what to port from this branch |
|---|---|
| #389 | `R/polities.R` — `resolve_polity_label()`; `data-raw/table_mappings.R` — `polity_label_aliases` loading/validation; `R/sources.R` — `.add_reporter_polity()` |
| #207 | `R/energy_co2_extension.R` — `.warn_areas_gleam_cannot_group()` |
| #404 | `inst/extdata/cow_to_lpjml.csv` + `tests/testthat/test_cow_to_lpjml_area_identity.R` |
| #401 | `data-raw/balance_coefficients.R` — ISO3 resolution for `urban_n_reference` |
| #399 | `data-raw/harmonization_tables.R` — the repair-on-read helper for mojibake |
| #406 | `data-raw/harmonization_tables.R` — `fill_adb_region()` |
| #470 | five small fixes: fodder, gdp-pop, residue warning, stock share, LUH2 micro-territory warning |
| #424 | `tests/testthat/test_exported_polity_column_boundary.R` — but **inverted**, see below |

**#424 inverts rather than ports.** This PR added a test that deliberately pins 22 exports as carrying *no* polity columns. Under the project rule that every territorial row carries a polity code string, that pin is the wrong call — 19 of the 22 should gain the columns. Two footprints (`build_grazing_feed_footprint`, `build_land_balance_footprint`) have no `year`, so resolving a polity there means choosing one, which is a modelling decision.

**Also worth keeping from this branch, unrelated to identity:** the 14 local CI gates in `inst/scripts/check_like_ci.R`, the magnitude baseline in `inst/scripts/compare_cbs_totals.R` + `cbs_totals_baseline.csv`, and `inst/scripts/compare_datasets_to_ref.R`. Nothing in #458 covers those and they caught real defects — including whep#425, which is now fixed and merged as #429.

**Open against this branch:** #444 — its column totals move `feed` +7%, `other_uses` +3%, `production` +2% against `main`, decomposed to the crosswalk (75%) and `regions_full` (25%). That investigation should be read before any child re-lands crosswalk changes, since the same movement will reappear.

lbm364dl:
Closing, as this PR's own last comment proposed: **superseded by epic #458**, which decomposes it into independently shippable children.

Closing does not lose the work. Every child names the files it ports, and the branch stays recoverable:

```bash
git diff origin/main...polities-integration -- <path>
```

Where the pieces went, and where they stand today:

| this PR's code | child | status |
|---|---|---|
| `R/polities.R` `resolve_polity_label()`, `polity_label_aliases` | #389 | PR #479 (draft) |
| `R/build_production.R` stock shares, gdp/pop, residue + LUH2 diagnostics | #470 | PR #481 (draft) |
| `fabio_row_prefix` un-fold | #459 | PR #480 (draft) |
| `R/energy_co2_extension.R` `.warn_areas_gleam_cannot_group()` | #207 | PR #475 |
| `R/footprint_paths.R` 17-area legacy-code note | #471 | open |
| `test_exported_polity_column_boundary.R` | #424 | inverts rather than ports it |

Two things worth stating plainly, since they are the reason this could not merge as one unit:

1. **#444 is unresolved.** The retracted neutrality claim — `feed +7.0%`, `other_uses +3.1%`, `production +2.0%`, `domestic_supply +1.9%`, localised to the rest-of-world bucket where feed moves 28.7x — still has no identified cause after five rejected hypotheses. Nothing should merge on the argument that it moves no numbers.
2. **Mechanically unlandable**: 136 commits behind `main`, 24 conflicts including `R/polities.R`, `data-raw/table_mappings.R` and five `data/*.rda`.

One correction for anyone reading this thread later: the merge-order note above is now satisfied — [lbm364dl/whep-polities#39](https://github.com/lbm364dl/whep-polities/pull/39) **merged**. But the published `label_alias_map.csv` has since moved to **881 aliases over 464 labels**, against the 869/470 this branch embedded, so a consumer of it must regenerate rather than reuse the snapshot here.

Issue #367, opened 2026-07-27
[SOC #123] Turnkey runs dominated by full-range build_primary_production + global gridding/livestock, not the SOC engine


lbm364dl:
Profiled from the first turnkey version in #123, and confirmed by a completed
real-data run.

## Finding
For a turnkey `build_carbon_balance()`, the cost is almost entirely in
**assembling the carbon inputs**, not in the SOC engine:

- A real Iberia-scoped end-to-end run: input read **768 s**, SOC engine
  (equilibrium + march + finalise) only **166 s**.
- The read is dominated by `get_primary_production()` → `build_primary_production()`
  which **always rebuilds the full 1850-2023 global series** regardless of the
  requested `years` (the year filter is applied *after*). Scoping it to
  2000-2003 alone builds in ~105 s vs. the full-range slog.
- The remaining read cost is the **global** manure `build_livestock_nutrient_flows()`
  + the national→grid gridding in `.sci_to_grid()`, which do not shrink with the
  year window (spatial, not temporal).

This **supersedes the carbon-path assumption in #353**: that issue targeted the
SWC/water read path, but that path is already fast (SWC read ~6 s via the
existing `ncvar_get(start=, count=)` slice), and the `.cb_march` O(cells^2) item
from #353 is already fixed in this PR.

## Directions (in payoff order)
1. **Year-aware `get_primary_production()`**: thread the requested year range into
   `build_primary_production(start_year=, end_year=)` and put the range in the
   cache key. Guard the historical back-cast (truncating `start_year` must not
   silently change extension results). Cross-cutting: every caller shares the
   `primary_prod` cache, so this needs care.
2. **Disk-persist `.build_cache`** (primary_prod / cbs_built) across sessions, so
   the expensive build happens once rather than per session. Needs a
   staleness/invalidation design.
3. **Spatial windowing** for `build_soil_carbon_inputs` / `.sci_to_grid` so a
   regional request does not pay global gridding + livestock-flow cost.

## Notes
- Not SOC-specific: `build_primary_production` / `get_wide_cbs` are shared package
  infrastructure (footprint + every extension). Best done in its own PR with a
  before/after timing + identical-output check, not as a rider on #123.
- Engine correctness is unaffected and already verified end-to-end on real data.

lbm364dl:
Direction 1 is not only a latency problem — it is a hard memory ceiling, and it
has been breaking the r-universe check on **every** build since 2026-07-27.

## Evidence

[...]

PR #383, opened 2026-07-28
perf(reads): unblock + speed up the global input-reading path (#367)

lbm364dl:
## Input-reading path: make global runs possible, then fast (#367, input half)

Input-reading counterpart to #376 (SOC-engine half). Found while running a **global** `build_carbon_balance` end-to-end.

### Part 1 — unblock global runs
A global `build_commodity_balances()` **crashed** in `Combining CBS sources`:
`fcoalesce(FAOSTAT_prod, FAOSTAT_FBS_New, FBS_Old_scaled, other_mean)` mixed integer and double source columns (the pivoted sources inherit the raw `value` type; the scaled/mean sources are doubles), which `fcoalesce()` rejects. The in-package fixtures are all-double, so this only tripped on real global data — **global CBS builds (and thus global carbon-balance runs) were impossible**. Fixed by coercing sources to `double` before coalescing. Regression test added.

### Part 2 — vectorise the source-selection hot paths
Profiling (line-level `Rprof`) showed the ~50-min global read was dominated by **per-group `dplyr::case_when`**: `.best_prod_source` / `.best_cbs_source` were called once per group, each re-parsing an 11-branch `case_when`, over millions of `(year, area, item)` groups. Rewrote all three sites (`.add_historical_yields`, `.prepare_historical_cbs`, `.cbs_extend_historical`) to **rank every row once, then pick the best source per group by ordering** — no R-function-per-group. Removed the now-dead `.best_cbs_source`. Each change verified **byte-identical** to the per-group logic across NA / tie / unrecognised / year-dependent-historical edge cases.

### Part 3 — vectorise the gridding carbon-input aggregation
`.sci_sum_components` (the per-cell aggregation in `build_soil_carbon_inputs`) ran four filtered `sum(.data$c_mass_mg[.data$input_type == "…"])` aggregations per cell via `dplyr::summarise`, re-resolving the `.data` pronoun over millions of global cells (~33% of that read). Rewrote it to pre-mask by `input_type` once with `data.table::fifelse`, then plain GForce grouped sums — **byte-identical** (`identical()` TRUE across missing-type / NA / tie cases; `data.table` `by=` preserves dplyr `.by`'s first-appearance order). `build_soil_carbon_inputs(years = 2000:2002)`: **1244 s → 428 s** (~2.9×).

### Measured impact (global, 1850–2023)

| Stage | Before | After |
|---|---|---|
| `build_primary_production` | (in the ~50-min read; `Adding historical yields` alone 26 min) | **2.6 min** |
| `build_commodity_balances` | crashed in source combining | **4.0 min** |
| **prod + CBS total** | ~50 min **and crashing** | **6.6 min** |

(Representative micro-timings: `build_primary_production(1940,1975)` 327s→43s; `build_commodity_balances(pp)` 312s→139s.)

### Deferred (tracked, non-blocking)
- **#388** — the residual CBS cost is now distributed data.table `setkey`/`merge` across the `.fix_cbs` balancing chain (no single-line anti-pattern); a deeper key-once/`on=`-join restructuring is tracked there. Not forced here because it can't be made byte-identical-safe as a quick change.
- **#390** — after the `.sci_sum_components` fix, the gridding/livestock read's residual cost is likewise distributed data.table `forderv`/`bmerge` (the `.sci_to_grid` many-to-many spread + livestock-flow merges); same nature as #388, deferred for the same reason.

### DIFF
--- a/R/build_cbs.R
+++ b/R/build_cbs.R
@@ -1679,13 +1679,18 @@ build_processing_coefs <- function(
     )
   ]
 
-  # Source selection: Primary > FBS_New > scaled FBS_Old > other
+  # Source selection: Primary > FBS_New > scaled FBS_Old > other. Coerce every
+  # source to double first: the pivoted FAOSTAT_prod / FAOSTAT_FBS_New inherit
+  # the raw `value` type (integer for some global sources), while FBS_Old_scaled
+  # and other_mean are computed doubles, and fcoalesce() aborts on a mixed
+  # integer/double set. Production quantities are continuous, so double is the
+  # correct common type.
   wide[,
     value := data.table::fcoalesce(
-      FAOSTAT_prod,
-      FAOSTAT_FBS_New,
-      FBS_Old_scaled,
-      other_mean
+      as.double(FAOSTAT_prod),
+      as.double(FAOSTAT_FBS_New),
+      as.double(FBS_Old_scaled),
+      as.double(other_mean)
     )
   ]
   wide[,
--- a/R/build_cbs.R
+++ b/R/build_cbs.R
@@ -1047,13 +1047,22 @@ build_processing_coefs <- function(
     "item_cbs_code",
     "element"
   )
-  out <- dt[,
-    .(
-      value = mean(value, na.rm = TRUE),
-      source = .best_cbs_source(source, year[1L])
-    ),
-    by = key_cols
+  # Best source per group WITHOUT calling .best_cbs_source() per group (which
+  # re-ran .cbs_source_rank()'s case_when for every group). Rank every row once,
+  # then take the mean value and the lowest-ranked (then alphabetical) non-NA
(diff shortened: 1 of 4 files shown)


lbm364dl:
### Measured global timing (1850–2023, with the fixes in this PR)

| Stage | Time | Rows |
|---|---|---|
| `build_primary_production` | **2.6 min** | 6,170,595 |
| `build_commodity_balances` | **4.0 min** | 12,646,249 |
| **Total (prod + CBS)** | **6.6 min** | — |

Before this PR the global read was ~50 min and crashed in `Combining CBS sources`; `Adding historical yields` alone was **26 min**. The crash is fixed and the two source-selection anti-patterns are vectorised, bringing the global prod+CBS build to **6.6 min**. Remaining distributed data.table `setkey`/`merge` cost in the CBS balancing chain is tracked in #388.

PR #398, opened 2026-07-29
fix(soc): repair cell × polity crosswalk, stop silent NaN drainage (#381)

### LINKED ISSUE #381: build_cell_polity_fraction() is orphaned and broken; its output grid also injects NaN drainage into build_water_balance()
Found while closing #340 end to end (PR #368). Three separate defects around the cell × polity crosswalk, in increasing order of consequence.

## 1. `build_cell_polity_fraction()` is never called

`prepare_spatialize_all()` invokes Sections 1, 2, 3, 3b, 4, 4b, 5, 6, 7, 8, 9, 9c, 9d, 10, 10a and 11 — **not 1b**. Nothing calls `build_cell_polity_fraction()`, and nothing saves its return value, so the pipeline cannot produce `cell_polity_fraction.parquet`. Its own header comment claims it *"Writes cell_polity_fraction"*, which it does not — it returns a tibble.

Consequence: `WHEP_POLITY_FRACTION_PATH` has nothing to point at on a fresh machine, so `build_cell_polity()` aborts, so every Module C function that takes `data$cell_polity` as a required input is unreachable — `build_water_balance()`, `get_soc_climate_drivers()`, `build_n_deposition()`, `build_urban_n()`.

## 2. It is broken against the current `polities` schema

```r
polities <- whep::polities
...
countries$area_code <- dplyr::left_join(
  tibble::tibble(iso3c = iso3c),
  dplyr::select(polities, "iso3c", "area_code"),   # <- area_code does not exist
  by = "iso3c"
)$area_code
```

`whep::polities` carries `polity_code`, `polity_name`, `iso3_code`, `cow_code`, `iso3c`, … but **no `area_code`**. Calling the function dies with a `tidyselect` subscript-out-of-bounds error. It was written 2026-06-19 (`9da989c4`), before the polities restructure removed that column, and being orphaned meant nothing ever caught the drift.

Workaround used to generate the layer: take `iso3c → area_code` from `inst/extdata/regions.csv` (209 rows). Whether that is the *intended* lookup post-restructure is a question for whoever did the refactor — it may be that the function should now map via `polity_code` instead.

Generated output, for reference:

```
countries matched to area_code:  201 of 258 Natural Earth features
rows:                            68,527
distinct cells:                  64,439
multi-polity (fractional) cells:  3,924
per-cell polity_frac sums:       min 1, max 1   (exact partition)
```

Note only 201 of 258 features matched; the 57 unmatched are mostly non-sovereign territories absent from `regions.csv` and contribute no cells at all. Worth enumerating before this is trusted.

## 3. The crosswalk grid ≠ the LPJmL grid, which silently injects `NaN`

This is the one with teeth. The crosswalk covers **64,439** Natural Earth land cells; LPJmL simulates **58,795**. `.wb_attach_polity()` does

```r
dplyr::inner_join(terms, crosswalk, by = c("lon", "lat"))
```

so every crosswalk cell survives, including cells LPJmL never simulated — which arrive with `NA` drainage.

Measured on the 1901-2023 run, 2009-2023, via `build_water_balance(resolution = "grid")`:

| | |
|---|---|
| rows per year | 68,514 |
| finite `drainage_mm` | 62,784 |
| **non-finite** | **5,730 per year** |
| of those, outside the LPJmL land mask | **5,730 (100%)** |
| non-finite *inside* the mask | **0** |
| LPJmL cells covered | 58,791 of 58,795 |

So the balance itself is sound — inside LPJmL's grid every cell is finite in every year. But the returned tibble carries ~5,730 cell-years of `NaN` drainage per year, and that flows straight into `calculate_n_leaching(method = "meisinger_drainage")`. That is precisely the silent-NaN propagation #340 was about, reappearing from a different cause: a grid mismatch rather than missing climate.

Also: 4 LPJmL cells have no polity in the crosswalk and are dropped — small islands, presumably, but they should be accounted for rather than silently lost.

## Suggested fixes

- [ ] Call Section 1b from `prepare_spatialize_all()` and persist the result (`.save_parquet(..., "cell_polity_fraction")`), or delete the function if it is superseded
- [ ] Fix the `iso3c → area_code` join against the current `polities` schema; add a test so schema drift fails loudly
- [ ] Decide the grid contract: either restrict the crosswalk to the simulated grid, or have `.wb_attach_polity()` warn/drop when a crosswalk cell has no model data — silent `NA` is the one option that should be off the table
- [ ] Report the 4 unmatched LPJmL cells and the 57 unmatched Natural Earth features instead of dropping them quietly

Context: the #340 drainage result is unaffected — `drainage_mm` is finite for all 58,795 simulated cells across 2009-2023, and 2009 reproduces the reference mean (209.7 mm with a one-polity-per-cell crosswalk, 214.1 mm with the fractional one).

lbm364dl:
## What #381 was

The cell × polity crosswalk (`cell_polity_fraction.parquet`) and the country
grid it derives from could no longer be rebuilt from WHEP code. The `polities`
restructure removed the `area_code` column, and both rasterisation sections in
`inst/scripts/prepare_spatialize_all.R` still looked it up on `polities`, so
`prepare_country_grid()` (Section 1) and `build_cell_polity_fraction()`
(Section 1b) crashed. On top of that, water-balance rows in cells LPJmL never
simulated carried non-finite drainage that silently propagated as NaN.

## What this PR does

**1. Repair the country-grid generation (both sections).**
Added a shared `.spatialize_area_lookup()` reading `inst/extdata/regions.csv`
(iso3c → area_code), replacing the removed `polities$area_code`. Both
`prepare_country_grid()` and `build_cell_polity_fraction()` use it.

**2. Fix the terra NoData sentinel leak.**
terra writes integer NoData as `-2147483648`, which is not NA, so the old
`filter(!is.na) |> as.integer` left ocean cells in and only NA'd them at
coercion. Coercing to integer *before* the NA filter drops them.

**3. Stop silent NaN drainage.**
`.wb_warn_uncovered_cells()` / `.wb_drop_unsimulated()` warn about and drop
water-balance rows in cells LPJmL never simulated (non-finite drainage),
instead of letting NaN propagate into footprints.

## Verification (real data, 58,795-cell simulated grid)

- **`prepare_country_grid()` reproduces the existing pin cell-for-cell**:
  58,795 cells, 0 NA, `all.equal()` identical codes vs `spatialize-country-grid`.
- **`build_cell_polity_fraction()`**: 58,791 cells, 0 NA, 0 cells outside the
  grid, per-cell fractions sum to 1.
- The fixed `country_grid` also feeds the existing `write_lpjml_static_inputs()`
  NetCDF writer, which then emits a correct `cow_gadm_30arcmin.nc` (0 area_codes
  unmatched by `cow_to_lpjml.csv`; 99.2% agreement with the LandInG reference
  where WHEP has a cell) — no gadm-bin dependency.
- Full `test_water_balance.R` suite green; `#381 guard` test now targets
  `regions.csv`.

Closes #381

### DIFF
--- a/R/water_balance.R
+++ b/R/water_balance.R
@@ -450,7 +450,47 @@ get_soc_climate_drivers <- function(
            {.field area_code}, {.field polity_frac}, {.field cell_area_ha}."
     ))
   }
-  dplyr::inner_join(terms, crosswalk, by = c("lon", "lat"))
+  .wb_warn_uncovered_cells(terms, crosswalk)
+  dplyr::inner_join(terms, crosswalk, by = c("lon", "lat")) |>
+    .wb_drop_unsimulated()
+}
+
+# Warn when simulated cells have no polity in the crosswalk (they are dropped by
+# the join, so their water would silently vanish from the polity totals). Keyed
+# on the water terms carrying a real (finite) drainage value.
+.wb_warn_uncovered_cells <- function(terms, crosswalk) {
+  sim <- terms |>
+    dplyr::filter(is.finite(.data$drainage_mm)) |>
+    dplyr::distinct(.data$lon, .data$lat)
+  missing <- dplyr::anti_join(sim, crosswalk, by = c("lon", "lat"))
+  if (nrow(missing) > 0) {
+    cli::cli_warn(c(
+      "!" = "{nrow(missing)} simulated cell{?s} ha{?s/ve} no polity in the
+        cell_polity crosswalk and are dropped from the polity aggregation.",
+      i = "Extend the crosswalk to cover the simulated grid to retain them."
+    ))
+  }
+}
+
+# Drop cell-polity rows that carry a polity label but no model data: crosswalk
+# cells outside the simulated (LPJmL) grid join in with non-finite drainage, and
+# passing that NaN downstream silently poisons leaching (see #381). Warn and
+# drop rather than propagate the NaN or silently discard it.
+.wb_drop_unsimulated <- function(joined) {
+  unsimulated <- !is.finite(joined$drainage_mm)
+  if (any(unsimulated)) {
+    cells <- joined[unsimulated, ] |>
+      dplyr::distinct(.data$lon, .data$lat)
+    cli::cli_warn(c(
+      "!" = "Dropped {sum(unsimulated)} cell-polity row{?s}
+        ({nrow(cells)} cell{?s}) with a polity label but no LPJmL model data
+        (crosswalk cells outside the simulated grid).",
+      i = "Restrict {.code data$cell_polity} to the simulated grid to avoid
+        the non-finite drainage they would otherwise inject."
+    ))
+    joined <- joined |> dplyr::filter(is.finite(.data$drainage_mm))
+  }
+  joined
 }
 
 # Stamp method_water (including the realized blue_green method), select the grid
--- a/inst/scripts/prepare_spatialize_all.R
+++ b/inst/scripts/prepare_spatialize_all.R
@@ -868,6 +868,22 @@ cft_to_pft <- c(
 #
 # Rasterize NaturalEarth to a 0.5-degree grid with WHEP area_code.
 
+# iso3c -> area_code from inst/extdata/regions.csv. whep::polities dropped its
+# area_code column in the polity restructure, which silently broke both
+# rasterisation sections against the current schema (#381); this is the
+# canonical replacement, shared so the country grid and the cell x polity
+# crosswalk stay consistent. Verified to reproduce the existing 58,795-cell
+# country_grid pin cell-for-cell with identical codes.
+.spatialize_area_lookup <- function() {
+  path <- system.file("extdata", "regions.csv", package = "whep")
+  regions <- utils::read.csv(path, stringsAsFactors = FALSE)
+  lookup <- regions[
+    !is.na(regions$iso3c) & !is.na(regions$area_code),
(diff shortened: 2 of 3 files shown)


lbm364dl:
### Verified by actually running `build_cell_polity_fraction()` on real data

Running it (Natural Earth shapefile + the real `country_grid`) surfaced — and fixed — a real problem beyond the unit tests:

- The crash is gone (was dying on the removed `polities$area_code`).
- Mapping `iso3c → area_code` via `regions_full` **over-matched** — it pulled in Antarctica (`ATA→30`) and other non-simulated territories, ballooning the crosswalk to **92.8k cells** (25.3k Antarctic) and leaking a few NA codes. The `regions.csv` workaround in the issue had quietly excluded these.
- Fix (the issue's preferred option): thread the simulated `country_grid` into the builder and **restrict the output to it**, drop unassigned subcells before normalising, and renormalise.

**Final crosswalk, measured on real data:**

| check | result |
|---|---|
| cells | **58,791** (= 58,795 simulated − the 4 with no polity, per #340/#381) |
| cells outside the simulated grid | **0** |
| Antarctic cells | **0** |
| NA area_code | **0** |
| per-cell `polity_frac` sum | **exactly 1** |
| fractional (multi-polity) cells | 3,852 |
| country features matched | 238/258 (unmatched reported, not silently dropped) |

The runtime `.wb_attach_polity()` warn-and-drop stays as a belt-and-suspenders guard, but with the build-time restriction it should now be a no-op on a correctly-built crosswalk.

Issue #147, opened 2026-07-22
n_prov_destiny: seed mass subtracted from residue/grass rows, not just the product


lbm364dl:
## What

`.remove_seeds_from_system()` (`n_prov_destiny.R:473-487`) left-joins `Seeds_used_MgFM` by `(Year, Province_name, Item)` onto `grafs_prod_combined`, which carries multiple `prod_type` rows per item (`Product`, `Residue`, `Grass`). The subtraction `production_fm = production_fm - Seeds_used_capped` and the 50% cap run **per row**.

## Why it matters

Seed mass (which is grain, i.e. the product stream only) is removed from the residue and grass rows of the same crop as well. For a crop with a Product row and a Residue row, up to the seed amount is removed from *each* → ~2× the real seed use removed, and residue biomass (hence residue-N routed to soil/feed) is understated.

## Suggested fix / check

Restrict the seed subtraction to `prod_type == "Product"` rows before the join/cap. Add a mass-conservation test: total removed == seed used (once), residue biomass unchanged.

---
_From a static read-only code audit (2026-07-22). Line references from source reading, not verified by running — confirm with a test before fixing. Related to #131._

lbm364dl:
## Triage (sweep, 2026-08-27): ALREADY FIXED — with a regression test

The quoted defect no longer exists on `origin/main`. `.remove_seeds_from_system()`
(`R/n_prov_destiny.R`) now gates both the subtraction and the 50% cap on
`prod_type == "Product"`:

```r
# Seed only comes off the Product row, not Residue/Grass rows for the
# same item.
Seeds_used_MgFM = dplyr::if_else(prod_type == "Product", coalesce(Seeds_used_MgFM, 0), 0),
Seeds_used_capped = dplyr::if_else(prod_type == "Product", <50% cap>, 0),
production_fm = production_fm - Seeds_used_capped
```

Fixed by **`938ec8c7`** ("fix(grafs): fix seed, FCI, manure-loss, and
cropland-surplus accounting bugs", 2026-08-18), verified an ancestor of
`origin/main`.

**A regression test exists and names this issue.** `tests/testthat/test_n_prov_destiny.R`:

- `".remove_seeds_from_system takes seed only from the Product row"` — comment cites
  #147; asserts Product 100→90, Residue 80 unchanged, Grass 60 unchanged, and
  `sum(before) - sum(after) == 10` (seed removed exactly once).
[...]

PR #416, opened 2026-07-30
Reconcile with main, add typology sensitivity/decomposition analysis, processed-item accounting change

AliceBeckmann:
## Summary
- Recovers previously uncommitted typology sensitivity analysis, LMDI decomposition of N surplus/losses (cropland, semi-natural, livestock, urban), circularity index, and typology panel plots.
- Moves processed-item calculation into create_n_prov_destiny() directly, instead of relying on an external, R script/file (previously the processed_prov_fixed pin, built outside the package): processing shares and processed amounts are now calculated directly from the processing_coefs pin. As part of this, processed items (wine, oil, flour...) now subtract their share from the primary crop's own production instead of being added on top of it. A remaining open question (zero-N processed items like wine/olive oil) is tracked in a separate issue.
- Reconciles alice/typologies with main, resolving conflicts in R/grafs_plot_df.R, R/input_output_plots.R, R/intensification_specialization_plot.R, R/typologies_spain_plot.R, NAMESPACE, R/utils.R, and inst/extdata/whep_inputs.csv.

## Data
inst/extdata/whep_inputs.csv keeps newer versions of 4 pins reflecting this branch's own work: n_balance_ygpit_all, n_excretion_ygs, npp_ygpit, intake_ygiac. All other pin entries match main.

## Test plan
- [x] rcmdcheck::rcmdcheck() — 0 errors, 0 warnings, 0 notes
- [x] devtools::test() — full suite green (4071 passing; 0 real failures)

## To do
- [x] Open an issue to discuss processed items (wine, olive oil, etc.) N accounting


lbm364dl:
Pushed three fixes onto this branch (plus a sync with `main`). Flagging them here rather than leaving them to be discovered, because **the first one changes the numbers this branch produces**.

### 1. The processed-item substitution now conserves N (#443, merged)

`96bcbe90`'s commit message said it "correctly conserve[s] production N". Measured against the pins, it did not — and it failed in **both** directions, not only the zero-N one already flagged as an open question.

`.expand_processed_items()` applied *every* output's `cf` to the same input mass. Σcf per input item for Spain in 2000: maize **5.22**, wheat **5.11**, barley **4.90** (beer and starch are water-diluted), against olives 0.24. So a tonne of maize sent to processing yielded 5.22 t of product, while grapes → wine and olives → oil destroyed N. Net **+1.875 Mt N (+1.56%)** over 1860–2023, worst year **+7.84%**.

This is not a defect in the pin: `R/supply_use.R:33-39` already documents those factors as the outputs of a *virtual* process, i.e. not co-product shares of one physical process.

**What moved.** Totals vs. before: 1860 **+11.1%**, 1900 +5.8%, 1950 +5.7%, 2000 −0.35%, 2020 −1.11%. In 2000, Cropland-origin N **−7.7%** and imports **+3.5%**; exports −9.4%. Any figures already generated from this branch — the LMDI panels and typology plots especially — are stale and want regenerating.

The residual N that near-zero-N outputs cannot account for now **stays with the primary crop** rather than vanishing. That keeps the balance exact but attributes grape pomace and olive cake to table-grape destinies. Where it should really go is #432, which is the issue your to-do list asked for.

### 2. `.finn_flow_matrix()` was built transposed (#430)

A hand-computed column-major index, `(from - 1) * n + to`, lands on `[to, from]` — the transpose of the documented `[from, to]`. `.calculate_finn()` then read `colSums()` as each compartment's **outflow** where its own comment says inflow. Since exports and food leave the four compartments, those differ everywhere.

Mean FCI barely moves (0.1105 → 0.1102) but individual province-years shift up to **0.035 absolute** — roughly half the typical modern value — and the decadal pattern tilts in opposite directions at the two ends (1860s −3.9%, 1920s +1.3%), which is exactly what `plot_finn_circularity()` compares.

### 3. `plot_typology_indicators_panel()` did not work as shipped (#431)

Bare `(p_ext + p_fci) / (p_pol + p_int)` uses patchwork's operators, which patchwork **exports** and which therefore only take effect once it is *attached*. patchwork is a Suggests and whep never attaches it, so the function errored with `Can't add p_fci to a <ggplot> object` for any ordinary user. `.n_losses_two_panel()` already did this correctly via namespaced `wrap_plots()`, which is why the other panel functions were fine.

### Also included

- **`Europe_FAO_completed.xlsx` (16.5 MB) removed** (#434). It is a 19-sheet, 35-country workbook of which the package read two sheets and 253 of 9,113 rows. Replaced by a 199 KB CSV extract with `data-raw/europe_fao_spain_trade.R` to regenerate it; output verified identical. Installed package **24.2 Mb → 8.6 Mb**. The line count in this PR grows because the CSV is 6,202 rows of text replacing a binary — net repo win. `readxl` moves to Suggests as a consequence, and the documented coverage is corrected to 1849–1960.
- **Typology rules de-duplicated** (#438). They existed in full in both `create_typologies_spain()` and `.reclassify_typology()`; this branch had just moved two thresholds in one copy and mirrored them by hand in the other. Now one `.typology_thresholds()` + `.classify_typology_base()`, proven equivalent over 2,239,488 grid cells including every boundary value, 0 mismatches.
- **All 33 new examples are runnable** (#435). They were commented out (`#' # decompose_cropland_surplus()`), which passes `R CMD check` while exercising nothing — and is why (3) went unnoticed. 16 use inline fixtures; 13 use the documented `example = TRUE` pattern with fixtures sampled from real output; 4 gained optional arguments for frames they were fetching internally (`typologies`, `area_df`/`typo_df`, `panel_data`, `baseline`), each defaulting to the previous reader, so existing calls are unchanged.
- **92 new tests** across `circularity_index`, `typologies_sensitivity` and `validate_national_trade`, which had none.
- **`.grafs_last_year()`** replaces three scattered `<= 2021` cutoffs, and `.warn_processing_coverage_gap()` warns instead of silently emitting 2022–23 with no processing at all (#433).

### Verification

`R CMD check --as-cran`: 0 errors, 0 warnings, 1 pre-existing NOTE; `checking examples ... OK` (67s). `devtools::test()`: 4667 passing, 0 failures. `lintr`: 0. `air format .`: clean. #443 was 9/9 green in CI across all five platforms.

### Still open, and yours to decide

#432 (where the unaccounted N goes), #433 (`processing_coefs` ends 2021), #440 (a stale `_old` Excreta graft that this branch's pin bump widens), #438's second half (the two specialised-cropping branches are not complementary, so a high-synthetic low-yield crop province falls through to "Disconnected crop-livestock"), and #428 (reuse `build_supply_use()`'s process definition so the Σcf problem cannot recur).

Happy to revert or rework any of it.

lbm364dl:
**Correction to my comment above.** I listed #440 (a stale `Excreta` graft inflating manure N ~2×) as an open item affecting this branch. That was wrong — it is **already fixed on `main`** by `85ee3c32`, and `alice/typologies` does not reintroduce it. I filed that issue from stale notes without checking the current code, and have closed it.

Concretely, this means the thing I flagged as a concern is not one: **the manure compartment of the LMDI decomposition is computed on the correct `Excreta` values**, straight from the `n_balance_ygpit_all` pin, with no graft. Nothing in this PR's manure figures needs discounting on that account.

Everything else in the comment above stands — the measurements, the three fixes, and the remaining open items (#428, #432, #433, and the second half of #438).

AliceBeckmann:
Thank you @lbm364dl!

For #428 and #438 I was writing a comment in the respective issue. 

#432 (unaccounted N from wine/olive oil/sugar). For now, the current behaviour is kept as it is. When grapes, olives, or sugar beet get processed into wine, oil, or sugar, most of the crop's N doesn't show up in the output (those products are nearly N-free), so that leftover N stays attached to the primary crop instead of disappearing. The alternative would be creating separate items for the by-products (grape pomace, olive cake, beet pulp) so that leftover N has its own place to go. We can't do that here now, since it would mean adding new items that don't exist in our current item list, and we're not in a position to introduce new ones for the first paper. Since there's no better destination available, I'm treating the current behaviour as it is. For the next paper (nutrient coefs etc.), we should think about adding these by-product items. It is explained and tracked as future work in #445.

#433 (processing_coefs stops at 2021, but our data now runs to 2023). Fixed by copying the last known value from 2021 forward into 2022 and 2023, for both processing_coefs and population_yg. I opened issue #449 to discuss whether this is good enough, or whether it's worth it to look for source data for 2022-2023 instead.

AliceBeckmann:
I ran a review pass over this branch looking for correctness bugs, then verified each one against the real dataset (not just test fixtures) before fixing it. Pushed in 938ec8c7.

Fixed, with a measurable effect on results:

- Manure loss fraction (decompose_manure_losses()): the denominator was excretion from LU-covered species only, while the numerator (applied manure) covers all species. Fixed to match scopes — shifts loss_frac up by ~0.1-0.9 percentage points, most visible in 2014-2023 (driven by aquaculture).

- Cropland surplus/size (decompose_cropland_surplus(), decompose_specialization_cov()): items with real cropland area/input but no tracked output (mainly sugar beet, sunflower, cottonseed, rapeseed — items almost entirely substituted into processed products) were silently dropped instead of counting as surplus. Now kept in a separate no_tracked_output bucket. ~1.3-1.6% of national input/area.

- Pollution panel: was filtering out legitimate negative surplus (net N depletion, common in extensive systems). ~1% of province-year rows affected.


Fixed, but currently no effect on real results (confirmed by running before/after on the actual data):

- Seed was being deducted from Residue/Grass rows, not just the Product row — matches main's own #147 fix, but no seeded item in this dataset actually collides across prod_type, so 0 rows changed.

- Fish/Agro-industry weren't counted as external inputs to the Finn Cycling Index — but origin never actually takes those values in the current data (only box does), so this is a dormant fix for now.

- A join-then-divide NA bug in the external-dependency panel — doesn't occur in practice either.

Smaller fixes: added a missing sign-change warning to decompose_manure_losses() (matching its siblings), added a warning (not a silent pass) when the urban-losses additive identity breaks, fixed a wrong default reference-year in two plotting functions, and removed a periods argument that two of plot_finn_circularity()'s three panels were silently ignoring.

All touched functions pass their tests, lintr, and air format; rcmdcheck is clean aside from a pre-existing PDF/LaTeX warning unrelated to this change. Also removed a stray invalid file (R/circularity_index) that was tripping up the build.

@lbm364dl this should be ready for reviewing now, so we can get this PR closed.

lbm364dl:
## Second review pass — verified against real data, not CI

Ran the whole gate locally rather than reading the badges, and re-ran
`create_n_prov_destiny()` on the real pins (4,316,580 rows, 1860-2023) to
check what the code actually does to published numbers.

| Gate | Result |
|---|---|
| `devtools::test()` | **FAIL 0, PASS 9214**, SKIP 9 (all env-gated) |
| `rcmdcheck` (clean export) | **0 errors, 0 warnings, 0 notes** |
| `air format --check .` | clean |
| pkgdown reference index | complete |

I pushed two commits: the merge with main (NEWS.md and `whep_inputs.rda`
conflicted; the `.rda` is a build product so I rebuilt it from the merged CSV
rather than picking a side, and `test_data_raw_freshness.R` passes against the
rebuild) and a docs/NEWS/test-hygiene commit. No behaviour changed.

### The 2023 extension holds up

Removing the 2021 clip is justified: `n_balance_ygpit_all`, `npp_ygpit`,
`intake_ygiac` and `n_excretion_ygs` genuinely cover 1860-2023 now, and
2022-23 are structurally complete — 50 provinces, every origin, every
destiny, ~29.5k rows/year against 29.8k in 2021. `population_yg` is the only
input still ending in 2021, and two years of flat population is fine.

National N does fall hard, 5,097,518 Mg (2021) to 4,607,382 (2022) to
4,201,534 (2023). I traced it: it is in the input pin, not in this code —
synthetic N drops 29.7% while cropland area stays flat at 50.56 Mha. That is
consistent with the 2022 fertiliser price shock, but **it has not been checked
against an independent source**, and it is worth one before those years are
published.

One thing worth knowing: the four analysis eras end at 2010-2020, so 2021-2023
appear in yearly and evolution views but in **no period aggregate at all**.

### Two of the four fixes in 938ec8c7 are inert

I measured each one rather than trusting the commit message:

- **Seed.** Provably a no-op. **Zero** Residue/Grass rows ever match a seed
  rate — they carry different `Item` names, so the join never fires; all
  424,757,488 Mg FM of seed already lands on Product rows. I reverted the fix
  and re-ran the full pipeline: output identical to the last decimal, every
  year and every destiny. Worth keeping as a guard, but the message's claim
  that it "removed ~3x the real seed use" is not true of current data.
- **FCI external inputs.** Also a no-op. `Fish` and `Agro-industry` are `box`
  values, never `origin` values; all 5.4 Mt N of that biomass already enters
  as `origin == "Outside"`, which was already in the external list.
- **Manure loss scope.** Real, small: the excretion base was 0.16%-1.26% too
  low (mean 0.29%), and the gap widens over time, so it tilts the trend.
- **Cropland surplus.** Real, and the important one — see below.

### The `no_tracked_output` change is the significant one

Swapping the `left_join` for a `full_join` recovers rows that were being
dropped silently: **1.32% of cropland area and 1.55% of N inputs** over the
full span, rising from 0.05% in the 1860s to about **5% by the 2020s** — so it
was distorting the trend, not just the level.

They are almost entirely the oilseed complex: sunflower seed (9.6 Mha
cumulative 2010+), rape and mustardseed, cottonseed, sugar beet — exactly the
crops whose output the processed-item change reassigns. That makes this a
**quantified instance of #432**, and I would suggest updating that issue with
these numbers.

Also confirmed real: dropping the `value >= 0` filter recovers 86 of 8,200
province-years, all pre-1950 (3.2% of them in the 1900s), and lowers pre-1950
decade means by up to 6%.

### Cleanups pushed

- **`NEWS.md` had no entry** despite the branch moving published values. Added
  four, each naming what moves numerically.
- **Two `test_that()` blocks were duplicated verbatim** under identical names
  (a merge artifact). Removed the older copy; PASS drops by exactly its six
  assertions.
- **Three documented claims did not match the code**: `origin` was documented
  as taking `Fish`/`Agro-industry` (it never does), `destiny` was missing
  `semi_natural_agroecosystems`, and both processing forward-fill docstrings
  said each series is held from its 2021 observation when they hold from each
  series' own last year — 2002 for coconuts, 2008 for sugar cane, 2018 for
  palm kernels. All three are negligible in Spanish production, so no value
  moves, but the docs should say what the code does.
- `plot_finn_circularity()`'s committed `.Rd` was ahead of its roxygen;
  restored the fuller `@param periods` wording into the source so
  `document()` is idempotent.

### Closes on merge

The earlier fixes all survived the merges intact, so this should close
**#430** (Finn transpose), **#431** (patchwork), **#434** (16.5 MB xlsx now a
194 KB CSV) and **#435** (33 runnable examples; `checking examples ... OK`).

### Recommendation

Mergeable on engineering grounds — everything is green and I verified it
rather than predicting it. But the `no_tracked_output` change is a **science
decision, not mechanical**: it moves every compartment-factor number and the
trend behind them, and it surfaces that ~5% of modern cropland has no tracked
destiny. Green CI says nothing about whether that is the accounting we want,
so this needs @eduaguilera's sign-off on that specific point before merge.

eduaguilera:
Flagging this before the branch merges, from the #449 pin work (#812).

## The code contradicts its own docs

`R/n_prov_destiny.R` reads `whep_read_file("processing_coefs")` — a July 2025
snapshot ending in 2021 — while `.spain_processing_coefs()` twelve lines below
documents its parameter as the output of `get_processing_coefs()`. `main` has
built that table from CBS since `cdebb11b` (#367); the pin now survives only
as the legacy baseline in `compare_global_whep.R`.

That stale pin is also why `.forwardfill_processing_shares()` exists in its
current form — its own description cites "`processing_coefs` currently stops
in 2021", which is the #449 forward-fill, living inside the package.

## #812 removes the need for it

With `faostat-fbs-new` refreshed to the 2025-10-28 FBS release,
`get_processing_coefs()` for 2023 goes from 359 rows and 108 Mt to **5,986
rows and 1,548 Mt**, all 45 input items instead of 14. `population_yg` now
reaches 2023 too. Neither quantity this branch forward-fills needs to be.

## A rewire is ready

`edu/spain-proc-coefs-rewire` (`6168535`) points `.spain_processing_coefs()`
at the builder. Not a one-line swap — the builder emits codes, so Spain is
selected on `area_code == 203`, `cf` becomes `final_conversion_factor`, item
names come from `add_item_cbs_name()`, and `value_to_process` is carried
directly instead of recovered as `value_proc / cf`. The two unit tests move to
the new schema; full suite green. Take it, rework it or ignore it.

The forward-fill helpers stay — coconuts end in 2002, sugar cane 2008, palm
kernels 2018 — they just stop spanning two whole years for every item.

## One decision it forces

This turns a pin read into a pipeline build: seconds become minutes (2.8 for a
5-year window), and the call passes no `years`, so it builds the full
1850-2023 default. Cached per session, so only the first call pays.

1. Leave it — correct and always current, but `create_n_prov_destiny()` gets
   slow.
2. Scope the years — CBS processing starts in 1961, so
   `get_processing_coefs(years = 1961:last_year)` skips the expensive
   pre-1961 build. Needs `last_year` derived before the reads.
3. Keep a pin, regenerated from the builder — fast, but reintroduces exactly
   the staleness that caused #449.

Option 2 looks right to me, but it reorders your function, so it is your call.
Option 3 I would not merge without scheduled regeneration.

Either way the rewire needs #812 merged first, otherwise
`get_processing_coefs()` still returns the partial 2023.

AliceBeckmann:
Thank you @lbm364dl and @eduaguilera for your reviews!

Here is what I adapted and what still needs to be confirmed:

processing_coefs — done. I decided to apply option 2. So I used your function (calling get_processing_coefs() instead of the old pin). I also added two things: limited it to years 1961 onward, since CBS processing data doesn't go back further anyway, so no point building all the way back to 1850 every time. And I moved the seed subtraction to happen before the processing split, not after — seed should come off the whole harvest, not just what's left once processing is taken out. This matches how it was originally meant to work.

population_yg — not done yet, needs real data, not code. Unlike processing_coefs, there's no function that builds this one fresh — it only comes from a pin, I applied, which still stops at 2021. I couldn't find the updated data in Spain_Hist. @eduaguilera can you please let me know where I can find the new dataset to create a new pin for the population numbers? Thank you.

no_tracked_output in decomposition analysis — Here I need your confirmation @eduaguilera. Some crops (mostly ones turned into sugar, oil, etc.) had their cropland area and N input silently dropped from one of the decomposition analyses, since the processing step renames the crop's output, so it looked like there was no output at all for some processed items. I fixed it so that mass now counts as surplus, since we treat this kind of loss as occurring during processing. This changes the trend a bit over time (barely anything in the 1860s, up to ~5% by the 2020s), so I wanted to flag it and get your OK on this before it's considered done.

eduaguilera:
Thanks, @AliceBeckmann.
About the population pin, I already uploaded it (also the faostat ones) but the PR is pending acceptance by @lbm364dl to be merged to main.
About the no-nitrogen items, ok to be considered as surplus. Most of it is waste. Some of it may be recycled to soil or feed, but we don't know how much and it's very little in any case. We will refine it later on.

lbm364dl:
## Third pass — #812 landed, so the blocker is cleared

Re-verified on a real `create_n_prov_destiny()` build over 1860-2023 after
merging main (which brings whep#812 and whep#772).

**The 2023 gap is genuinely closed.** `get_processing_coefs()` for 2023 now
returns **5,999 coefficient rows globally and 82 for Spain**, against 358 and
1 before — in line with every other year (2021: 5,970/80, 2022: 6,023/82).
`population_yg` covers 1860-2023, so `.forwardfill_population()` is now a real
no-op (8,200 rows in, 8,200 out) rather than quietly synthesising two years.

Both caveats I added while #812 was pending are therefore reversed: the NEWS
entry no longer says 2023 is carried forward from 2022, and the population
forward-fill doc no longer claims the pin stops at 2021.

**Re-measured national N** after the pin refresh: 5,096,664 Mg (2021),
4,607,682 (2022), 4,203,893 (2023). All three years remain structurally
complete — 50 provinces, 8 origins, 7 destinies. The NEWS figures are updated
to these.

### Gates, run locally on the merged tree

| Check | Result |
|---|---|
| `devtools::test()` | **FAIL 0, PASS 9577** (365s) |
| `rcmdcheck` | **0 errors, 0 warnings, 0 notes** |
| `air format --check .` | clean |
| pkgdown reference index | complete |
| `test_data_raw_freshness.R` | passes against the rebuilt `.rda` |

The merge conflicted only on `data/whep_inputs.rda`. That is a build product,
so neither side was picked: it was rebuilt with `Rscript
data-raw/whep_inputs.R` from the merged CSV, which carries main's three
refreshed pins alongside this branch's four.

### What still needs a human, not a green tick

Everything mechanical is done and verified. Two things remain that CI cannot
settle:

1. **`no_tracked_output` is a science decision.** Replacing the `left_join`
   with a `full_join` recovers rows that were being dropped silently — 1.32%
   of cropland area and 1.55% of N inputs over the full span, rising from
   0.05% in the 1860s to about 5% by the 2020s, so it moves the trend and not
   just the level. The recovered rows are almost entirely the oilseed complex
   (sunflower 9.6 Mha cumulative 2010+, rape and mustardseed, cottonseed,
   sugar beet) — the crops whose output the processed-item change reassigns.
   That makes it a quantified instance of whep#432. Whether booking them as a
   zero-output surplus bucket is the accounting we want is @eduaguilera's
   call.

2. **The 2022-23 decline is still uncorroborated.** National N falls 9.6% then
   8.8%. It is in the input pin rather than in this code — synthetic N drops
   29.7% between 2021 and 2022 while cropland area stays flat at 50.56 Mha —
   and it is consistent with the 2022 fertiliser price shock, but it has not
   been checked against an independent source. Worth one before those years
   are published.

### Two findings from earlier passes worth keeping on the record

Both were measured, not inferred, and neither is a defect to fix:

- The **seed-from-Product-row** fix is inert on real data. No Residue or Grass
  row ever matches a seed rate — they carry different `Item` names, so the
  join never fires. Reverting it and re-running the pipeline gave output
  identical to the last decimal. It is a sound guard; it just does not
  currently change anything.
- Adding **Fish and Agro-industry** to the FCI `external_inputs` list is also
  inert. Those are `box` values, never `origin` values; all 5.4 Mt N of that
  biomass already entered as `origin == "Outside"`, which was already in the
  list.

### Closes on merge

whep#430 (Finn transpose), whep#431 (patchwork), whep#434 (16.5 MB xlsx now a
194 KB CSV), whep#435 (33 runnable examples), and whep#449 via the
coefficient-builder switch.

AliceBeckmann:
Thanks for the third pass, @lbm364dl !

On no_tracked_output: Edu said above it's fine as surplus for now, so that one should be fine.

On issue #432 (zero-N processed items): I did make a change for this — I added a separate processing_losses destiny so the leftover N from wine/oil/sugar processing doesn't stay mixed into the primary crop's own destinies anymore, and falls into surplus properly instead. But since this PR is basically ready to merge, I don't want to reopen review with something new at this point. So I'm keeping it out of #416 and opening a separate PR for it.

On the 2022-23 decline: I'll keep it on my to-do list to check it, but I don't think it should hold up merging this PR since it's about the input pin, not this PR's code.

So from my side, this PR is ready to merge as is and I will open a new smaller one after.

lbm364dl (approved):

eduaguilera (approved):

Issue #361, opened 2026-07-24
biomass_coefs: Edible_N_kgFM (and the Edible_* nutrient columns) are entirely empty; edible-portion nutrition never activates


eduaguilera:
## Problem

`whep::biomass_coefs$Edible_N_kgFM` is **all-NA** (0 of 421 rows populated). The same column is empty in the **afsetools source** (`afsetools/inst/extdata/Biomass_coefs.xlsx`, `Coefs` sheet) too, so this is an unpopulated coefficient in the upstream data, not a migration loss.

Consequence: the edible-nutrient coalesce chain in `build_food_supply()` (whep_native) -

```
protein_frac_kgfm = coalesce(Edible_N_kgFM, N_kgN_kgFM, Product_kgN_kgDM * Product_kgDM_kgFM) * 6.25
```

never uses its preferred `Edible_N_kgFM` branch. It always falls back to **total commodity N** (`N_kgN_kgFM`, 142 items) or product N (33 items), i.e. it applies whole-commodity nitrogen to the whole food mass. For high-non-edible-fraction foods (meat with bones, fruits/roots with peel, shell), this over-states edible protein even when the food quantity is correct.

The same emptiness very likely affects the sibling edible-portion columns (`Edible_kgDM_kgFM`, `Edible_kgC_kgFM`, `Edible_kgP_kgFM`, `Edible_K_kgFM`), so any edible-nutrient (energy, C, P, K) calculation silently degrades to whole-commodity values.

## What IS available

The afsetools `Coefs` sheet does carry the ingredients to derive edible N:
- `Edible_portion` (edible fraction of fresh matter),
- `N_kgN_kgFM` (whole-commodity N),
- `NonEdible_kgN_kgFM` (non-edible N), and `Edible_kgDM_kgFM` / `NonEdible_kgDM_kgFM`.

So `Edible_N_kgFM` could be populated as (whole-commodity N minus non-edible N), or from an edible-portion x edible-N-density relationship, rather than left NA.

## Suggested actions

1. Populate `Edible_N_kgFM` (and the other `Edible_*` nutrient columns) in the source `Biomass_coefs` from the edible-portion / non-edible fractions already present, so the coalesce chain's edible branch activates.
2. Until then, `build_food_supply(method = "whep_native")` should be understood to return **whole-commodity** protein (an over-estimate for low-edible-fraction foods), and consumers should prefer FAOSTAT FBS per-capita supply where an edible basis is required.

## Scope

[...]

Issue #359, opened 2026-07-24
N-loss cascade lacks global per-cell drivers (climate/tillage/irrig_cat/som_share/cn_input); Mode-B pathway boundary cannot run gridded


eduaguilera:
## Problem

`build_nitrogen_balance()`'s loss cascade cannot run on a global gridded domain because the per-cell driver columns the loss functions require have **no producers** in the package - they are consumed via caller-supplied `data$n_balance_drivers` / `data$n_balance_leaching_drivers` or default to `NA`:

- `calculate_soil_n2o(method = "ipcc2019" / "ipcc2006")` needs `climate` (`ATL` / `MED`) and `irrig_type`. `.n_check_climate()` aborts on anything but `ATL`/`MED`; there is no global classifier that assigns these per cell.
- `calculate_nh3(method = "manner")` (the default) needs the full MANNER driver set per `(key, fert_type)`.
- `calculate_n_leaching(method = "meisinger_drainage")` needs `climate`, `irrig_cat`, `tillage`, `som_share`, `cn_input`, `land_use` **and** `drainage_mm`.

Drainage IS now available (via `build_water_balance()`, using the 2009 LPJmL proxy - see #340). The **categorical drivers** (`climate` ATL/MED, `tillage`, `irrig_cat`) and the `som_share` / `cn_input` continuous drivers are the missing piece: these were Spain-specific in the port and have no global per-cell producer.

Consequence: a global run either aborts inside `calculate_*()` naming the missing column, or must be fed constant/placeholder drivers (which is not a real Mode-B result). The N **surplus** (Mode A, harvest-removal) does not use the loss cascade and is unaffected; the **pathway boundary** (Mode B: WHEP's own NH3 / NO3 vs medium-specific critical loads) is blocked.

## Suggested actions

1. Add a per-cell `climate` (ATL/MED, or a finer scheme the EF tables support) classifier - e.g. a Koppen/aridity rule from CRU temperature+precipitation, which covers 2010 natively.
2. Provide global `tillage` and `irrig_cat` per-cell layers (or documented defaults) for the leaching driver join.
3. Derive `som_share` and `cn_input` from the SOM/carbon balance and the N-input mix once those are wired (see the SOM issue).
4. Until then, document that Mode-B requires these drivers and is not runnable globally with the current package.

## Scope

Not part of the current nitrogen-balance (Mode A surplus) work; that ships and is validated. This blocks the Mode-B pathway boundary the SJOS-N analysis wants (water + air).

[...]

Issue #278, opened 2026-07-23
Urban N input uses Spain historical night-soil per-capita rate as a global 2010 constant (~7x realized flux)


eduaguilera:
## Problem

The urban/human-excreta N input term (`.n_inputs_urban()` -> `build_urban_n()`, `R/n_urban.R`) applies a per-capita rate of ~1.263 kg N/cap for 2010, producing **4.41 Tg N** globally (verified: mass-conserving generator over the 2010 HYDE urban population of 3.495e9, 99.9% polity-matched).

The realized global flux of urban/human-excreta N actually recycled to agricultural land in 2010 is only ~0.4-1 Tg N (Moree et al. 2013 GBC; realized recycling ~2% of ~20-25 Tg total human-excreta N). So the term overstates the modern global flux by roughly an order of magnitude.

## Root cause

The per-capita rate is Spain's **historical night-soil-recycling** series (documented in `R/n_urban.R` as a placeholder). Spain historically returned a large share of urban excreta to farmland; globally by 2010 almost none is. A second, smaller inflation: the mass-conserving transport counts the "residual" urban N that finds no cropland room within the search ring as an input at the source cell, so 4.41 Tg is closer to "all urban N generated in agricultural polities" than "N placed on cropland."

## Fix direction

Replace the Spain-historical constant with a time- and region-varying realized-recycling rate (or a documented modern global fraction), and decide whether unplaced residual should count as an agricultural input. Belongs to the urban-N calibration, not the current N-balance PR.

## Scope

Not blocking: the term produces the correct `fert_type = "urban"` with the correct schema; the engine is sound. Absolute magnitude is small (~4 Tg of a ~260-430 Tg input total), so it barely moves the surplus. Interim SJOS-N runs use it as-is and flag it.

eduaguilera:
Scientific work claimed by @eduaguilera under `plans/2026-08-08-sjos-n-scientific-resolution.md` (PR unit S5 / decision D50; realized urban-N recycling). Please coordinate here before starting overlapping scientific or methodological work. Cata's mechanical scope remains in its existing workstream and is not claimed.

lbm364dl:
## Triage (sweep, 2026-08-27): STILL VALID (verified on `origin/main`)

`R/n_urban.R` still carries the Spain-historical placeholder, and says so in its own
header:

```
# The per-capita urban-N-to-agriculture rate is a documented placeholder:
# Spain's own historical benchmark series (urban_n_reference, ...)
```

[...]

Issue #277, opened 2026-07-23
Livestock intake / manure N ~2x inflated (feed-demand exceeds realistic; applied > excreted)


eduaguilera:
## Problem

WHEP's manure N applied to agricultural land for 2010 comes out at **213.3 Tg N** (cropland solid+liquid 63.5 + grassland-deposited excreta 149.8), assembled through `.n_inputs_manure()` -> `build_livestock_nutrient_flows()` from `redistribute_feed()` intake.

The decisive tell: WHEP's **applied** manure N (213 Tg) **exceeds the entire world's manure N excretion** per the literature (~123-131 Tg N: FAOSTAT 2016 = 123; Zhang et al. 2017 ESSD = 131 in 2014). You cannot apply more manure N to land than livestock excrete, so the pipeline is inflated, not merely mis-scoped.

Component-wise: cropland-applied 63.5 Tg vs literature ~25-28 Tg (~2.3x); grassland-deposited excreta 149.8 Tg vs literature ~80-85 Tg (~1.8x).

## Root cause (upstream, not the manure term)

The manure engine (`build_livestock_nutrient_flows()`) is mass-conserving and faithful. The inflation is upstream in **feed demand / intake**: ~17 Gt DM intake vs ~6 Gt realistic, driven by unbounded grass because the intake step supplied no grass-availability cap (`redistribute_feed(..., options = list(distribute_surplus = FALSE))` with no availability ceiling). Excretion then inflates to ~263 Tg (vs ~130 literature), and applied follows.

## Fix direction

Belongs to the intake stage (`redistribute_feed()` / feed-demand assembly), not the N-balance wiring. Cap grass availability so intake matches realistic totals. Tracked separately from the current N-balance PR.

## Scope

Not blocking: the manure term produces the three correct `fert_type`s (`manure_solid`, `manure_liquid`, `excreta`) with the correct schema; only the magnitude is wrong. Interim SJOS-N runs use it as-is and flag it.

eduaguilera:
This is still live, and there is now a repo-shipped check that quantifies it — with a detail that distinguishes it from #440.

Running `validation/n_balance_invariants.R` on the current pins (literal output, exit code 1):

```
→ PASS  sum(N input streams) == N_input_std       ℹ max relative gap 8.57e-15
→ FAIL  manure N <= livestock excretion N, every year
        ℹ worst year 1860 at ratio 1.65 (105 of 164 years exceed 1)
→ PASS  no year has zero total Excreta
[...]

PR #422, opened 2026-07-30
feat(sjos-n): Safe and Just Operating Space for nitrogen

eduaguilera:
### What this adds

The Safe and Just Operating Space for nitrogen (SJOS-N), built on WHEP's own gridded nitrogen balance rather than on external national totals. Fifteen new exported functions, chained end to end by `build_sjos_nitrogen()`:

| Stage | Functions |
| --- | --- |
| Critical-N archive | `read_critical_n()` (Schulte-Uebbing gridded layers; `dir` -> `WHEP_CRITICAL_N_DIR` -> local cache, fetched from Zenodo on demand) |
| Agricultural land support | `build_ag_land_support()` (physical cropland + grassland hectares per cell/polity/item) |
| Surplus | `calculate_n_surplus()` (`harvest_removal` default, `full_balance` alternative) |
| Boundary, surplus mode | `build_n_boundary_exceedance()` (grid / polity / country / IMAGE region) |
| Boundary, pathway mode | `build_n_pathway_exceedance()` (per-medium critical loads, binding boundary) |
| Nourishment axis | `build_food_supply()`, `normalize_nourishment()`, `calculate_food_gini()`, `disaggregate_ussr()` |
| Per-capita axis | `build_n_percapita()`, `build_n_boundary_percapita()` |
| Classification | `classify_sjos_n()` |
| Embodied-N trade | `build_n_exceedance_extension()`, `build_sjos_n_footprint()` |
| Driver | `build_sjos_nitrogen()` |

Package data: `coello_synthetic_n` (crop-specific synthetic-N rates, Coello et al. 2025), `n_boundary_params`, `nourishment_thresholds`, `sjos_levels`, `nourish_levels`.

Every exported function runs from injected fixtures under `example = TRUE`, so the whole chain is exercised without the heavy gridded inputs.

### Behaviour changes to already-merged functions

These change outputs that consumers of `main` already read, so they deserve the closest review:

1. `build_n_inputs()` / `build_nitrogen_balance()` gain a `method_synthetic` provenance column, and the synthetic term is routed through the crop-specific Coello rates by default, with the previous crop-share method still selectable.
2. Atmospheric deposition is multiplied by agricultural hectares (cropland and grassland support) instead of whole-cell area. On the real 2010 grid this takes deposition from 63.4 Tg N/yr whole-territory to 27.9 Tg N/yr on agricultural land (12.2 cropland + 15.7 grassland), a 44% share. See the open item below: the intended design is a land-use-resolved accounting covering every cell fraction, filtered per analysis, so this step is an interim narrowing rather than the final contract.
3. Nitrogen inputs that are not inherently crop-specific are allocated over observed agricultural support, area-weighted and mass-reconciled: the allocation aborts if source and allocated mass differ.
4. `build_nitrogen_balance()` emits per-crop `area_ha`, needed by the per-hectare boundary comparison.
5. `build_crop_soil_n2o_extension()` shares the Coello crop split.
6. The agricultural land support the non-crop-specific terms allocate over is now **derived natively** by `build_ag_land_support()` from the `cell_polity` / `type_cropland` / `crop_patterns` / `gridded_pasture` surfaces the gridded pipeline already reads. It is materialised once and shared by `build_n_inputs()` and `build_nitrogen_balance()`, so the deposition term, the unattributed-input allocation and the per-hectare `area_ha` denominator cannot disagree. A caller-supplied `data$ag_land_support` still overrides it.
7. `read_critical_n()` no longer requires a hand-downloaded archive. It falls back to a `rappdirs` cache and, when that is empty, downloads the 18.4 MB CC-BY-4.0 Zenodo archive (record 6395016), verifies it against the published MD5 and unpacks it. Extraction tries the `archive` package (Suggests-only) then a `7z` binary; with neither it aborts *after* a verified download, naming the one command left to run. Same download-on-demand pattern `.provinces_shapefile()` already uses.
8. The manure engine's own `data$resolution` is no longer overwritten. It now defaults to `"subnational"` at `resolution = "grid"` (cell-level nitrogen needs cell-level manure) and `"national"` otherwise, and an explicitly supplied value is always honoured.

### Methodological decisions taken

Cropland support is the **physical** LUH2 cropland area apportioned among crops by a normalised `crop_patterns` composition, never the raw `harvest_fraction` (a harvested basis that sums above 1 under multicropping and would over-credit every per-hectare term). Verified on the real grid: support equals physical cropland x `polity_frac` to a relative difference of 1.6e-16. Grassland support defaults to the `gridded_pasture` surface, which agrees with `read_luh2_landuse()` to the hectare (3208.5 Mha for 2010) but spans 1851-2023 rather than stopping at 2015; `grassland = "luh2" | "none"` stay selectable. No invented historical intensive/extensive grassland split. CBS 3000 is kept for all grassland; only nitrogen over observed agricultural support is allocated to crops; genuinely unattributed inputs stay explicit and mass-reconciled rather than being silently assigned to crops; `land_use = "ara"` is the primary historical boundary comparison, with the source archive's intensive-grass mask reserved for source benchmarking.

### Verification

- `air format .`: clean.
- `lintr::lint_package()` with the CI linter set: 0 lints.
- `devtools::test()`: 4907 passed, 0 failed, 0 errors, 11 skipped.
- Real-data smoke test (not just fixtures): `build_ag_land_support()` over 1961/1990/2010/2015/2020/2023 gives 6.9M rows, no NA, no non-positive area, unique on the support key, grassland always CBS 3000, cropland 1293 -> 1547 Mha and total agricultural area 4267 -> 4700 Mha, both tracking FAO. `.n_inputs_deposition()` runs end to end for 2010 (1,147,291 rows, no NA). 8251 cell-years (12.1 Mha, 0.8% of cropland) have cropland but no crop-pattern composition and are named by a warning rather than dropped silently.
- `read_critical_n()` verified against the real Zenodo archive for the first time: all seven layers read correctly (including through the archive's own misspelled `Exeedance of critical N surpluses` directory), `critical_n_surplus` giving 28,573 cells over [-316.6, 496.9] kg N/ha/yr and `binding_threshold` the documented categorical 1-8. The download-and-MD5 path ran for real; the two extraction back-ends could not be exercised locally (no libarchive, no `7z`), so their dispatch is covered by injected fakes rather than end to end. Confirmed with the cache hidden that the test suite downloads nothing.
- Merge check against current `main` (including #398's cell-polity crosswalk repair): no textual conflict, 4978 tests pass, and the real-data support totals are unchanged.
- `rcmdcheck()`: 0 errors, 0 warnings, 2 NOTEs. Both NOTEs are inherited from `main` and untouched here (unused `readxl` import; `.sci_sum_components` NSE variables in `soil_carbon_inputs.R`).
- pkgdown: every `man/*.Rd` topic appears in `_pkgdown.yml`.

### Known limitations (open issues, not addressed here)

- Whole-territory accounting. WHEP quantifies nitrogen over agricultural support only; this branch narrows deposition to agricultural hectares, which removes the whole-cell over-count but drops the non-agricultural share instead of accounting for it (56% of deposition in 2010). The agreed design is that every cell fraction carries a quantified land-use class (including natural and urban) so the whole territory is accounted, with each analysis filtering afterwards, driven by WHEP's own land-use estimation. Tracked in #423, which generalizes #276. (#276 itself, the ~5x full-cell-area over-count, is FIXED here: measured 63.4 Tg N/yr whole-territory against 27.9 Tg on agricultural land, matching that issue's own predicted cropland figure of 12.33 Tg to within 1%.)
- #277 livestock intake / manure N is roughly 2x inflated upstream. It largely self-cancels in the surplus, since the inflated herd inflates both the excreta input and the grazed-forage removal.
- #278 urban N still uses a Spain historical night-soil per-capita rate.
- #359 the Mode-B pathway boundary has no global per-cell loss drivers, so `build_n_pathway_exceedance()` runs from supplied critical loads and a supplied balance, not from a turnkey global cascade.
- #361 `biomass_coefs$Edible_N_kgFM` is empty, so the nourishment axis falls back to whole-commodity nitrogen.
- #446 the chain from `build_nitrogen_balance()` through the surplus, classification and footprint has never been run on real inputs: there is no driver assembling the real `data` list, so every test and example on that path is fixture-driven. The two data entry points either side of it (the gridded land support / deposition term, and `read_critical_n()` plus the boundary arithmetic) *are* now verified against real inputs.
- #450 the nourishment ("just") axis has never run on real data either: `cbs_food` works, but the `gdp-population` pin is ISO3-keyed while `build_food_supply()` wants numeric codes. Smaller than #446, and cross-checkable against the `"faostat_fbs"` method.
- #451 the two 7z extraction back-ends behind the new Zenodo fetch could not be exercised where this was written (no libarchive, no `7z` binary); their dispatch is covered by injected fakes, the calls themselves are not.

### Suggested reading order

`R/n_surplus.R` → `R/critical_n.R` → `R/n_boundary_exceedance.R` → `R/n_pathway_boundary.R` → `R/food_supply.R` + `R/nourishment.R` → `R/sjos_n_class.R` → `R/n_exceedance_extension.R` + `R/sjos_n_footprint.R` → `R/build_sjos_nitrogen.R`. `R/ag_land_support.R` is the producer for the land support those terms allocate over. The changes to `R/n_balance_inputs.R`, `R/n_balance.R` and `R/n_balance_spatialize.R` are the ones that alter existing behaviour.


lbm364dl (approved):
Approving from our side. Two things to declare up front, then one decision that is yours.

## Scope of this approval

**I am not an independent reviewer of all of it.** Six of the commits on this branch are mine (`7701ebb2`..`d8363333`), so roughly a third of the diff is not something I can review — only describe.

**And I did not read the new science.** The original ask was "is this merge-ready", so I went at the risk to existing `main` consumers: the five behaviour changes to already-merged functions, plus CI. That is where the blocker was. I never opened these, ~1,600 lines:

`n_pathway_boundary.R` (372) · `food_supply.R` (196) · `sjos_n_footprint.R` (175) · `n_boundary_percapita.R` (169) · `n_percapita.R` (163) · `food_gini.R` (162) · `n_surplus.R` (127) · `n_exceedance_extension.R` (122) · `nourishment.R` (113)

Nobody but you has read those. If you want a real second pair of eyes on the methodology, that is a separate review and worth asking someone who knows the nitrogen work.

**Why approve anyway:** the new modules are additive — nothing on `main` calls them, so a latent bug there cannot break anything today. The part that could hurt current users is the part that was checked, and fixed.

## What is now verified against real data

| | |
| --- | --- |
| Cropland support vs physical LUH2 x `polity_frac` | exact, rel. diff 1.6e-16 |
| Agricultural N deposition (2010) | 27.9 Tg N/yr of 63.4 Tg whole-territory |
| `read_critical_n()` | all seven layers, incl. the archive's misspelled `Exeedance...` dir |
| Boundary arithmetic vs the archive's own exceedance layer | exact on all 26,497 positive-critical cells |

That last one is a known-answer test rather than a plausibility check: the archive publishes both the critical surplus and its own exceedance, so `actual = critical + exceedance` is recoverable and can be fed back through `build_n_boundary_exceedance()`.

It also confirms #276's predicted numbers by a completely independent route — that issue expected 12.33 Tg cropland-masked over 1,489 Mha; measured 12.2 Tg over 1,446 Mha. I have suggested closing it there.

## Your call: unmatched cropland

#447 stated that *"unmatched physical cropland must remain present with an `NA` item code"*. This branch does the opposite — `build_ag_land_support()` drops it with a warning — and I did not change that, because it is a methodology decision, not a defect.

Measured on the real 2010 grid:

- **8,251 cells, 12.12 Mha** (0.8% of 1,489 Mha) have LUH2 cropland but no EarthStat crop composition.
- **100% of them are cells where EarthStat reports nothing at all.** Zero are lost to a crosswalk gap — WHEP maps 264 codes against EarthStat's 144. So this land is *permanently* unattributable, not recoverable with a better mapping.
- Cost: roughly **0.1 Tg N/yr** of deposition, about **0.4%** of the agricultural total.
- Likely worse away from 2000, since EarthStat is a single static snapshot stretched across 1851-2023.

Arguments both ways, briefly. Keeping it as `NA` conserves the land base and the mass, which is what a *balance* wants and what #423 is heading toward. Dropping it keeps every tonne attached to a traceable product, which is what a *footprint* needs — an `NA`-crop tonne cannot be traced through the IO model at all.

Worth knowing before you decide: this branch **deliberately eliminated `NA` item codes** (see the `R/n_balance_inputs.R` header) and `.nb_validate_input_grain()` now aborts on them at grid resolution. So reinstating them is not a local change — it means relaxing that guard and deciding what unattributed nitrogen does downstream: does it enter the boundary verdict, the classification, a separate unallocated line? Happy to implement whichever way you want it.

## Gates before any SJOS-N number is published

Merging the machinery is safe. Trusting its output is not yet:

- **#446** — `build_nitrogen_balance()` -> surplus -> classification -> footprint has never run on real inputs; no driver exists. Same shape as the bug this branch had to fix: a green suite over a path nobody had run.
- **#450** — the nourishment axis likewise. Sharper than it sounds, since #429 fixed `get_wide_cbs()` returning row counts instead of tonnes only days ago, and that is exactly this axis's input.

Also opened #451 (the two 7z extraction back-ends behind the new Zenodo fetch could not be exercised where they were written) and #452 (manure N is split by harvested area while synthetic N now uses crop-specific rates, two lines apart in `build_crop_soil_n2o_extension()`).

PR #448, opened 2026-08-03
fix(bnf): correct faba-bean N harvest index misread as a sample size

eduaguilera:
## What

`inst/extdata/coefs/bnf.csv` stored `Fava bean` `n_harvest_index = 0.34`. Anglade et al. (2015) *Ecosphere* 6(3):37 Table 1 gives faba bean **NHI median 74%, n = 34** — the sample size was transcribed into the index column.

Table 1 is six side-by-side `(n, median)` blocks. Only this one cell took the first element of its pair.

## Why we are confident, before changing a shared coefficient

Three independent checks, each reproducible from files in the repo:

| check | result |
|---|---|
| Column-sum identity | the NHI `n` column sums 34 + 55 + 100 = **189**, exactly the table's own "All" row — this pins which element of each pair is the count |
| Sibling rows | Lentils `55 75a` → 0.75 and Pea `100 75a` → 0.75 are stored correctly, and every stored `ndfa` is a median and never an `n`, so exactly one cell is wrong |
| Independent route | `bio_coefs` grain/residue N, which never reads this table, implies faba-bean NHI **0.785** |

## Impact

`R/bnf.R:646-654` divides by `n_harvest_index`, so the stored value inflated faba-bean BNF by `0.74 / 0.34 = 2.176x` wherever `crop_bnf_anglade_t` was produced.

**No result changes.** `crop_bnf_anglade_t` and `bnf_product_ratio_anglade` are computed but excluded from the total at `R/bnf.R:402`, which sums `crop_bnf_t + weed_bnf_t + nonsymbiotic_bnf_t`. A characterization run over the full coefficient table moves exactly **1 of 31 rows**, leaving `crop_bnf_t`, `bnf_t`, `bnf_product_ratio_npp` and `ndfa_adj` bit-identical. A downstream sweep found nothing that consumes the Anglade route in any output, figure or deposit.

## The test, and why the existing one could not catch this

`tests/testthat/test_bnf.R` reads a coefficient from the table and checks the function reproduces it, so it passes at 0.34 or 0.74 alike. That is the shape that let this survive.

The new `tests/testthat/test_bnf_coef_sources.R` asserts stored coefficients **against their cited sources**. It transcribes Anglade Table 1 and Herridge Tables 2 and 6 literally, then proves its own transcription two ways before asserting anything — the `n` columns must sum to the "All" row, and every median must lie inside its own IQR. Those two checks are exactly what a sample-size-in-an-index-column destroys. It also guards the mechanism directly: no Anglade-cited fraction may equal a sample size from its own row.

Coverage is declared, not implied: of 60 non-missing numeric cells, 18 are source-asserted, 3 derivation-asserted, and **39 explicitly declared unasserted with a stated reason**. A partition test fails if a cell is added or moved between ledgers.

Teeth demonstrated by re-introducing 0.34: **three independent tests fail** (the Table-1 median, the sample-size coincidence guard, and the `bio_coefs` cross-check, which never touches this file).

## Worth a reviewer's opinion

1. **0.74 vs 0.75.** Table 1's median is 74. But the Discussion says *"high aboveground NHI values (0.75 for grain crops and 0.9 for alfalfa) were assumed"*, and the registry already stores 0.75 for Lentils, Pea, Beans and Other pulses — whose Table-1 medians happen to be 75 either way, so the existing convention is ambiguous between "species median" and "the paper's assumed grain value". This PR uses **0.74**, the literal table reading. The difference is 1.35% against a 118% error corrected, so nothing material rides on it, but the convention is worth settling.

2. **Three declarations record coefficients their cited source does not support**, left unchanged here and flagged for review: `Beans`/`Groundnuts` `n_harvest_index` and `below_ground_n_ratio` against Herridge's harvest indices (0.35, 0.4) and 1.4 multiplier; `Soyabeans` `ndfa` 0.57 against Salvagiotti's 52/58. None shows the sample-size signature — the mechanism guard covers all ten Anglade-cited fractions exhaustively and they pass — so there is no second instance of *this* defect, but these are separate citation mismatches.

3. **Stale copies elsewhere** will re-introduce 0.34 on merge: `WHEP-ratio-decomposition` @ `51c01711`, and the installed/renv copies until reinstalled. `Spain_Hist/output/Org_calibration/BNF.csv` also carries a committed stale mirror (its NHI is not read).

The same one-line correction and an equivalent test are on `afsetools` main as `fd568ce`.

### DIFF
--- a/inst/extdata/coefs/bnf.csv
+++ b/inst/extdata/coefs/bnf.csv
@@ -2,7 +2,7 @@ name_bnf,ndfa,n_harvest_index,below_ground_n_ratio,nonsymbiotic_base_kg_ha,legum
 Alfalfa,0.7,0.9,1.7,,1,"Anglade et al., 2015"
 Beans,0.4,0.75,1.3,,1,"Herridge et al., 2008"
 Clover,0.86,0.9,1.7,,1,"Anglade et al., 2015"
-Fava bean,0.75,0.34,1.3,,1,"Anglade et al., 2015"
+Fava bean,0.75,0.74,1.3,,1,"Anglade et al., 2015"
 Green leguminous,0.68,0.5,1.3,,1,"Lassaletta et al., 2014, based on Anglade et al., 2015"
 "Groundnuts, with shell",0.68,0.5,1.3,,1,"Herridge et al., 2008"
 Lentils,0.66,0.75,1.3,,1,"Anglade et al., 2015"
--- a/tests/testthat/test_bnf_coef_sources.R
+++ b/tests/testthat/test_bnf_coef_sources.R
@@ -15,6 +15,12 @@
 #   source      asserted against an external cited publication
 #   derivation  asserted against the registry's own declared derivation rule
 #   unasserted  declared here, with the reason it cannot be checked
+#
+# The ledgers are `tribble`s rather than parallel vectors so that each
+# assertion is one readable row. A value and its citation split across two
+# position-aligned vectors can slip against each other silently, since the
+# citation is only ever read back in a failure message -- the same class of
+# misalignment this file exists to catch.
 
 # ---- Anglade et al. (2015), Ecosphere 6(3):37 -------------------------------
 # Anglade, J., G. Billen, and J. Garnier. 2015. Relationships for estimating
@@ -31,7 +37,7 @@
 # distinction a sample size transcribed into an index column destroys.
 # Alfalfa and clover are forage legumes and carry no NHI.
 .anglade_2015_table1 <- function() {
-  data.frame(
+  tibble::tibble(
     species = c("Alfalfa", "Clover", "Faba bean", "Lentil", "Pea", "All"),
     shoot_n_pct_n = c(101, 327, 40, 48, 55, 571),
     shoot_n_pct_median = c(2.9, 3.8, 2.5, 2.2, 2.7, 3.2),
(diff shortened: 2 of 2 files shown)


lbm364dl:
## Verified against the sources, and one convention call for you

I checked the transcription against the publications rather than against the PR body. Table 1 is a JPEG in the Wiley OA HTML, which is why it resists text extraction — pulled the image and read it directly.

**The fix is right.** Table 1, faba bean NHI block: `n = 34`, median `74a`, IQR 63–77. The stored `0.34` was the sample size.

**The transcription is exact.** All 138 non-NA cells of `.anglade_2015_table1()` match the published table — not one slip. Both Discussion quotes are verbatim. Herridge Table 2 (common bean 40, soybean/groundnut 68, *experiments* column — the farmers' fields column next to it reads 36 and 58) and Table 6 (rice 33, sugar cane 25) also check out. Herridge Tables 3–4 confirm the reason string too: 18.1/51.7 = 0.35 for common bean, 37.6/93.9 = 0.40 groundnut, and 1.45/1.03 = 1.41 for the below-ground multiplier.

Also worth recording: **"no result changes" is provable at code level**, which is stronger than the characterization run. `n_harvest_index` appears in `R/bnf.R` only at `:299` (a `select`) and `:647`/`:652`, so only `crop_bnf_anglade_t` and `bnf_product_ratio_anglade` *can* move, and `bnf_t` at `:402` excludes both.

I pushed a follow-up commit for `air format` (`format-suggest` was failing) which also moves the ledgers to `tribble`s — see the commit message for why that pairing wasn't arbitrary. No assertion changed; reintroducing 0.34 still fails the same three tests.

---

### 0.74 vs 0.75 — your call, and the ledger already has an opinion

Worth settling, because the ledger currently applies **two different sourcing rules** and faba bean is the only row where they disagree:

- **Alfalfa**'s NHI (0.9) comes from the *Discussion assumption*.
- **Fava bean**'s NHI comes from the *Table-1 species median*.

Lentils and Pea are 75 under either rule, so they can't disambiguate. Faba bean is the single row that distinguishes them — and it's the row being changed.

**Case for 0.75** (i.e. the Discussion-assumption rule, applied consistently):

1. Table 1's own post-hoc letters read faba **74a**, lentil **75a**, pea **75a** — same letter, **not significantly different**. And they ran that test on NHI deliberately: *"Since the homogeneity of variance was respected for NHI, we performed an ANOVA and a Tukey test for a multiple comparison of means between grain legumes (lentil, faba bean and pea)."* The paper is telling us 74 and 75 are the same number.
2. The Table-1 "All" NHI median is **75**.
3. The paper applies **0.75 for grain crops** in its own N-balance calculation — so 0.75 is what Anglade *uses*, not just what the table reports.
4. All six grain-legume rows become uniformly 0.75 — one rule, no exceptions.
5. The `bio_coefs` cross-check is marginally better at 0.75 (0.785/0.75 = 1.047) than at 0.74 (1.061). Noise, but it doesn't argue against.

**Case for 0.74:** it is the literal species-specific median, and species-specific beats a group assumption where the species is actually measured (n = 34).

The 1.35% difference is immaterial against the 118% error being corrected — this is purely about which rule the registry follows. Note the Table-1-median rule **cannot source alfalfa at all** (alfalfa and clover have no NHI row), so under that rule 0.9 stays a permanent exception; the Discussion-assumption rule covers every row.

Either way it's a one-line CSV change plus one ledger row switching between `.anglade_nhi("Faba bean")` and the `.anglade_2015_text` constant. Happy to make it 0.75 if you prefer — say the word and I'll push it.

### Two things worth separate issues

1. **Your second "Worth a reviewer's opinion" point is structural, not clerical.** `bnf.csv` has one `source` per *row*, but a row's coefficients come from different places: `Beans` cites Herridge (correct for `ndfa` = 0.40) while its `n_harvest_index` = 0.75 is plainly Anglade's grain-crop assumption. Those aren't mis-cited wrong values — they're reasonable values the schema **cannot cite properly**. Per-column provenance (`source_ndfa`, `source_n_harvest_index`, …) would retire the whole class and let the 39 unasserted declarations move out of the test and into the data.

   ⚠️ For whoever picks that up: Herridge's 0.35/0.40 is a **dry-matter** HI while `n_harvest_index` is an **N** ratio — grain is N-richer than straw, so NHI > HI_dm always. Different quantities, not competing values. I made that explicit in the reason string so nobody "fixes" 0.75 → 0.35.

2. `bnf.csv` spells the soybean source **"Salvagliotti"**. Correct is **Salvagiotti** (F. Salvagiotti et al. 2008, *Field Crops Research* 108:1–13, [doi:10.1016/j.fcr.2008.03.001](https://doi.org/10.1016/j.fcr.2008.03.001), verified via Crossref). Your new test file already spells it right, so the CSV is the wrong one. Out of scope here, one character.

lbm364dl (approved):
Approving. All 9 checks green.

Verified the coefficient against the publication rather than the PR description: Anglade et al. (2015) Table 1 gives faba bean NHI `n = 34`, median `74a`, IQR 63–77 — the stored 0.34 was the sample size, and the correction is right. All 138 non-NA cells of the transcribed table match the published table exactly, both Discussion quotes are verbatim, and the Herridge Table 2 and 6 values check out (including that the %Ndfa figures come from the experiments column, not the farmers' fields column beside it).

"No result changes" also holds structurally, not just empirically: `n_harvest_index` appears in `R/bnf.R` only at `:299`, `:647` and `:652`, so only `crop_bnf_anglade_t` and `bnf_product_ratio_anglade` can move, and `bnf_t` at `:402` excludes both.

The test is the good kind — it would have caught this, and reintroducing 0.34 does fail three independent tests, including the `bio_coefs` cross-check that never reads this file.

I pushed one follow-up commit: `air format` (the `format-suggest` failure) plus the ledgers moved to `tribble`s, which is what lets the row-per-assertion layout survive formatting and removes the positional value/citation pairing. No assertion changed.

The 0.74 vs 0.75 convention question is left to you — see my comment above; I would not block on it either way. The oldrel-1 red earlier was an unreachable `saco.csic.es` pin fetch in `test_n_prov_destiny.R`, unrelated to this PR; it passed on rerun.

Issue #424, opened 2026-07-31
22 of 32 area-keyed exported outputs carry no polity columns — extend, or document as intentional?


lbm364dl:
## Census: 10 of 32 area-keyed exported outputs carry polity columns; 22 do not

Completing the coverage question this integration has answered for datasets and for the eight primary builders, but not for exported functions generally.

**Method.** All 172 exported functions; 64 take an `example = TRUE` argument, so they can be run cheaply and their output inspected. Of those 64, 32 return a frame with an area-ish column (`area_code`, `area`, `sub_territory`), 3 error or return a non-frame, and the rest are not area-keyed.

**Result.** 10 carry polity columns. **22 do not:**

```
build_carbon_balance               build_land_balance_footprint    build_urban_n
build_carbon_inputs                build_livestock_ghg_extension   build_water_balance
build_crop_soil_n2o_extension      build_n_deposition              get_arable_permanent_land
build_energy_co2_extension         build_n_inputs                  get_crop_land_extension
build_feed_demand                  build_nitrogen_balance          get_faostat_data
build_feed_intake_local            build_soil_carbon_inputs        get_soc_climate_drivers
build_grass_natural_carbon_inputs  build_grassland_land_extension  read_luh2_landuse
build_grazing_feed_footprint
```

## This is not a correctness gap, and that is worth establishing before anyone "fixes" it

Their `area_code` is the **polity-derived** key, inherited from the builders upstream — `get_primary_production()` emits `polity_area_code` *as* its `area_code`. Checked rather than assumed, on the example path:

| function | codes | resolve to a polity bucket |
|---|---|---|
| `build_nitrogen_balance` | 1 | 1 |
| `build_carbon_balance` | 1 | 1 |
| `build_n_inputs` | 1 | 1 |
| `get_arable_permanent_land` | 1 | 1 |
| `build_water_balance` | 4 | 4 |

So the rows *are* attributable to polities; they are not self-describing. `get_faostat_data` is the one that should stay as it is — it returns `area` as a raw FAOSTAT **name** (`"Portugal"`), because it is the pre-resolution reader and resolution is the caller's job.

## The decision

Should the 22 carry `reporting_polity_code` and siblings, as the eight primary builders do?

[...]

PR #473, opened 2026-08-04
fix(luh2): retire the states pin for a verified Zenodo download (#457)

### LINKED ISSUE #457: luh2_v2h_states pin payload is corrupt, forcing an unrecorded local-disk LUH2 vintage
`luh2_v2h_states` is registered in `inst/extdata/whep_inputs.csv` (row 9) against the SACO pins manifest, so LUH2 states are meant to be served from SACO like every other registered input. **The pin payload is not readable.**

## What is wrong with the payload

Documented in the header of `R/luh2_landuse.R`, verified 2026-06-30:

> the registered pin `luh2_v2h_states` currently ships a serialized HDF5/NetCDF blob, not a tidy table: its `.parquet` member is a single STRING column holding the chunked HDF byte stream, and its `.csv` member is the HDF5 file itself (with a CRLF-mangled signature). Neither reads as a tabular LUH2 grid through the tabular readers.

`.luh2_read_states()` therefore errors, and `.luh2_read_states_source()` only reaches it as a fallback:

```r
# Choose the states source: the clean LOCAL LUH2 v2h states.nc when present
# (the registered pin payload is corrupted, see file header), otherwise fall
# back to the pin reader. The directory comes from the WHEP_LUH2_DIR env var.
```

## Why it matters beyond the read failing

The local-disk workaround has no way to record *which* LUH2 product it read, and two different v2h trees are currently sitting on the same machine:

| Path | `source_id` | Time steps |
|---|---|---|
| `C:\XL_files\LUH2\LUH2 v2h` | `UofMD-landState-LUH2-GCB2022` | 1,173 (850–2022) |
| `C:\XL_files\whep_lpjml\input\LUH2\LUH2 v2h` | `LUH2 v2h` | 1,166 (850–2015) |

They are different products with different time coverage, and `read_luh2_landuse()`'s roxygen `@source` cites only the base v2h release (Hurtt et al. 2020) while `WHEP_LUH2_DIR` may point at either. Preflight diagnostics for the polycell spatial-support work reproduce one set of residual statistics on the GCB2022 tree and a different set on the base tree, with nothing on disk recording which was used.

Restoring a clean pin would make `WHEP_LUH2_DIR` a genuine fallback rather than the primary path, and would remove the ambiguity at its source rather than asking every consumer to track it.

## Suggested fix

Re-upload the pin as either a clean NetCDF or a tidy parquet with `lon`, `lat`, `year`, `state`, `fraction` — `.luh2_read_states()` is already written to decode both, so no reader change should be needed once the payload is right. Name the vintage explicitly in the pin so consumers can record it.

Note the raw `states.nc` is ~6 GB, which is why `CLAUDE.md` keeps multi-GB rasters on local disk; a pin of the *aggregated* 0.5 degree land-use fractions would be far smaller and is what `read_luh2_landuse()` actually produces.

Related: #456 (the tracked `.Renviron` hides `WHEP_LUH2_DIR` entirely, so the fallback path is what most sessions hit first).

/cc @lbm364dl — flagging for Catalin Covaci, since this is SACO/pins infrastructure shared across pipelines.

lbm364dl:
Closes #457.

Two commits: the first fixes the reported read failure; the second removes the pin that caused it to exist at all, after review established the pin was redundant.

## What #457 actually was

The payload was never corrupt. Per-version member lists from the board:

| version | members | what it is |
|---|---|---|
| `20251006T152247Z-942cc` | `.csv` 1.56 GB + `.parquet` 1.53 GB | the serialized-HDF5 blob the issue describes |
| `20260701T083449Z-582d8` (registered) | `states.nc` 6.66 GB | valid NetCDF, LUH2-GCB2022, 850–2022 |

The bad payload was replaced on 2026-07-01 — one day *after* the header note ("verified 2026-06-30") that declared it unreadable. The real bug was `.luh2_read_states()` asking `whep_read_file()` for the default `type = "parquet"`, so `.read_file()` aborted with *"this input has no parquet file"*, which the wrapper relabelled as an unreadable pin.

## Why the pin is gone

The pin was a byte-identical mirror of an openly published, DOI-stable file. Three MD5s agree:

```
Zenodo states4.nc (record 15556812)   411ef3d657c3108942954c895f658a17
retired pin payload                   411ef3d657c3108942954c895f658a17
local WHEP_LUH2_DIR tree              411ef3d657c3108942954c895f658a17
```

So the pin was a second, unversioned, unchecksummed copy of `doi:10.5281/zenodo.15556812` (LUH2-GCB2022, CC-BY-4.0). It was registered as a trailing bullet in commit `d4d6b314`, whose subject is *"Remove FBS splice logic from CBS and production build"* — no rationale recorded anywhere.

Worse, the repo had **two acquisition paths producing different products**: `download_luh2.R` fetched `luh.umd.edu/LUH2/LUH2_v2h` (6,212,694,984 bytes, base release, 850–2015) while the pin held GCB2022 (850–2022). Nothing recorded which you had. That is #457 one level deeper than the issue framed it.

## Changes

- **`states.nc` is downloaded on demand** into `rappdirs::user_cache_dir("whep")/luh2` and verified against the published MD5 — the pattern `read_critical_n()` and `.provinces_shapefile()` already use. At 6.7 GB a cache hit is checked by **byte size** (re-hashing costs ~20 s per read); the full MD5 runs once on download, and a mismatching payload is **deleted** rather than left where the size check would later accept it.
- **`states_source = c("auto", "local", "zenodo")`**. `"auto"` prefers a `WHEP_LUH2_DIR` tree (it costs nothing); `"zenodo"` insists on the verified reference and is the lever for a reproducible run; `"local"` errors rather than downloading.
- **An off-vintage local tree warns.** Verification is the real gain here: which v2h tree produced a result used to be a recorded claim, and is now checkable.
- **`download_luh2.R` takes `states.nc` + `management.nc` from Zenodo with MD5 verification**, into an explicitly named `LUH2-GCB2022/` directory instead of reusing `LUH2 v2h/` for different content. `staticData_quarterdeg.nc` is published *only* under `luh.umd.edu/LUH2/LUH2_v2h` (404 under `LUH2_GCB_2022`) and is vintage-independent, so that one file still comes from there — keeping the `-k` workaround for luh.umd.edu's unverifiable TLS chain (confirmed still broken: `curl: (60) unable to get local issuer certificate`).
- `.read_file()` aborts for a missing `nc`/`nc4` member instead of returning `NULL` — the silent `NULL` is why the original failure surfaced away from its cause.
- Docs: stale corrupt-payload note replaced; `@source` cites Chini et al. 2021 (ESSD 13, 4175–4189, verified from the publisher) and the data DOI; `inst/soil_balance_inputs.md` documents the source table and the checksum; `CLAUDE.md`'s "readers abort when unset" claim now notes the two download-on-demand exceptions (it was already inaccurate for critical-N on main).

## Two bugs the tests caught before shipping

1. **The cli error paths were themselves broken.** `{.luh2_states_url()}` inside a cli string is read as a *style* by cli ≥ 3.4, so the download-failure abort and the off-vintage warning would both have raised a cli error instead of their message.
2. **R's 60 s default download timeout** would have killed a 6.7 GB fetch every single time. `.luh2_download_states()` now raises it for the duration and restores it, with a test asserting both.

## Verification

- **Download-then-verify and delete-on-MD5-mismatch run end-to-end against real Zenodo**, exercised with a small asset (record 6395016) so the code path is proven without moving 6.7 GB. The `states4.nc` URL serves `content-length: 6657587367` with the HDF5 signature `\x89HDF\r\n\x1a\n`.
- Year-2000 global cropland **1.467 Gha**, total land **12.993 Gha** — unchanged from the pin path (FAO ~1.5 Gha).
- `devtools::test()`: 124 files, 1663 tests, **5037 passed, 0 failed, 0 errors**.
- `rcmdcheck()`: 0 errors, 0 warnings, 2 NOTEs — **identical to the `origin/main` baseline** I ran for comparison.
- `lintr` clean; `air format .` and `devtools::document()` run; every `man/*.Rd` still in `_pkgdown.yml`.

## For you to decide

1. **The pin's bytes are still on SACO.** Unregistering is reversible; deleting 6.7 GB is not, so I left it. Say the word and I'll delete it, or leave it as a cold mirror.
2. **`WHEP_LUH2_DIR` now points somewhere new** if you re-run the downloader (`LUH2-GCB2022/` instead of `LUH2 v2h/`). Deliberate — the old path name would otherwise hold different content than before. Your existing setting keeps working; it will just warn if that tree is the base release.
3. `build_carbon_balance()` still doesn't propagate the provenance record into its own output, so writing a result to disk loses the vintage. Separate change — happy to file it.


lbm364dl:
Pushed `13b404d2` and did the disk cleanup.

## No download timeout

Replaced the 4 h cap with `options(timeout = 0)` — any finite budget is a guess about someone else's bandwidth, and at 6.7 GB the guess is load-bearing.

Verified `0` is genuinely unlimited rather than a silent fallback to the 60 s default, using a local drip server:

```
timeout=60  -> WARNING  after 60.0s, bytes=160/200   (cut off)
timeout=0   -> 0        after 75.1s, bytes=200/200   (completed)
```

That distinction mattered: the previous code did `max(old_timeout, budget)`, which with `0` would have quietly left the 60 s default in place. It is now set to `0` unconditionally and restored on exit, with a test asserting `identical(seen, 0L)`.

The trade is stated in the code: a *stalled* connection now hangs rather than erroring. An interrupted download is recovered by re-running; a timeout that fires on a merely-slow link can never be got past.

## Stale rasters removed — one copy left on the machine

There were **four** byte-identical 6.7 GB copies of `states.nc`, plus 3.3 GB of the corrupt pin payload. All MD5-verified against the published digest before anything was touched:

| location | size | action |
|---|---|---|
| `WHEP/LPJmL_inputs/LUH2/LUH2 v2h/states.nc` | 6.7 GB | **moved** → `~/.cache/whep/luh2/states.nc` (same filesystem, instant rename) |
| `~/Downloads/states.nc` | 6.7 GB | deleted |
| `~/.cache/pins/url/350fbd…/states.nc` | 6.7 GB | deleted (retired pin's cache) |
| `~/.cache/pins/url/c833cb…/*.csv,*.parquet` | 3.3 GB | deleted (cache of the **corrupt** version) |
| SACO board `luh2_v2h_states/` (both versions) | 9.3 GB | deleted, per your call |

Moving rather than deleting the tree copy means the canonical cache is already warm — **nobody has to download 6.7 GB**, including CI-less local runs.

Also renamed `LPJmL_inputs/LUH2/LUH2 v2h/` → `LUH2-GCB2022/`, matching what `download_luh2.R` now creates. Its `management.nc` MD5s to the GCB2022 digest, so the old name was describing the wrong product. No code referenced the path (only the product *name*, which is still correct where it appears). It keeps `management.nc` + `staticData_quarterdeg.nc`, which `prepare_spatialize_all.R` and `_gen_type_cropland.R` still need.

**Board hygiene:** deleting the pin folder would have left a dangling entry in the shared `_pins.yaml`, which could break board operations for the other 69 pins. I removed just the `luh2_v2h_states:` key and its two version lines — line-based, so the other entries stay byte-identical rather than being reformatted by a YAML rewrite — then confirmed the manifest still parses, that `luh2-areas` is untouched, and that the sync propagated (remote manifest: 69 entries, pin absent). Verified the board still serves other pins: `whep_read_file("luh2-areas")` returns 961,248 rows. A backup of the original manifest is in the scratchpad.

End state — one `states.nc` on the machine:

```
6657587367  /home/usuario/.cache/whep/luh2/states.nc
```

`read_luh2_landuse()` with nothing configured: `origin: zenodo | version: 10.5281/zenodo.15556812 | vintage: UofMD-landState-LUH2-GCB2022`, cropland 1.435 Gha.

Reclaimed ~19.3 GB local + 9.3 GB on SACO.

`devtools::test()`: 1663 tests, **5054 passed, 0 failed**, 9 skipped (4 fewer than before — the real-data checks now resolve from the cache too, so `test_luh2_landuse.R` skips nothing). `rcmdcheck()`: 0 errors, 0 warnings, 2 NOTEs, both inherited from main. `lintr`: 0 lints.

One consequence worth knowing: until this merges, `main` still lists `luh2_v2h_states` in `whep_inputs`, and those bytes are now gone — so anything on main reading that pin 404s. On main it already failed (that was the bug), but it's a reason not to leave this sitting open too long.

eduaguilera:
Thanks for digging past the framing in #457 — "the payload is corrupt" was me repeating the stale header note in `R/luh2_landuse.R` rather than checking the board, and your version table shows the payload was replaced on 2026-07-01, a day after that note was written. The real cause being `.luh2_read_states()` requesting `type = "parquet"` is a better bug than the one I reported.

Independent corroboration for the part that matters most here — **the two acquisition paths producing different products**. Working on the polycell spatial support branch, I found the same thing from the data side before this PR existed, and it changed a decision:

- Two v2h trees are on this machine: `C:\XL_files\LUH2\LUH2 v2h` reports `source_id = UofMD-landState-LUH2-GCB2022` with 1,173 time steps (850–2022), and `C:\XL_files\whep_lpjml\input\LUH2\LUH2 v2h` reports `LUH2 v2h` with 1,166 steps (850–2015).
- They are **not interchangeable for validation**. The identity `sum(12 states) + icwtr = 1` at 2015 gives residual mean `5.07e-10`, range `[-5.57e-07, 2.35e-05]` on the GCB2022 tree, and mean `-6.91e-11`, range `[-5.63e-07, 1.72e-06]` on the base tree. A preflight number quoted without naming the vintage is therefore unreproducible, which is exactly what happened to us.
- `staticData_quarterdeg.nc` is byte-identical in both, so anything derived from it — terrestrial area, `carea`, `icwtr` — is vintage-independent. Global land from `(1 - icwtr) * carea` is 12.9931 Gha either way. Useful to know that the vintage question only bites the states, not the static layer.

Given the MD5 agreement you show between Zenodo, the retired pin and the local tree, our branch's default (GCB2022, recorded in an output column so a result names its own vintage) lines up with what this PR makes the verified reference. The `states_source = c("auto", "local", "zenodo")` argument is the same shape we arrived at independently, so no conflict — we will drop our own `@source` correction in favour of this once it lands.

One small thing from our side that supports retiring the pin: `staticData_quarterdeg.nc` being published only under `LUH2_v2h` and not `LUH2_GCB_2022` matches what we see on disk, and since it is vintage-independent that asymmetry is harmless rather than a gap.

lbm364dl:
Merged `main` in — **`MERGEABLE`**, no conflicts remaining. `BLOCKED` is just `REVIEW_REQUIRED`, not a merge problem.

## The conflicts, and how each was resolved

Four files conflicted. All were resolved by keeping **both** sides, not by picking one:

**`R/luh2_landuse.R`** — the substantive one. Main added `.add_reporting_polity_columns()` to the return pipe (#424); this branch added `attach_provenance()`. Both apply, and **the order is load-bearing**: `.add_reporting_polity_columns()` reshapes the table, so an attribute attached before it would not survive. Provenance goes last:

```r
.luh2_to_polity(grid, country_grid, resolution) |>
  .add_reporting_polity_columns() |>
  attach_provenance(get_provenance(states))
```

Verified on a real read that both features coexist at both resolutions — the polity columns are present *and* `get_provenance()` still returns `origin: zenodo | version: 10.5281/zenodo.15556812`, with cropland unchanged at 1.435 Gha.

**`CLAUDE.md`** — main's new bullet ("LPJmL outputs are the one input a user cannot obtain, so they *are* pinned") and this branch's note on the two inputs that download-on-demand are the two halves of the same rule. I split the download-on-demand text into its own bullet next to main's rather than leaving it buried in the env-var paragraph.

**`data/whep_inputs.rda`** and **`man/read_luh2_landuse.Rd`** — generated files, so regenerated (`data-raw/whep_inputs.R`, `devtools::document()`) rather than hand-resolved. `whep_inputs.csv` auto-merged: main's two new `lpjml-*` pins plus this branch's removal of `luh2_v2h_states` = 71 rows.

## One thing to know

Main was red when I merged: `test_exported_polity_columns.R` failed because `read_population()` didn't carry the polity columns. I confirmed it failed identically on clean `origin/main` (1 failed / 28 passed), so it wasn't from this branch, and I fixed it here since it would have made this PR's CI red.

**Then #502 landed the same fix on main while I was running the suite.** I dropped mine and took main's, which is better — it also asserts `expect_false(anyNA(out$reporting_polity_code))`. So there's no duplicate fix in this branch; `R/population.R` and its tests are byte-identical to main.

## Verification on the merged state

- `devtools::test()`: 129 files, 1725 tests, **5356 passed, 0 failed, 0 errors**, 9 skipped.
- `test_luh2_landuse.R` alone: 30 tests, 84 assertions, **nothing skipped**.
- `rcmdcheck()`: **0 errors, 0 warnings, 2 NOTEs** — the same two inherited from main (`readxl` unused import, `.sci_sum_components` globals).
- No conflict markers anywhere; `air format .` and `devtools::document()` clean.

One practical note: main is advancing fast — it moved 23 commits during a single full-suite run, and twice mid-task. This branch merges cleanly against current main and I've verified this state end to end, but if it sits for long the merge is worth redoing at merge time rather than me chasing it.

lbm364dl:
@eduaguilera — ready to merge, and blocked only on a formal review.

Your comment above reads as a substantive review, but GitHub still records `reviewDecision: REVIEW_REQUIRED` with 0 reviews, so the base-branch policy refuses the merge. Could you re-submit it via **Files changed → Review changes → Approve**? That's the only thing outstanding.

State right now: `MERGEABLE`, no conflicts, CI **9/9 green**. Verified locally on the merged state: 1725 tests / 5356 passed / 0 failed, and `rcmdcheck()` 0 errors, 0 warnings, 2 NOTEs — the same two inherited from main.

I can't do it from here: I have `push`/`triage` only, so `--admin` is not available to me, and repo-level auto-merge is disabled so I couldn't arm it to land on your approval either.

Two notes for whenever you get to it:

1. **Timing matters a little.** `main` still lists `luh2_v2h_states` in `whep_inputs` and those bytes are deleted from SACO, so anything on main reading that pin 404s until this lands. It already failed there — that was the bug — but the window is worth closing.
2. **One open design point** that your comment treats as settled and I don't think is. You describe your polycell branch recording the vintage "in an output column"; this PR records it in a provenance *attribute* via `get_provenance()`. Those are different mechanisms for the same fact, and yours is arguably better — a column survives being written to disk, an attribute does not (flagged in the PR body). No rush and nothing incorrect either way, but worth deciding when your branch lands rather than shipping both.

Also, unremarked in your comment but a useful corroboration: you derive global land as `(1 - icwtr) * carea` = **12.9931 Gha**, and this PR's test measures **12.993 Gha** by summing the 12 states with its own spherical `.luh2_cell_area_ha()`. Two independent derivations agreeing to four significant figures — that validates the area formula against LUH2's own `carea` layer, which nothing here had checked.

Issue #458, opened 2026-08-04
[epic] Every territorial identity in whep carries a polity code string


lbm364dl:
## Goal

**Every row, cell, coefficient and join key in whep that identifies a territory must carry a polity code string** — `ESP-1846-1914`, `SDN-2011-2025` — not a raw FAOSTAT area code, an ISO3 string, a country name, a COW code, a hand-maintained region table, or a static present-day grid.

This is deliberately stronger than "adopt `polity_area_code` repo-wide", which is what #290 proposed and why that issue is closed in favour of this one. **`polity_area_code` is not a polity.** It is `coalesce(fabio_code, area_code)`, a numeric aggregation bucket. Measured on the shipped crosswalk:

- 544 distinct `polity_code` collapse into 201 `polity_area_code`
- `(polity_area_code, year)` recovers 194 of 201 buckets — the temporal folds are fine
- **7 buckets it cannot recover.** Bucket `999` holds up to **17 territories simultaneously**; bucket `206` holds Sudan and South Sudan simultaneously 2011-2024. The other five (15, 28, 101, 170, 273) are crosswalk overlapping-period **defects**, not designed folds
- Of the 62 area codes folded into `999`, **16 already have their own real polity code** — Bermuda `BMU-1684-1968`, Faroe Islands `FRO-1800-2025`, Eq. Guinea `GNQ-1886-1968`, French Guiana `GUF-1816-1946`, Guadeloupe, Guam, Cook Islands. The polity exists; the numeric fold discards it

So the numeric key is a legitimate FABIO-matrix-grain aggregation key, and it must stop being the thing rows are *identified* by. That makes #419 and #414 critical path rather than side quests: until the folds stop, attaching a polity string to that data just writes `ROW-1850-2023`, which is not a territory.

**Gridded data is in scope.** Where no historical boundary exists, resolving to the present-day polity is acceptable — but as an explicit, documented choice with a real `polity_code` column, never as an unstated side effect of joining against a present-day-only lookup.

## Already polity-keyed on main — do not redo this

Confirmed by direct measurement / code reading; no PR needed:

- **Core FAOSTAT ingestion**: `get_primary_production()`/`build_primary_production()` and every
  builder downstream of `.aggregate_to_polities()` (`R/read_raw_inputs.R:297-349`,
  `R/polities.R:37-219`) — year-aware interval join, `polity_area_code` already renamed onto
[...]

Issue #387, opened 2026-07-28
Pre-independence FAOSTAT years resolve to the modern polity: one area maps to only one prefix


lbm364dl:
Found while working #382, by cross-checking against whep-polities' own FAOSTAT matching pipeline (`pipelines/faostat-era-matching/state/faostat_aliases.csv`). `test_upstream_faostat_agreement.R` in #382 now tracks this.

## The bug

`polity_area_crosswalk` maps each FAOSTAT area to **exactly one polity prefix**, then joins every polity sharing it. That works when a chain keeps one prefix (`F51` Czechoslovakia, `FRS` Djibouti). It fails when upstream gives the colonial and modern polities **different** prefixes:

| area | upstream colonial polity | upstream modern | prefix WHEP carries |
|---|---|---|---|
| 7 Angola | `ANG-1905-1975` | `AGO-1975-2025` | `AGO` only |
| 20 Botswana | `BEC-1885-1966` | `BWA-1966-2025` | `BWA` only |
| 181 Zimbabwe | `SRH-1953-1964` | `ZWE-*` | `ZWE` only |
| 206 Sudan | `SUD-1956-2011` | `SDN-2011-2025` | `SDN` only |
| 237 Viet Nam | `F237-1954-1975` | `VNM-*` | `VNM` only |
| 249 Yemen | `F249-1918-1990` | `YEM-1990-2025` | `YEM` only |
| 251 Zambia | `NRH-1953-1964` | `ZMB-1964-2025` | `ZMB` only |

A single prefix cannot reach the colonial polity, so pre-independence years fall through to the modern one:

```r
add_polity_code(data.frame(area_code = c(7L, 206L, 181L),
                           year = c(1965L, 1970L, 1962L)), "area_code", "year")
#  area 7,   1965 -> AGO-1975-2025  "Angola (independent, 1975-2025)"
#  area 206, 1970 -> SDN-2011-2025  post-secession Sudan
#  area 181, 1962 -> ZWE-1964-1980  span starts 1964
```

**This is not data being dropped** — that would be visible. It is data attributed to a polity that did not exist. Sudan is the worst case: 1970 FAOSTAT "Sudan" covered the territory that became South Sudan in 2011, and it is being attributed to `SDN-2011-2025`, which by definition excludes it. Same class of error as the retired-polity routing that motivated the upstream manifest.

Upstream already made the correct determination for every one of these areas. This package simply cannot express it.

## Two candidate fixes

[...]

PR #479, opened 2026-08-04
Resolve country labels to polities instead of dropping them

### LINKED ISSUE #389: Three datasets carry area identifiers that resolve to no polity (latent, 779 rows)
Found while working #382 by sweeping every exported dataset for columns naming a country, area or region. `test_dataset_area_identifiers.R` in #382 baselines all three.

| dataset | column | unresolved | rows |
|---|---|---|---|
| `mueller_synthetic_n` | `iso3c` | 10 codes | 328 / 5043 |
| `lassaletta_grassland_share` | `Country` | 9 names | 441 / 6909 |
| `urban_n_reference` | `area_code` | `"ESP"` | 10 / 10 |

**`mueller_synthetic_n$iso3c` is not ISO3.** Ten values are FAO-style legacy codes:

```
BZE -> BLZ (Belize)      COS -> CRI (Costa Rica)    ELS -> SLV (El Salvador)
GUA -> GTM (Guatemala)   HAI -> HTI (Haiti)         HON -> HND (Honduras)
ROM -> ROU (Romania)     SRM -> SUR (Suriname)      TRI -> TTO (Trinidad)
ZAR -> COD (Zaire, renamed 1997)
```

**`lassaletta_grassland_share$Country` holds name variants:** `Cape Verde` (Cabo Verde), `Swaziland` (Eswatini), `Turkey` (Türkiye), `Reunion` (Réunion), `Cote d'Ivoire` (unaccented), `Belgium-Luxemburg`, `DPRepublic of Korea`, `Occupied Palestinian Territory`, and `FSU` — which is an aggregate, not a country, so it needs an aggregate polity rather than a spelling fix.

**`urban_n_reference$area_code` holds `"ESP"`.** Everywhere else in this package `area_code` means the numeric FAOSTAT reporting area. A consumer joining on it gets zero rows, silently. It would resolve fine as an `iso3`, which is what it is.

## Severity: latent, not active

None of these is joined by anything in `R/` today — `mueller_synthetic_n` and `lassaletta_grassland_share` have no consumers at all, and `urban_n_reference` appears only in documentation of a derivation. **No output is currently wrong.**

They matter because they are primed for the first person who writes the obvious join. That is exactly how `regions_full$polity_code` — 271 bare ISO3 prefixes in a column named for polity codes, resolving to nothing — sat unnoticed until #382 renamed it to `polity_prefix`.

## Suggested fix: upstream aliases, not a lookup here

whep-polities already resolves source labels to polities through its alias mechanism, and every one of these is a textbook alias case. Hand-mapping them in this package would create a second authority for the same question — the mistake that produced the Sudan misattribution in #387.

So: add aliases upstream for the 19 labels, then resolve these datasets through the same path as everything else. `FSU` needs a decision rather than an alias (which aggregate polity, over which years).

For `urban_n_reference`, the column should probably be renamed `iso3` — but it ships in a public dataset, so that is a maintainer call.

lbm364dl:

## What was wrong

`add_polity_code()` resolves numeric FAOSTAT/FABIO area codes. Nothing resolved a country
**label**, so every dataset naming its territories in words carried an identifier that
resolved to no polity.

## Evidence it reproduced, before any change

Measured on `polity/389` at its base commit (`a9389b4f`), with `pkgload::load_all()`:

```
--- exists resolve_polity_label: FALSE
--- exists polity_label_aliases: FALSE

== mueller_synthetic_n: 5043 rows
iso3c not in polities$iso3_code: 10  BZE COS ELS GUA HAI HON ROM SRM TRI ZAR
rows affected: 328 / 5043

== lassaletta_grassland_share: 6909 rows
Country not matching any polity_name (normalised): 13
  Belgium-Luxemburg, Cape Verde, DPRepublic of Korea, Ethiopia PDR, FSU,
  Lao People's Democratic Republic, Occupied Palestinian Territory, Oman,
  Republic of Korea, Swaziland, United Republic of Tanzania, Viet Nam, Yugoslav SFR
rows affected: 637 / 6909

== urban_n_reference: 10 rows
area_code = "ESP" (character) in a column that means the numeric reporting area
everywhere else in the package -> 10 / 10 rows join to nothing
```

`lassaletta` measured the issue's other way too -- exact string match against
`regions$area_name`, which is what `inst/scripts/prepare_spatialize_all.R:2033-2040` does:
**6,370 / 6,909 (92.2%)**, i.e. 539 rows dropped, matching the issue's number.

## What I changed

- `resolve_polity_label(label, source, year)` in `R/polities.R`, exported.
- `polity_label_aliases` (869 aliases over 470 labels): loading, `col_types` contract and
  target validation in `data-raw/table_mappings.R`, roxygen in `R/table_mappings.R`, plus the
  built `data/polity_label_aliases.rda`.
- `.add_reporter_polity()` in `R/sources.R`, wired into `expand_trade_sources()`.
- `tests/testthat/test_resolve_polity_label.R`, `_pkgdown.yml` index entries, `globalVariables`.

The alias map is **embedded, not recomputed**. A label's meaning is a fact about the source and
whep-polities already decides it; a second lookup here would be a second authority for the same
question, which is the mistake behind the Sudan misattribution in #387.

### Ported from #382, adapted in three places

`main` has moved, so this is not a verbatim port:

1. **The aggregate-refusal guard.** #382 derived it from a separate published contract
   (`faostat_deliberate_area_codes`) that does not exist on `main`. Here it is derived from
   `polity_area_crosswalk`'s own `mapping_status == "unmapped"` rows -- today exactly FAOSTAT 351
   "China". Same effect, one authority, and it tracks the crosswalk automatically.
2. **No `sf` dependency.** `sf` is only in `Suggests`, so `polities` is read column-by-column
   rather than through `sf::st_drop_geometry()`.
3. **An ISO3 fallback, which #382 did not have.** Without it the primitive resolved **380 of
   mueller's 5,043 rows** and **860 of crops_manure's 31,648** -- the alias map is keyed on the
   labels a curator had to decide about, and a plain current ISO3 is not one of them. That is not
   a usable label route for the datasets this issue is about. It mirrors the second half of
   upstream's own rule ("alias, then ISO/name family + year containment") under the same two
   guards: exactly-one-live-polity, and a curated rule covering that year outranks it.

## How I verified

Coverage, measured on this branch with the shipped 603-polity snapshot:

| source | column | before | after |
|---|---|---|---|
| `mueller_synthetic_n` | `iso3c` | 0 | **4,999 / 5,043** (99.1%) |
| `crops_manure_n` | `ISO` | 0 | **31,476 / 31,648** (99.5%) |
| `lassaletta_grassland_share` | `Country` | 6,370 (92.2%) | **6,781 / 6,909** (98.1%) |
| `urban_n_reference` | `area_code` | 0 / 10 | **10 / 10** |
| `expand_trade_sources()` | `Reporter` | no column | **1,133 / 1,133** |

The issue's four named `lassaletta` cases resolve: `Ethiopia PDR` 1980 -> `ETH-1952-1993`,
`Sudan (former)` 2000 -> `SUD-1956-2011`, `Occupied Palestinian Territory` 2000 ->
`PSE-1948-2025`, `Swaziland` 2000 -> `SWZ-1894-2025`. `crops_manure_n`'s 172 "RoW" rows, which
were dropped unconditionally, resolve to `ROW-1850-2023` through a curated `crops-manure-n`
alias -- so the RoW policy question the issue raises is already answered upstream, not here.

What stays `NA` stays `NA` on purpose: dissolved aggregates a source keeps reporting past their
end (FSU after 1991, Czechoslovakia after 1993, Yugoslav SFR, Belgium-Luxemburg, Ethiopia PDR),
South Sudan before 2011, and `"SDN"` in 2000 -- the only polity carrying that ISO3 starts in
2011, so a guess would repeat #387. That is 128 of `lassaletta`'s 6,909 rows.

Gates, all locally green:

- `air format` on every file touched -- clean.
- `lintr::lint_package()` with the exact CI config -- **no lints**.
- `test_resolve_polity_label.R` -- **36 assertions, 0 failures**. With the `R/` changes stashed
  and the test kept, **13 fail**, so the test would not pass without this change.
- `test_polities.R`, `test_sources.R`, `test_polity_output_coverage.R` -- 82 assertions, 0
  failures, 0 warnings (the missing-`Reporter` warning an earlier draft introduced into
  `test_sources.R` is gone).
- `devtools::document()`; `man/` and `NAMESPACE` committed, `RoxygenNote` unchanged.
- No non-ASCII anywhere in the added R code or comments.

**No published value moves.** Every change is additive: a new exported function, a new dataset,
and one new column on `expand_trade_sources()` output. No existing call site was rewired, so
there is nothing to compare magnitudes on.

## Draft, because of an upstream merge-order dependency

`data/polity_label_aliases.rda` was built from `label_alias_map.csv` as it stands on
**whep-polities PR #39** (`polities-consumer-contract`), which is **still open**. The file is not
on that repo's `main`, so:

- `data-raw/table_mappings.R` now aborts with an explanatory message (naming the file, the
  gating script, and the `WHEP_POLITIES_LABEL_ALIAS_MAP` override) rather than silently building
  an empty alias table. Regeneration is blocked until #39 merges; **use** is not, because the
  built `.rda` is committed.
- All 869 alias targets were checked against whep-polities `main`'s 740-polity database: **0
  unknown**. So the map is consistent with upstream `main` today; only the *file* is unmerged.

**This blocks merge.** whep-polities#39 should land first, and this branch should then be
regenerated from it so the embedded copy provably matches a merged revision.

## Two things I deliberately did not do

1. **The three `inst/scripts/prepare_spatialize_all.R` call sites** (`mueller`, `crops_manure`,
   `lassaletta`) are **not** rewired, and their hand-written 14-entry `recode()` list at
   `:2054-2070` still stands. They resolve labels to a numeric `area_code` for a downstream grid,
   not to a polity, so swapping in `resolve_polity_label()` is not a mechanical substitution -- it
   needs either a polity -> area bridge or a restructuring of the nitrogen spatialization, and it
   moves published values (`lassaletta` would gain 411 rows). That is a separate PR with its own
   magnitude comparison, and it is where the `needs-expert` label really applies.
2. **`urban_n_reference$area_code` is not renamed to `iso3`.** It ships in a public dataset and
   the issue calls that a maintainer decision. `resolve_polity_label()` now resolves the column's
   contents regardless, so the latent join failure is addressed without a breaking rename.

One thing worth flagging separately: this package's `data/polities.rda` is a **603-polity**
snapshot while upstream `main` publishes **740**. 225 of the 869 aliases point at one of 115 codes
`get_polity_geometries()` cannot yet return a row for. That is pre-existing staleness, documented
in `?resolve_polity_label`, and refreshing `polities` closes it without changing any resolution --
but the refresh regenerates `polity_area_crosswalk` too, so it belongs in its own PR under #458.

Closes #389


lbm364dl:
## The stated blocker is cleared — but a different one appeared, so I have left this as a draft

**whep-polities#39 is merged.** `label_alias_map.csv`, `faostat_area_polity_map.csv` and `polities_manifest.json` are published on `main`, so the merge-order dependency this PR was drafted for is gone. Verified against merged upstream:

- committed `polity_label_aliases`: **869 rows / 14 sources** — matches upstream exactly
- `resolve_polity_label()` works: `"Ivory Coast"` @2000 → `CIV-1960-2025`, `"Czechoslovakia"` @1985 → `F51-1947-1993`
- this PR's claim that all 869 alias targets resolve against upstream's 740-polity database is **correct**

## Why it should still wait

The package ships a **stale** `polities` snapshot: **603 rows against upstream's 740**. That means **115 of the 869 alias targets point at polities `whep::polities` does not contain** — `ETH-1907-1936`, `ALK-1867-1959`, `AOI-1936-1941`, `ASM-1900-2025` and others.

So merging this now would ship a resolver that can hand back a polity code the caller cannot then look up. Nothing in this PR causes that — the stale snapshot does, and this PR had no reason to suspect it — but this PR is what makes it reachable.

**#485 refreshes the snapshot** (603 → 740, crosswalk 550 → 621) and is itself draft, because the refresh surfaced three upstream data defects now filed as lbm364dl/whep-polities#58 (overlapping polity periods on areas 15 and 248, making `add_polity_code()` ambiguous) and #59 (14 polities declaring a `polygon_status` with no geometry).

## Suggested order

1. lbm364dl/whep-polities#58 and #59 — fix the ambiguous data
2. #485 — refresh the snapshot, with a magnitude comparison
3. **this PR** — then the 115 dangling targets do not exist and it can merge as-is

I was asked to mark this ready for review and deliberately have not, because the dangling references are a real defect rather than a cosmetic one. If you would rather land it first and accept them temporarily, it is one command — `gh pr ready 479` — and nothing here needs changing.

lbm364dl:
Brought up to date with `main` (99 commits). Full suite **5460 pass, 0 fail, 8 skip**, lintr and air clean, `man/` regenerated.

## Your declared blocker is gone

[lbm364dl/whep-polities#39](https://github.com/lbm364dl/whep-polities/pull/39) **has merged**, so `label_alias_map.csv` is on that repo's `main` and the abort message now names a file that exists.

## A narrower one replaced it, and it is #530

Regenerating against upstream `main` aborts on **this branch's own target validation** — which is the guard working. The published map is now 891 rows over 468 labels (it was 869/470 when you built the `.rda`), and it names three polities the embedded 740-row vintage does not hold:

```
CAN-1886-1949   CAN-1949-2025   ROW-1850-2025
```

All three are retirements tracked by #530, and PR #551 is landing that re-sync. So the committed `.rda` stays, and the structure of your original argument is unchanged: **measured, 0 of its 869 targets are missing from the shipped `polities`**, so use is unaffected and only regeneration waits.

## Three coverage assertions moved — none because the resolver changed

**`mueller_synthetic_n`: 4832 of 5043, not the 4999 pinned.** The 167-row gap is your ambiguity guard firing on an upstream defect. Five ISO3 families ship a spanning row alongside the finer split of the same family, so two live polities cover 2000 and the resolver refuses to pick by row order:

| ISO3 | rows | overlap |
|---|---|---|
| BRA | 54 | `BRA-1800-2025` beside `BRA-1909-2025` |
| ARG | 44 | `ARG-1800-2025` beside `ARG-1902-2025` |
| AGO | 41 | `AGO-1816-2025` beside `AGO-1975-2025` |
| GRC | 39 | `GRC-1919-2025` beside `GRC-1947-2025` |
| IRQ | 33 | `IRQ-1921-2025` beside `IRQ-1932-2025` |

Measured against upstream `main`: **all five spanning rows are now retired or superseded**, leaving exactly one live polity covering 2000 in each family. So #530 lifts this back above 4,999 with no resolver change. I lowered the bound to 4832 and recorded why; it stays a lower bound precisely so the re-sync raises it without a test edit.

**`resolve_polity_label("SDN", year = 2000)` now returns `SUD-1956-2011` rather than NA** — and that is the outcome #387 wanted, reached upstream instead of here. `SUD-1956-2011` has gained `SDN` as its ISO3, so a pre-secession year resolves to the unified Sudan that actually existed rather than being refused for want of a candidate. The guard itself is untouched; Montenegro still tests it.

**`urban_n_reference` is fixed on `main` and no longer needs this route at all.** It shipped `area_code = "ESP"` in a column meaning the numeric FAOSTAT area; it now ships `area_code = 203L` and carries `polity_code = "ESP-1800-2025"` directly. The assertion that the label route rescued its 10 rows is replaced by one pinning the dataset's fixed shape, so the regression would still be caught.

Worth noting for the merge order: this is now the third PR in the epic whose numbers are waiting on #530 (with #480 and #302). #551 landing first would let all three record their final figures.

Issue #460, opened 2026-08-04
Route the shared cell↔area grid (build_cell_polity/build_cell_polity_fraction) through the polity crosswalk


lbm364dl:
> [!IMPORTANT]
> **Target state superseded: this is delivered by the polycell work on branch `edu/polycell-spatial-support` (@eduaguilera).** Do not start editing `build_cell_polity()` / `build_cell_polity_fraction()`.
>
> The original text below proposed resolving the grid to `polity_area_code` as an explicit **present-day snapshot**, on the grounds that "the grid itself carries no year". That is now rejected as the target. `polity_area_code` is `coalesce(fabio_code, area_code)` — a numeric bucket in which `999` holds up to **17 territories simultaneously** and `206` holds Sudan and South Sudan together 2011-2024 — so a snapshot keyed on it cannot carry the polity code string that epic #458 requires.
>
> **The adopted unit is the polycell:** `polycell_id = f(cell_id, polity_code)`, the intersection of a 0.5° cell with a polity over that polity's validity interval, cut by true geodesic intersection against the polygons `get_polity_geometries()` already ships, with territory decomposed into `polity_area_ha = land_area_ha + inland_water_ha + ice_area_ha`. The grid therefore *does* gain a year dimension, and `RUS-1991-2014` and `RUS-2014-2025` are distinct polycells in the same physical cell.
>
> Measured on `main` and relevant to anyone reading this issue: the whole-cell base **over-counts land by 11%** (12.9931 Gha terrestrial vs 14.3195 Gha whole-cell); the border-split quantum is **1/36, not 1/144** (`subcells = 6L`, `inst/scripts/prepare_spatialize_all.R:949`); and **3,925 multi-polity cells (6.1%) hold 7.4% of global terrestrial area**, with 23 of 191 polities holding over half their land in shared cells (Singapore and Macao 100%).
>
> `data$cell_polity` is a **shared object** — `build_n_deposition()`, `build_urban_n()`, `spatialize_country_n_to_crops()`, `build_water_balance()`, `read_soil_ph()`, `read_soil_hydraulic()`, the carbon clay path, and `build_ag_land_support()`. Reshaping it in place would change every unmigrated consumer on one commit and make per-consumer attribution impossible, so the transition uses a superset shim that reproduces today's crosswalk bit-for-bit.
>
> **Retained here as the tracking issue for that work.** Original text follows for the record.

---

[...]

Issue #444, opened 2026-07-31
polities-integration moves feed 7% and production 2% against main, and the cause is not identified


lbm364dl:
Splitting this out of #382 because it invalidates that PR's central merge argument and needs deciding on its own.

## The claim that failed

#382 says "It changes no published numbers — every column total of a full-range `get_wide_cbs()` is within 1% of `main`". That was measured **before #425 was fixed**, when every cast value was a row count. Row counts compare group *sizes*, not quantities, so the measurement could not have detected a value change; it certified that rows-per-group barely moved and was read as certifying value-neutrality.

Re-measured on the fixed build, full range, `main` at `3f479f7a` with #429 in it:

| metric | main | branch | ratio |
|---|---|---|---|
| `feed` | 9.0636e11 | 9.6982e11 | **1.070** |
| `other_uses` | 9.2130e10 | 9.4941e10 | **1.031** |
| `production` | 3.0076e12 | 3.0674e12 | **1.020** |
| `domestic_supply` | 3.1490e12 | 3.2098e12 | **1.019** |
| `stock_withdrawal` | 1.4828e11 | 1.4908e11 | 1.005 |
| `export` | 7.9264e10 | 7.8750e10 | 0.994 |
| `food`, `seed`, `processing`, `import`, `stock_addition` | | | within 0.5% |
| rows | 2,019,253 | 1,999,609 | 0.990 |
| **areas** | 195 | 195 | 1.000 |
| **supply-use imbalance > 50%** | 47,514 | 47,514 | **identical to the row** |

Four columns are outside the 1% band. This is far outside #420's noise, which is 0.08% on rows and ~5e-7 on values.

## Not main's own recent work

My first hypothesis was that main had moved under the branch — it had, by 12 commits including #427's LPJmL 6.11 radiation migration and the `consolidate_sources` changes. Tested by merging current main into the branch and re-running against the pre-merge baseline: **all 13 metrics within 0.15%**, `production` ratio 1.0000. Main's new work does not move commodity balances here. The gap is the branch's.

## Isolated: about one third code, two thirds data

Using the swap method (branch code + main `data/`, with `R/sysdata.rda` kept from the branch — it differs only by three new objects, every shared object byte-identical, so the isolation is clean):

| metric | main | branch code + main data | branch | code effect | data effect |
|---|---|---|---|---|---|
| `feed` | 9.0636e11 | 9.2060e11 | 9.6982e11 | +1.6% | **+5.3%** |
[...]

Issue #419, opened 2026-07-30
Areas FABIO folds into ROW: their observed data is attributed to ROW, and promoting them inflates global feed 13.7x


lbm364dl:
## Summary

Areas that FABIO folds into rest-of-world (`fabio_code` 999) but that report data of their
own have their data summed into area 999 and attributed to `ROW-1850-2023`. Promoting them
to their own numeric `polity_area_code` — the obvious fix — inflates global totals by up to
13.7x, so the attribution problem is left OPEN rather than traded for a larger error.

Both halves are measured. This issue records them so the modelling decision can be made on
numbers rather than on the shape of the code.

## The attribution problem (real)

`polity_area_code` is the numeric key the builds actually key on:
`get_primary_production()` emits it AS its `area_code`, and `build_trade.R` assigns
`area_code := polity_area_code` outright. So an area folded at this level is folded in the
output no matter what `polity_code` says.

The crosswalk resolves the Faroe Islands to `FRO-1800-2025` and Palestine to
`PSE-1948-2025`, but their numbers still land in 999: **2,458** raw production rows for the
Faroe Islands and **9,606** for Palestine. Nothing warns, and "0 rows unresolved" actively
conceals it, because 999 resolves to ROW perfectly well.

## Why promoting them is not the fix (measured)

Full-range `get_wide_cbs()`, promoting the 16 such areas, against `main`:

| column | main | promoted | ratio |
|---|---|---|---|
| feed | 8.232e10 | 1.132e12 | **13.7x** |
| export | 3.011e10 | 3.988e11 | **13.2x** |
| production | 1.736e12 | 3.172e12 | **1.8x** |
| domestic_supply | 1.729e12 | 2.796e12 | 1.6x |
| other_uses | 8.305e9 | 2.634e10 | 3.2x |
| processing | 1.6367e12 | 1.6363e12 | 1.00 |
| import | 1.5693e10 | 1.5720e10 | 1.00 |

Isolated to that one expression by swapping datasets one at a time: **branch code + main
data reproduces main's totals exactly**, and refolding these areas — with every other
change in place — returns every column to main's total. No R change moves a number; the
crosswalk's numeric key does all of it.

## It is an artifact, not recovered data

The entire `feed` increase lands on **one area, 212 Syrian Arab Republic: 1.049e12** —
twelve times the world total inside a single country. Decomposed:

- concentrated in item codes **2590** and **2598** (870e9 and 174e9 of it)
[...]

Issue #414, opened 2026-07-30
Bucket 206 sums Sudan + South Sudan but is attributed to SDN-2011-2025: the value and its polity describe different territories


lbm364dl:
Surfaced while fixing the duplicate-key defect in #382 (`.aggregate_to_polities()` and
`.get_traded_residues()` grouped by the reporting bucket **and** the area name, so folded
areas were never summed). With that fixed, a question the old behaviour was hiding becomes
answerable and needs a modelling decision.

### What happens now

`polity_area_code` inherits `fabio_code`, and FABIO's Sudan region is **206**:

| FAOSTAT area | name | `fabio_code` | polity |
|---|---|---|---|
| 206 | Sudan (former) | 206 | `SUD-1956-2011` (and earlier periods) |
| 276 | Sudan | 206 | `SDN-2011-2025` |
| 277 | South Sudan | 206 | `SSD-2011-2025` |

So from 2014, when both successors report, a 2015 bucket-206 row now correctly holds
**Sudan + South Sudan summed** — that is what a FABIO region means, and the fold is
deliberate.

But the crosswalk maps `area_code` 206 for 2015 to **`SDN-2011-2025`**, i.e. Sudan *without*
South Sudan. `.add_reporting_polity_columns()` therefore labels a two-territory value with a
one-territory polity. The number and its polity attribution describe different extents, and
nothing warns.

### Why it matters beyond tidiness

The polity columns exist so a consumer can ask "whose territory is this?". For bucket 206
post-2011 the honest answer is "Sudan and South Sudan", and there is no polity that means
that — `SUD-1956-2011` did, but it has ended. Anything joining on `reporting_polity_code`
and then using polygon area, population or a per-capita denominator gets Sudan's extent for a
value covering both.

### Options, none obviously right

1. **Un-fold 276/277** so each keeps its own reporting code, as the 17 areas that already
   opt out of FABIO's folding do. Cleanest semantically; changes the matrix regionalisation
   and diverges from FABIO.
2. **Give the bucket an aggregate polity** for 2011+, the way `BLX-1850-1999`
   "Belgium-Luxembourg" and `ANT-1961-2010` serve FAOSTAT 15 and 151. Keeps FABIO's regions
   and makes the attribution truthful; needs a new polity upstream.
3. **Leave it and document it**, with a warning where a folded bucket's polity covers less
   than the bucket does.

Option 2 matches how this repository already handles exactly this shape, so it is the one I
[...]

PR #480, opened 2026-08-04
fix(polities): give regions.csv's grid countries and pre-2011 Sudan their real polities

### LINKED ISSUE #459: Fix polity_area_crosswalk for regions.csv's 9 present-day-invalid area codes
**Problem**
`inst/extdata/regions.csv` (209 present-day area codes, no year column) feeds every static
country-grid pin in the codebase. 9 of its codes resolve, via `whep::polity_area_crosswalk`, to
the wrong polity: 7 (Eq. Guinea 61, French Guiana 69, New Caledonia 153, N. Macedonia 154,
Eswatini 209, Syria 212, Palestine 299) fall through to `polity_area_code=999`/`ROW-1850-2023`
("Rest of World", non-territorial, no geometry) for *every* year, even though whep::polities has
real dedicated polities for all of them. 2 (Sudan 276, South Sudan 277) resolve via the crosswalk's
nearest-match fallback to the post-2011 polities `SDN-2011-2025`/`SSD-2011-2025` for *all* years
including pre-2011.

**Evidence**
- `inst/extdata/regions.csv:120-121,163,166,199-203` — the 9 rows.
- MEASURED: `whep::polity_area_crosswalk[area_code %in% c(61,69,153,154,209,212,276,277,299)]`
  returns `polity_area_code=999` for 7 codes and `SDN-2011-2025`/`SSD-2011-2025` (single crosswalk
  row, `start_year=2011`) for the other 2.
- `whep::polities` contains real dedicated polities for these territories (e.g. `SWZ-1894-2025`,
  `SYR-1967-2025`/`SYR-1946-1967`/`SYR-1922-1945`/`SYR-1920-1922`), confirming this is a crosswalk
  gap, not an absence of a real polity to route to.

**Target state**
In `data-raw/harmonization_tables.R`/`data-raw/table_mappings.R` (or upstream in whep-polities):
give area_code 61/153/154/209/212/299 real per-country `polity_area_code` mappings instead of 999;
add a pre-2011 crosswalk row for area_code 276/277 pointing at the unified-Sudan polity that
covered those years.

**Acceptance criteria**
- None of the 9 codes resolve to `ROW-1850-2023` or to a post-partition polity for pre-partition
  years.
- Regression test pins the corrected mapping for all 9 codes.

**Depends on**: none. **Blocks**: improves the correctness of #460, #461, #424 but does not block them.

**Moves published values**: yes.

**Decision needed**: whether French Guiana (69) and New Caledonia (153) — French overseas
territories, not independent states — get their own standalone polity or fold into France's; the
other 5 have unambiguous real polities.

---

---

lbm364dl:
## What was wrong

`inst/extdata/regions.csv` is a static present-day grid of 209 area codes with no year column, and
nine of them resolved through `whep::polity_area_crosswalk` to the wrong polity:

- **Seven** (61 Eq. Guinea, 69 French Guiana, 153 New Caledonia, 154 N. Macedonia, 209 Eswatini,
  212 Syria, 299 Palestine) were forced onto `ROW-1850-2023` — "Rest of World", a non-territorial
  aggregate — for **every year**, because the crosswalk's `fabio_row_prefix` override put FABIO's
  rest-of-world bucket ahead of the area's own polity in the `mapping_prefix` coalesce.
- **Two** (276 Sudan, 277 South Sudan) resolved to the *post*-secession polities
  `SDN-2011-2025` / `SSD-2011-2025` for every year including pre-2011 ones, because their ISO3
  supplies only the successor prefix, so pre-2011 years missed the year-aware interval join and were
  rescued by its nearest-match fallback.

## Evidence it reproduced before the change

`add_polity_code()` on `origin/main`'s data, this branch, before any edit:

| area_code | 1990 | 2020 |
|---|---|---|
| 61, 69, 153, 154, 209, 212, 299 | `ROW-1850-2023` | `ROW-1850-2023` |
| 206 "Sudan (former)" | `SDN-2011-2025` | `SDN-2011-2025` |
| 276 Sudan | `SDN-2011-2025` | `SDN-2011-2025` |
| 277 South Sudan | `SSD-2011-2025` | `SSD-2011-2025` |

The new test file fails on the pre-change data with exactly those three failures (the fourth test,
which pins the two codes this PR deliberately leaves folded, passes either way by design).

## What I changed

Both edits are in `data-raw/table_mappings.R`; `data/polity_area_crosswalk.rda`,
`data/regions_full.rda` and `data/polities_cats.rda` are the regenerated outputs.

1. **The rest-of-world override yields to an area's own polity when the package models that area as
   a country.** `fabio_row_prefix` now applies only where the area is *not* both (a) in the compact
   country grid and (b) carrying a polity family the database holds. That un-folds **61, 153, 154,
   209, 212** onto `GNQ-*`, `NCL-1800-2025`, `MKD-1991-2025`, `SWZ-1894-2025`, `SYR-*`. The other
   areas FABIO folds keep folding — they are not individually modelled here, so routing them
   individually would diverge from FABIO's aggregation for nothing (#419 measures why the
   *numeric*-level version of that is unsafe: 13.7x on feed). `mapping_note` stops asserting a fold
   that no longer happened.

2. **The three Sudan areas get a second prefix.** `manual_area_prefixes` is already the mechanism for
   an area needing a prefix its ISO3 does not supply, and more than one row per area is how it says
   "more than one prefix" (the join fans out, each prefix contributes its own periods). Adding `SUD`
   to 206, 276 and 277 gives all three the unified-Sudan chain
   (`SUD-1899-1934`/`SUD-1934-1956`/`SUD-1956-2011`) for the years it covers, while `SDN`/`SSD` keep
   2011 onwards.

   **206 is included on purpose**, even though the issue names only 276/277: it is the area the
   builds actually key on. `.aggregate_to_polities()` renames `polity_area_code` onto `area_code`,
   and all three Sudan areas fold to bucket 206, so 206's crosswalk rows are what label the
   aggregated output. Fixing 276/277 without 206 would have been cosmetic.

This is a re-land of the `fabio_row_prefix` idea from #382 (which un-folded at the polity level and
left the numeric key alone), but **not** a port of its implementation: #382 gated the un-fold on
`observed_rows` from upstream's `label_alias_map.csv`, which belongs to #389 and is not available
here. The in-repo criterion — "the package's own country grid lists this area as a country" — selects
exactly the five codes this issue is about and leaves the eleven other folded areas that have
polities alone. Three of those eleven would be *actively wrong* against the shipped `polities`
vintage (Western Sahara and Palau have no geometry; the only Réunion polity ends in 1946), which is
why a blanket un-fold is not what landed.

## How I verified

- **Regeneration is honest and isolated.** `data-raw/table_mappings.R` reads the upstream GeoPackage,
  which has since moved from 603 to 740 polities. I pinned it to the vintage already shipped in
  `data/polities.rda` (round-tripped through a temp GeoPackage, verified column-for-column lossless),
  so `polities.rda` comes back **byte-identical** and the committed crosswalk stays consistent with
  the committed polities. Running the same procedure with the source edits reverted reproduces
  `main`'s crosswalk **exactly** (`all.equal` TRUE, 550 rows) — so the diff below is attributable to
  the two edits and nothing else.
- **The numeric aggregation key did not move.** `polity_area_code` is identical for **all 267** area
  codes, diffed area by area. The five un-folded areas keep bucket 999 and the Sudan areas keep
  bucket 206; only the polity *string* they are identified by changed.
- **Crosswalk delta:** +18 rows (the un-folded areas' own periods, plus the SUD chain on
  206/276/277), −5 rows (the ROW rows they replace). No new `NA` `polity_code` (351 China remains the
  only one). The only crosswalk row without geometry is still `CHL-1810-1883`
  (`polygon_status == "unassigned"`).
- **`regions_full` / `polities_cats`:** only `reporting_polity_code` and `reporting_polity_name`
  change, for 5 and 4 codes respectively, every one of them a polity with geometry (so
  `test_polity_output_coverage.R`'s geometry invariant still holds).
- **Tests:** the full local suite is green — `testthat::test_dir("tests/testthat")` reports
  **0 failures, 0 errors, 12 skips, 5006 passes**, including the heavy full-range CBS/production
  tests that do exercise areas 999 and 206. `air format` clean; the exact CI lintr config reports
  **zero lints**. No roxygen changes, so no `man/` churn.
  *Green tests are not a magnitude check* — a 266x inflation once passed 5151 of them in this repo —
  which is why the section below exists rather than a value-neutrality claim.

## Draft: this moves published values and I did not measure the magnitude

Opened as a draft. **A magnitude comparison against `main` has not been run and is required before
merge.** The `example = TRUE` paths are 10-row toy fixtures containing no area 999 or 206 rows at all
— I checked, and every row count, column sum and label compares bit-identical before and after,
which proves nothing about the areas that moved — and a full-range build is out of budget here.

Two specific mechanisms a magnitude check must cover, both found by reading rather than measuring:

1. **Bucket 999 goes from one polity name to six.** `.aggregate_to_polities()` groups by
   `polity_area_code` **and** `polity_name`, so where `main` emitted one `area_code == 999` row per
   (year, item, element), this emits up to six. `.select_best_source()`'s `dcast` already carries
   `fun.aggregate = sum` from #429, and its comment states that FABIO's rest-of-world *is* the sum of
   its members, so the bucket total should be preserved — but "should" is not "measured", and any
   consumer that keys on `area_code` without aggregating could now double-count.
2. **The free-text `area` label for bucket 999 may flip.** `.select_best_source()` picks it with
   `unique(dt_raw[, .(area_code, area)], by = "area_code")`, i.e. first-wins, so bucket 999 could
   come out labelled "Eswatini" instead of "Rest of World". The authoritative
   `reporting_polity_code` is unaffected — `.current_area_lookup()` still resolves area 999 to
   `ROW-1850-2023` and area 206 to `SDN-2011-2025`, both verified.

Bucket 206 moves the other way: pre-2011 it goes from two polity names to one, so it gains coherence
and loses duplicate keys. Both mechanisms are symptoms of the identity-vs-bucket mismatch that #414
and #419 own; neither is fixed here.

## Decision needed from the owner

**French Guiana (69) and Palestine (299) are not fixed**, and the new test pins them as still folding
into `ROW-1850-2023` so a future change trips it rather than moving them silently.

- Neither has a polity to route to in the `polities` vintage this package ships. Upstream
  whep-polities *now* carries `GUF-1816-1946`, `GUF-1946-2025` and `PSE-1948-2025`, but adopting them
  means refreshing `polities.rda` from 603 to 740 rows (+137/−7, including removals that re-map
  Canada, Cape Verde, Hungary, Pakistan, Panama, Suriname and South Africa). That is a separate
  change with its own unexplained magnitude shift — see #444 — and bundling it here would have made
  this PR unattributable.
- **French Guiana additionally needs a methodological decision I will not guess at**: as a French
  overseas *département*, does it get its own standalone polity or fold into France's? New Caledonia
  is the same class of question, but there the database already answers it — a dedicated
  `NCL-1800-2025` exists — so this PR follows the database. If the answer for French overseas
  territories is "fold into France", then `NCL-1800-2025` is the row that needs revisiting upstream,
  not this crosswalk.

Closes #459

Part of the polity migration epic #458. Related: #414 (bucket 206's post-2011 attribution), #419 (why
the numeric-level un-fold is unsafe), #444 (the unexplained shift on the closed #382 branch).
Improves the correctness of #460.


lbm364dl:
Brought up to date with `main` (99 commits), and reconciled with what `main` fixed in the meantime. **Three of your deliberate decisions were overtaken by `main` and I resolved them; the French Guiana one is the only one you may want to reverse.** Full suite: **5453 pass, 0 fail, 8 skip.**

## 1. Area 206 is redundant now — dropped

You included 206 on purpose, arguing it is "the area the builds actually key on". That was right then. `main` has since adopted upstream's published FAOSTAT map (`f5a2037e`), and measured on clean `main` with no PR applied:

| area | 1961 | 2000 | 2020 |
|---|---|---|---|
| 206 | `SUD-1956-2011` | `SUD-1956-2011` | `SUD-1956-2011` |
| 276 | `SDN-2011-2025` | `SDN-2011-2025` | `SDN-2011-2025` |
| 277 | `SSD-2011-2025` | `SSD-2011-2025` | `SSD-2011-2025` |

206 is already fixed upstream-side. **276 and 277 are not** — their pre-2011 years still resolve to states that did not exist, which is exactly your defect. So they keep the `SUD` prefix and 206 no longer needs an entry.

## 2. The successor prefixes duplicated the map's own rows — dropped

You added `SDN` for 276 and `SSD` for 277 alongside `SUD` because the old build let `manual_polity_prefix` win the coalesce outright, so `SUD` alone would have lost 2012 onwards. The map now declares those spans itself, so repeating them produced **4 duplicate `(area, polity, span)` rows** — which surfaced as `-50` and `-12` year gaps in `main`'s span-partition assertion. With only `SUD` added, gaps are uniformly `1` and there are no duplicates beyond the 4 `area_code`-NA crown-dependency rows `main` already carries.

## 3. Your tripwire fired — French Guiana and Palestine now un-fold

This is the one to look at. Your test said:

> A future refresh of `polities` that adds GUF/PSE rows will trip this test. That is the point -- it forces the decision instead of silently changing the attribution.

It fired. `main`'s 740-row vintage carries `GUF-1816-1946`, `GUF-1946-2025` and `PSE-1948-2025`. I resolved it by **applying your own rule rather than adding an exception** — an area in the compact grid carrying a polity family the database holds takes that polity, which is precisely how you handled New Caledonia ("there the database already answers it ... so this PR follows the database"). Upstream has now answered French Guiana the same way.

**What made me willing to decide it rather than escalate:** it moves no number. `polity_area_code` stays `999` for all seven areas, so only the identity label changes.

| | before | after |
|---|---|---|
| crosswalk rows | 598 | 609 |
| `fabio_row_fold` areas | 62 | **54** |
| `upstream_map` rows | 245 | 254 |
| un-folded | — | 61, 69, 153, 154, 209, 212, 299 |
| `polity_area_code` for those 7 | 999 | **999 (unchanged)** |

If you want French Guiana folded into France's polity, that is an upstream change to `GUF-*`, not an exception here — I have said so in the test comment.

## Three pins retargeted, not deleted

`main` added `test_polity_faostat_map.R`, whose fold test asserts the opposite of this PR. Its comment is explicit that it is pinned "because lifting it moves every Rest-of-World figure ... not because it is right", and that "this number must move only on purpose". This PR is that purpose — and the pin's stated reason is **measurably false for a polity-level un-fold**, since the numeric key is untouched. So the tripwire now asserts the invariant that actually matters (`polity_area_code == 999` for all seven) rather than the identity that was standing in for it. Same for `test_polity_grid_area_codes.R` and `test_datasets.R`'s `fabio_row_sources` block. Lifting the fold on the **numeric** key — 13.7x on feed, #419 — is still not done here.

## One thing to know about the regenerated data

I regenerated against upstream revision **`eadf1db`**, not upstream `HEAD`. `eadf1db` is the vintage `main`'s committed crosswalk reproduces exactly, so this PR's data delta is this change alone. Upstream `HEAD` is at 749 rows and regenerating against it would silently fold the pending re-sync (#530) into this PR and make both unattributable. **After #530 lands, this branch should be regenerated on top of it.**

lbm364dl:
**Correction to my comment above, and to two test comments now fixed on the branch.**

I justified leaving the numeric fold alone by citing #419's 13.7x inflation of global feed. **That figure has been disproved.** PR #555 ran the measurement #419 never had — two complete `get_wide_cbs()` builds over 1850-2023, current `main` against all 61 Rest-of-World members promoted:

| column | main | promoted | ratio |
|---|---|---|---|
| feed | 9.050144e11 | 9.049817e11 | **1.0000** |
| production | 3.007430e12 | 3.005864e12 | 0.9995 |
| processing | 1.767704e12 | 1.766518e12 | 0.9993 |
| food | 3.143402e11 | 3.143159e11 | 0.9999 |
| stock_addition | 3.935775e10 | 3.887598e10 | 0.9878 |

Largest move of any column: **1.2%**, not 13.7x. The original figure was measured against a baseline corrupted by the `dcast()` duplicate-key `length()` fallback (#425, fixed by #429) — whose own code comment records the duplicates as being "in areas 206 and 999", so the 13.7x came from comparing a promoted build against a corrupted one.

**Nothing about this PR changes.** It still moves only the identity, `polity_area_code` still stays `999` for all seven areas, and the un-fold is still value-neutral. What changes is the *stated reason* the numeric fold stands: it is the modelling question it always was — FABIO matrix comparability, and the promotion's effect past CBS, which #555 explicitly did not measure — rather than a magnitude argument that does not survive.

Pushed as `58a2ac28`, comments only, 735 pass / 0 fail on the three affected test files.

lbm364dl:
Reverted from \`main\` in #561, and re-opened as a design question in **#563**. Apologies — I merged this and the revert is on me.

**My value-neutrality argument was correct but measured the wrong invariant.** I verified `polity_area_code` is unchanged for all 266 areas, no area enters or leaves a bucket, and bucket 999 keeps its members. All true. But the numeric bucket is what values are summed *into*; `area` — the polity name — is what `.aggregate_to_polities()` groups *by*. Un-folding the identity while leaving the bucket splits the group without moving a number, so every check I ran was blind to it:

| with the fold | with this PR |
|---|---|
| `999  Rest of World  340` | `999  Syrian Arab Republic  300` |
| | `999  Rest of World       40` |

Mass conserves, 347 in and 347 out. But bucket 999 stops summing, and one `area_code` now carries two `area` labels — which is exactly the shape #382 measured and withdrew ("four inner joins use `c("area", "area_code")`", 702,166 rows dropped), and duplicate `(area_code, year, item)` keys are what fed #425's `dcast()` `length()` fallback.

Caught by #555's `.aggregate_to_polities() sums a fold and reports it`, which failed only once this landed.

**The substance of this PR is still right** and the branch keeps all of it — the un-fold, the regenerated data on the `eb02dcb` vintage, the Sudan minimal fix, the reconciled test pins. What it needs first is for the aggregator to stop keying on the member's polity name; #563 lays out the three options and the cost of each. #459 stays open.

Issue #407, opened 2026-07-30
LUH2 land for six crown dependencies and overseas territories is dropped, though the crosswalk knows their sovereign


lbm364dl:
Found on #382 by quantifying a warning that fires on every production build. The message is fixed there; this is the modelling decision I did not make.

## What the warning was hiding

`.read_land_areas()` warned:

```
LUH2 ISO3 codes not found in polity_area_crosswalk, dropping: JEY, GGY, IMN, -99, BLM, ALA, and SXM
```

One message, two completely different facts, in wildly different proportions:

| dropped | area | share of all LUH2 area |
|---|---|---|
| `-99` — LUH2's own unassigned marker | 17,876 Mha | 0.391% |
| six real territories | 35 Mha | 0.0008% |

The sentinel is **510×** the territories. So almost everything the warning appeared to be losing is land LUH2 itself does not attribute to any country — not a coverage gap here. Reporting them together made a source property read as a polity-mapping failure. #382 now reports the sentinel as informational and the territories as a warning.

## The part that is a real (small) gap

"Not found in polity_area_crosswalk" was **wrong** about the six. They are all there, each with its sovereign's polity:

| ISO3 | territory | polity in the crosswalk |
|---|---|---|
| JEY | Jersey | `GBR-1921-2025` |
| GGY | Guernsey | `GBR-1921-2025` |
| IMN | Isle of Man | `GBR-1921-2025` |
| ALA | Åland | Finland |
| BLM | Saint Barthélemy | France |
| SXM | Sint Maarten | Netherlands |

What they lack is a FAOSTAT `area_code`, and the bridge is built with `include_unmapped = FALSE`, which drops exactly those rows. So the mapping exists; the aggregation key does not.

## The decision

Their land could be attributed to the sovereign polity the crosswalk already names. **I have not done it, because it changes what an existing series means:** LUH2 reports `GBR` separately, so folding Jersey, Guernsey and the Isle of Man into it makes `GBR` mean "UK including crown dependencies" where it previously meant the UK proper. Same for Finland/Åland, France/Saint Barthélemy, Netherlands/Sint Maarten.

That is a modelling call, and at 35 Mha (0.0008% of LUH2 area) it is unlikely to be worth a change in series definition — but it should be a decision rather than an accident of `include_unmapped = FALSE`.

Three options, in my order of preference:

1. **Leave dropped, now that the warning says why.** The magnitude does not justify redefining `GBR`.
[...]

PR #481, opened 2026-08-04
Port core-production-cbs's #382 fixes: stock-share territory key, gdp-pop labels, residue and LUH2 diagnostics

### LINKED ISSUE #470: Port core-production-cbs's 5 small #382 fixes
*Merges: core-production-cbs F1 (fodder), F2 (gdp-pop), F3 (crop-residue warning), F4
(stock-share territory key), F5 (LUH2 microterritory warning split) — and F5 is a **duplicate** of
luh2 slice's F3, both describing the identical `.read_land_areas()` warning-split fix.*

**Problem** (5 independent small bugs, all with fixes already written on #382)
1. **Fodder** — `.read_fodder_euadb()` joins on an ISO3/prefix string (`regions_full$polity_code`)
   rather than a numeric code; Czechoslovakia's canonical ISO3 is `F51` not `CSK`, so the historical
   prefix silently fails to join. ~13% of fodder rows dropped (`R/build_production.R:595-620`).
2. **GDP/population proxies** — `.fill_with_proxies()` joins by raw area *name*; 11 of 196
   countries' spellings diverge from the crosswalk's canonical name (Bolivia, Taiwan, DR Congo,
   Iran, Ivory Coast, Lao, North Korea, Syria, Tanzania, Turkey, Venezuela), leaving 1,892 rows
   (6.4%) silently unfilled (`R/build_cbs.R:653-659, 2067`).
3. **Crop residues** — `get_primary_residues()` resolves area by exact name match via
   `add_area_code()`, not `resolve_polity_label()`; unresolved rows travel silently with NA polity
   columns all the way to `build_supply_use()`'s output (`R/production.R:92-119`,
   `R/code_names.R:77-90, 244-260`).
4. **Livestock stock-share splitting** — `.compute_stock_shares()` groups by `(year, area_code,
   Item_Code)` only; 16 distinct reporting territories share legacy aggregate buckets (e.g. a
   rest-of-world "999" bucket), so a territory's split share is computed against a mixture of
   unrelated countries' stock counts. Measured example: Eswatini's broiler share computed as 0.9419
   instead of the correct 0.0027 (`R/build_production.R:1224-1269`).
5. **LUH2 microterritories** — `.read_land_areas()` emits one undifferentiated warning conflating
   LUH2's own `-99` unattributed-land sentinel (17,876 Mha, a source artefact) with 6 real
   territories that DO have a sovereign polity but no FAOSTAT area_code (Jersey, Guernsey, Isle of
   Man, Saint-Barthelemy, Aland, Sint Maarten — 35 Mha total) (`R/build_production.R:413-432`). This
   is the identical finding reported independently by the luh2-slice audit.

**Evidence**: file:line citations above; row-count/share figures are #382's own measurements
(comment text in the branch diff), consistent with the described mechanisms.

**Target state / Acceptance criteria**: port each fix as-is —
1. Numeric-code join in `.read_fodder_euadb()`. 2. `.canonicalise_gdp_pop_area()` +
call site. 3. `.warn_residues_no_area()` (diagnostic only — the underlying name-based join itself is
NOT fixed by #382 either; that's a follow-on once the Tanganyika/Tanzania 1961-1964 crosswalk gap is
resolved upstream in whep-polities). 4. `.compute_stock_shares()`/`.carry_forward_shares()` gain
`area` to their grouping key; regression test locks the Eswatini example. 5. Split the warning into
a `cli::cli_inform()` for the `-99` sentinel and a `cli::cli_warn()` naming the 6 territories.

**Depends on**: none. **Decision needed**: (3) whether a Tanganyika-era alias is worth adding given
FAOSTAT itself only reports the post-1964 union name for that period; (5) whether to actually
attribute the 6 territories' land to their sovereign polity is explicitly out of scope here (open
as whep#407 — check before filing a duplicate).

**Moves published values**: yes for (1),(2),(4); no for (3),(5) (diagnostics only).

**Port from #382**: yes, all 5, fully or diagnostically written — `R/build_production.R`,
`R/build_cbs.R`, `R/code_names.R` call sites, `tests/testthat/test_stock_share_territory_key.R`.

---

---

lbm364dl:
Part of the polity migration epic #458. Closes #470.

**Draft, because two of the four changes move published values and a magnitude
comparison against `main` has not been run.** The stock-share fix changes how
livestock slaughter counts are split between broilers/layers and dairy/non-dairy
in FAOSTAT bucket 206, and the gdp/population fix lets the pre-1962 historical CBS
proxy fill reach areas it previously missed. Both need a total-by-column diff
against `main` before merge. I did not run one: the instruction for this task was
not to run a full-range build, and neither change can be measured any other way.

## What the issue asked for, and what actually reproduced

The issue lists five #382 fixes. I reproduced each on this branch first.

| # | Claim | Reproduced on main? |
|---|---|---|
| 1 | fodder joins on an ISO3/prefix string, ~13% of rows dropped | **No** |
| 2 | gdp/population proxies join by name, 11 names diverge, 1,892 rows unfilled | Yes, but the fix in #382 is wrong here -- see below |
| 3 | crop residues resolve by exact name, unresolved rows travel with NA polity columns | Yes |
| 4 | stock shares grouped by bucket, not territory | Yes, in a different form |
| 5 | LUH2 emits one warning conflating `-99` with six real territories | Yes |

### (1) fodder -- not ported, does not reproduce

`whep::regions_full$polity_code` on main is still ISO3 and Czechoslovakia's value
is still `CSK`, which is also its `area_iso3c` in the crosswalk. Measured: all 26
ADB_Region rows bridge to an area code, 0 unjoined. The `F51` rename that broke
this on #382 is not in main's data.

The second half of that hunk (`polity_name` vs `FAOSTAT_name` for the `area`
label) is already handled differently and more generally on main:
`.merge_euadb_fodder()` rekeys EU AgriDB rows onto the FAO name for the same
`(year, area_code)`, which also covers the periodized names `polity_name` produces.
So I changed nothing here.

### (2) gdp/population -- reproduced, but re-implemented

The defect is real: `.fill_with_proxies()` joins population on `c("year", "area")`,
a name. But #382 canonicalises the pin towards the crosswalk's `area_name`, and
that is not the vocabulary the frame being filled carries. Everything that comes
through `.aggregate_to_polities()` carries the period-specific `polity_name`.
Measured on main:

- 35 of the pin's 196 area names (5,346 rows, 18.0%) are not names any builder
  emits -- not 11/1,892, which is the count against `area_name`;
- 5 of the names #382 would have rewritten (Bolivia, Iran, Tanzania, Venezuela,
  North Korea) **match today** as `polity_name` and would have stopped matching.

So `.canonicalise_gdp_pop_area()` here maps the pin's own ISO3 `area_code` to the
polity name for the row's year, using `.iso3_to_fao_area_code()` and
`.add_polity_columns_dt()` -- the same two helpers `.aggregate_to_polities()` uses.
The two vocabularies then agree by construction, including the pre-1962 back-cast:
1900 Bangladesh resolves to "East Pakistan (1947-1971)" on both sides because both
anchor at 1961. It is a code lookup, not a synonym list, which is the point of the
epic.

Two guards, both measured:

- it only renames a label that differs from its own polity name, so a working
  label cannot be rewritten. Checked against the 1961 CBS extract's own area
  vocabulary: **110 of 169 labels matched before, 148 after, 0 went from matching
  to not matching**;
- it only renames an area that IS its polity's canonical reporting area
  (`area_code == polity_area_code`). Syria and Equatorial Guinea fold into Rest of
  World, so relabelling them would attribute one member's population to the whole
  bucket and put several pin rows on one `(year, area)` key. Whether a bucket's
  proxy should be the SUM over its members is a real question and not one a
  relabelling can answer, so those 6 areas stay unfilled. **0 duplicate
  `(year, area)` rows are created.**

### (3) crop residues -- ported as the diagnostic it is

Measured on main: 44,985 of 475,688 rows (9.5%) come out of `add_area_code()` with
no `area_code`, over 14 labels and years 1961-2021, and 3,937 rows of
`get_primary_residues()`'s output carry NA polity columns as a result. Every one of
the 14 is a short form of an area the crosswalk holds under a FAOSTAT long form.
`.warn_residues_no_area()` names them; it does not drop them, because whether an
unattributable residue row should be removed is a modelling question.

`.read_crop_residues()` also gains the `.warn_unmapped_codes()` call the two trade
drop sites already have, at the point the 3,937 rows actually disappear. Today it
always reports the single code `NA`; the comment says so, and it stays as the
tripwire for a code that is real but unmapped.

(The issue's "200 rows for Tanzania 1961-1964" is #382's *post*-fix number, after
its `.fill_area_code_via_polities()` fallback in `R/code_names.R`. That fallback is
not in this PR and is not in this issue's acceptance criteria, so the numbers here
are main's.)

### (4) stock shares -- reproduced, different example

The issue's Eswatini/999 example does not exist on main: `.aggregate_to_polities()`
sums every ROW-folded area into a single "Rest of World" row, so bucket 999 holds
one territory. The defect is live for a different bucket:

**FAOSTAT bucket 206 holds both Sudan and South Sudan in the same years.**
Measured on the full `faostat-emissions-livestock` pin:

- 48 distinct `(year, area_code, Item_Code, item_cbs_code)` keys carry **two**
  share rows -- all in bucket 206, all 2012-2023 -- so which one
  `.split_slaughter_by_shares()`'s `inner_join` used was a matter of row order;
- 2015, item 1057 (chickens):

  | area | share before | share after |
  |---|---|---|
  | Sudan, broilers | 0.6071 | 0.8000 |
  | Sudan, layers | 0.1518 | 0.2000 |
  | South Sudan, broilers | 0.1659 | 0.6880 |
  | South Sudan, layers | 0.0752 | 0.3120 |

- the only symptom was `fill_linear`'s "Duplicate year values found within groups.
  2 group/time combination(s)", which reads like ordinary source noise.

After adding `area` to the grouping, the carry-forward key and the join key: 0
duplicate keys, and **all 40,824 share groups sum to 1** (before, no group summing
to 1 meant what it appeared to mean, because groups spanned territories).

I checked the tightened `inner_join` cannot now drop slaughter rows: only bucket
206 has more than one `polity_name` for any year across the whole crosswalk, both
pins derive "Sudan"/"South Sudan" from the same area codes, and after
`.carry_forward_shares()` both territories have shares for every year 1961-2023.

I did not port #382's `intersect()`-derived dynamic key. The one caller always
supplies `area`, so the column is a contract rather than an option, and a branch
that cannot run in production hides that. The existing `.carry_forward_shares` test
was updated to carry `area`.

### (5) LUH2 micro-territories -- ported, message split only

Before: `LUH2 ISO3 codes not found in polity_area_crosswalk, dropping: BLM, ALA,
-99, SXM, JEY, GGY, and IMN`.

Measured over the whole pin, 1850-2022:

| | Mha | share of LUH2 area |
|---|---|---|
| `-99`, LUH2's own unassigned marker | 8,620 | 0.358% |
| the six real territories | 19 | 0.0008% |

459 parts in 460 of the apparent loss is land the source itself attributes to no
country. And the six -- Jersey, Guernsey, Isle of Man, Saint-Barthelemy, Aland,
Sint Maarten -- ARE in the crosswalk, each under its sovereign's polity; what they
lack is a FAOSTAT area code, so the old wording was false for them.

Split into a `cli_inform()` for the sentinel and a `cli_warn()` naming the six.
**Drop behaviour is unchanged** -- attributing their land to the sovereign changes
what GBR means and is #407.

## How I verified

Each new test was run against unmodified sources and fails there:

```
test_stock_share_territory_key.R   5 failures  (incl. "object 'area' not found",
                                                "Duplicate year values found within groups")
test_build_production.R (LUH2)     4 failures  ("LUH2 ISO3 codes not found in
                                                polity_area_crosswalk, dropping: -99 and JEY")
test_build_cbs.R (gdp-pop)         2 errors    (".canonicalise_gdp_pop_area not found")
test_production.R (residues)       2 errors    (".warn_residues_no_area not found")
```

and passes with the change. End-to-end reruns of the four defects on real pins are
in the commit message. Gates:

- `air format` on all seven changed files -- clean;
- `lintr::lint_package()` with the exact CI configuration -- **no lints found**;
- `test_build_cbs.R`, `test_build_production.R`, `test_production.R`,
  `test_stock_share_territory_key.R`, `test_luh2_landuse.R`,
  `test_read_raw_inputs.R` -- all pass;
- every added line is ASCII (checked mechanically against `HEAD`);
- no roxygen blocks touched, so `man/` is unchanged. `.warn_residues_no_area()` is
  deliberately defined *after* `get_primary_residues()`; #382 put it between that
  function's roxygen block and the function, which would have reassigned the docs.

## Noticed but deliberately not fixed

`.fill_with_proxies()` joins its *land* proxy on `c("year", "area")` against
`land_wide`, whose `area` is the crosswalk's `area_name` -- a third vocabulary
again, differing from `polity_name` for 99 of 265 lookup rows. So the `agriland`
proxy has the same class of defect this PR fixes for `pop`. It is not in #470's
scope and fixing it moves more values, so it wants its own issue.


lbm364dl:
Ran the total-by-column diff against `main` that you said both value-moving changes needed. It is far smaller than when you opened this, and the reason is instructive.

Two full-range `get_wide_cbs()` builds:

| | |
|---|---|
| rows | 2,039,344 → 2,039,344, **+0** |
| columns that move | **3 of 11** |
| supply-use imbalances | 48,235 → 48,235, **identical** |

| column | ratio |
|---|---|
| production | 1.00001905 |
| domestic_supply | 1.00001799 |
| processing | 1.00003240 |
| the other eight | **1.00000000** |

So roughly **+0.002%**, and only on three columns.

**Why so small: #511 landed first and absorbed most of it.** Your gdp/pop concern and #511's proxy fix are the same defect approached from two sides — you canonicalise the pin's labels toward the polity name, #511 stopped joining on names at all and keyed the proxies on `polity_code`. With that merged, `.canonicalise_gdp_pop_area()` has almost nothing left to correct. What remains is your stock-share territory-key fix, which is the ~0.002% on production/processing.

That also vindicates the call you made in the PR: you re-implemented the gdp/pop fix rather than porting #382's version, on the grounds that #382 canonicalised toward a third vocabulary and would have *broken* five labels that matched. Had it been ported as-is it would now be fighting #511.

The imbalance count being bit-identical is the check I care most about after today — #382 conserved mass and still dropped 702,166 rows through a join-key change, so equal totals are not sufficient evidence on their own.

Full suite 5740 pass, 0 fail. Your assessment that claim (1), the fodder join, does not reproduce still holds on today's main.

Issue #59, opened 2025-10-15
Avoid CRAN check fails when running examples


lbm364dl:
> "Packages which use Internet resources should fail gracefully with an informative message if the resource is not available or has changed (and not give a check warning nor error)."

Let's just have the examples locally.

PR #485, opened 2026-08-04
Refresh the stale polities snapshot from upstream (603 -> 740)

lbm364dl:
Part of the polity migration epic #458. **Draft** — see "Blocked on upstream" below.

## What was wrong

`data/polities.rda` was a **stale snapshot**: the package shipped **603** polities while upstream `whep-polities` ships **740**. Nothing detected it, because nothing compared the two.

The consumer manifest published by whep-polities#39 (merged today) makes it visible for the first time — `counts.total = 740`. This is exactly the drift that manifest exists to catch, found on its first use.

**It is not a deliberate filter.** `data-raw/table_mappings.R:12` reads the GeoPackage and writes at line 271 with no filtering, so the 137 absent polities are simply un-regenerated. They span every type, including **78 typed `national`**.

## What prompted it

Of the 869 aliases in the newly published `label_alias_map.csv`, **115 point at polities the package did not ship** — `ETH-1907-1936`, `ALK-1867-1959`, `AOI-1936-1941`, `ASM-1900-2025` and others. So `resolve_polity_label()` (whep#389, PR #479) could return a code a caller then cannot find in `whep::polities`. All 869 resolve cleanly against upstream's 740-row database, so the dangling references come entirely from the stale copy.

## What changed

Both `data-raw` scripts were re-run, because `polity_area_crosswalk` feeds `regions_full` and `polities_cats`; regenerating only the first would leave those disagreeing with it.

| dataset | before | after |
|---|---|---|
| `polities` | 603 rows | **740** |
| `polity_area_crosswalk` | 550 rows | **621** (462 → 528 distinct polity codes) |
| `regions_full` | 272 rows | 272; `reporting_polity_code` changed on 11 |
| `polities_cats` | 198 rows | 198; `reporting_polity_code` changed on 11 |

Eight further `data/*.rda` files were rewritten with byte differences but **identical content** (`animals_codes`, `cb_processing`, `crops_eurostat`, `items_full`, `items_cbs`, `items_prod`, `liv_lu_coefs`, `primary_double`) and have been reverted, so this diff carries only real changes.

## This moves published values, and it is most of whep#444

The crosswalk reaching **621 rows is the same change** I isolated in #444 as roughly **75% of that PR's value movement** — `feed` +5.3% of the data effect, `production` +1.5%. So a large part of what #444 records as *unexplained* is this snapshot catching up with upstream, not something #382 introduced.

**A magnitude comparison has not been run here.** It should be, before merge.

## Blocked on upstream: the refresh surfaces three real data defects

Eight tests fail. None of them is the test pinning stale state — each is a genuine ambiguity the 603-row snapshot happened to hide.

**1. Overlapping polity periods make `add_polity_code()` ambiguous** (`test_polities.R:14,15,31,32`). Two areas now have periods covering the same year, so which polity a year maps to is a tie-break rather than an answer:

```
area  15  BLX-1850-1999  and  BLX-1921-1999      -> both cover 1921-1998
area 248  F248-1920-1947, F248-1920-1991,
          F248-1947-1991                          -> three-way overlap
```

The tests expected `BLX-1850-1999` and `F248-1920-1991`; resolution now returns `BLX-1921-1999` and `F248-1947-1991`. Neither answer is wrong — the data admits both.

**2. Fourteen polities have no geometry yet declare a `polygon_status`** (`test_polity_output_coverage.R:179`), which the test reads as a contradiction — claiming a polygon status while carrying no polygon:

```
assigned : ALK-1867-1959 (Territory of Alaska), GCT-1919-1956 (Gold Coast and British Togoland)
proxy    : CAN-1800-1866, IDN-BLB-1949-1951, IDN-JVM-1949-1951, IDN-OTH-1949-1951,
           RYU-1937-1945, RYU-1945-1972
estimate : CAN-1866-1870, FRIN-1816-1954, PTIND-1816-1961, SER-1918-1945
```

**3. One crosswalk row carries a NULL `polity_code` with a NULL `polygon_status`.**

**4. Coded rows now reference geometry-less polities** (`test_polity_output_coverage.R:186`), which follows from 2 — 44 of the 740 have no geometry, and the stale snapshot excluded some of them.

**Correction (see comment below):** defects 1 and 3 are NOT upstream -- they are whep failing to filter `wiki_status`, tracked as **#486**. `F248-1920-1991` is already marked `retired` upstream and should never have been a resolution candidate; the BLX pair is a live aggregate-vs-national distinction needing a `polity_type` preference. Only defect 2 is upstream (lbm364dl/whep-polities#59). This PR is blocked on **#486** and lbm364dl/whep-polities#59.

## Why not just update the tests

Because they are right. `add_polity_code()` returning either of two overlapping polities is not a test problem, and a polity with `polygon_status = "assigned"` and no polygon is not either. Loosening these assertions would convert a detected upstream defect into a silent one.

Closes nothing yet. Depends on the upstream fixes and on a magnitude measurement.


lbm364dl:
## Correction: the test failures are a whep bug, not upstream data defects

I filed the overlapping-period failures as an upstream defect (lbm364dl/whep-polities#58). **That was wrong** — pushback prompted a proper check and the database is right in both cases. #58 is retracted and closed; the real bug is **#486**, here in whep.

**`F248-1920-1991` is already `wiki_status = "retired"`.** The coarse Yugoslav period was superseded by the `1920-1947` / `1947-1991` split, and upstream marked it. whep's crosswalk build carries `wiki_status` through (`data-raw/table_mappings.R:29`, `:176`) and never filters on it, so a retired polity stays a live resolution candidate. Measured: whep ships 8 `retired` + 19 `superseded`, and **22 of those 27 appear in `polity_area_crosswalk`** — so this is 22 latent collisions, not the 2 the failures happened to expose.

The newly returned answers are the *correct* ones. `test_polities.R:14` expected `F248-1920-1991`; it now gets `F248-1947-1991`, which is right. The test was passing for the wrong reason, because the stale 603-row snapshot did not contain both members of the pair.

**`BLX` is a different question and also not upstream's fault.** Both rows are live and both are correct:

```
BLX-1850-1999   Belgium-Luxembourg                 aggregate   iso3 BLX
BLX-1921-1999   Belgium-Luxembourg Economic Union   national    iso3 BEL
```

The FAOSTAT reporting aggregate spanning 1850-1999, and the actual union founded 1921. For area 15, named "Belgium-Luxembourg", the aggregate is the right target and `test_polities.R:15` expected exactly that. Filtering dead polities will not fix it — resolution needs a `polity_type` preference. Also covered in #486.

## What this changes for this PR

Blocked on **#486**, not on upstream. lbm364dl/whep-polities#59 (14 polities declaring a `polygon_status` with no geometry) still stands as a genuine upstream defect and still blocks the geometry failures.

Retracting one further thing I wrote above: I said "defects 1-3 belong upstream". Only #59 does. Updating the PR body accordingly.

lbm364dl:
## Superseded by #527 — closing

This PR existed to refresh the stale embedded snapshot from 603 polities to 740. **That has landed**, as a side effect of #527 taking the FAOSTAT area map as the resolution authority: implementing it required regenerating `data/`, which brought the snapshot forward at the same time.

Verified on current `main`:

```
whep::polities              740 rows   (was 603)
whep::polity_area_crosswalk 598 rows   (was 550)
```

Regenerating on top of main now produces a **byte-identical crosswalk** and no change to any `data/*.rda` except geometry — so there is nothing left in this branch to contribute.

## The ten failures this PR carried are all resolved, and none needed the geometry work upstream

Re-running them on main: **no failures.** For the record, how each closed:

- **`test_polities.R:14`** expected `F248-1920-1991`, a *retired* polity. #499's dead-polity filter plus #527's map (which routes area 248 via `manual-replace` to `F248-1947-1991`) fixed it. The test was pinning the wrong answer.
- **`test_polities.R:15`** expected `BLX-1850-1999` and was **right** — the published map confirms `iso-equal` to that row. The prefix join was returning `BLX-1921-1999`; #527 fixed the code, not the test.
- **`test_polities.R:31,32`** followed from 15: `BLX-1921-1999` is `national`, so preferring it also disabled the "aggregates are not extended beyond their range" rule.
- **`test_datasets.R:93,94,100`** were Djibouti. Area 72 now resolves to `FRS-1884-1977` / `FRS-1977-2025`, which the map has always carried and the prefix join could never reach from `DJI`.
- **`test_polity_output_coverage.R:173,179`** were the geometry assertions I had expected to need upstream builders. They pass on main.
- **`test_build_cbs.R:625`** expected `IND-1800-1893`, which is `superseded` — replaced by the finer `IND-1800-1886` / `IND-1886-1893` split. Corrected in #527.

Three of those tests were pinning answers that were **wrong**, which is the part worth remembering: the refresh did not break them, it exposed them.

## One small thing left, not worth a PR of its own

`polities.rda`'s geometry column now differs from upstream, because lbm364dl/whep-polities#67 attached polygons for `ALK-1867-1959` (Alaska) and both `RYU` rows (Okinawa, after fixing a feature id that pointed at Shizuoka). Any future `data-raw` regeneration picks those up; nothing depends on them today.

Also superseded: whep#444's attribution. Its "75% of the movement" figure was measured against a 621-row crosswalk. Main now ships 598 — the dead-polity filter removed 23 of those rows — so that number does not carry over and would need re-measuring against the current state.

PR #506, opened 2026-08-04
Make the LUH2 dependency-land drop a decision instead of a side effect

### LINKED ISSUE #407: LUH2 land for six crown dependencies and overseas territories is dropped, though the crosswalk knows their sovereign

lbm364dl:

## What was wrong

`.read_land_areas()` builds its ISO3 -> polity bridge from
`.current_area_lookup(include_unmapped = FALSE)`, and that function keeps only
crosswalk rows with a FAOSTAT `area_code`:

```r
out <- out[!is.na(area_code)]
```

Six territories are in the crosswalk *with* their sovereign's `polity_code` but
with no `area_code` of their own, so the bridge cannot see them and LUH2 drops
their land. The identity was available; the numeric aggregation bucket was not,
and the bucket is what the bridge is keyed by. (The issue attributes the drop to
`include_unmapped = FALSE`; the actual filter is `!is.na(area_code)` -- the six
have `mapping_status == "matched"`, so `include_unmapped` never touches them.)

## Evidence it reproduced before the change

On `polity/407` before any edit, `.read_land_areas(years = 2000L)`:

```
WARN: LUH2 ISO3 codes not found in polity_area_crosswalk, dropping: BLM, ALA,
-99, SXM, JEY, GGY, and IMN
raw total Mha:  13589.39
kept total Mha: 13546.49
six territories in raw: 0.1085224 Mha   (sentinel -99: 42.78987 Mha)
six present in output? FALSE
```

And the mapping the warning denied exists:

| ISO3 | crosswalk `polity_code` | `area_code` | sovereign bucket |
|---|---|---|---|
| JEY, GGY, IMN | `GBR-1921-2025` | `NA` | 229 |
| ALA | `FIN-1940-2025` | `NA` | 67 |
| BLM | `FRA-1919-2025` | `NA` | 68 |
| SXM | `NLD-1830-2025` | `NA` | 150 |

Full range 1850-2022: the six are 18.77 Mha-years of 2,409,085, i.e. 0.0008% of
LUH2 land, matching the figure in the issue.

## What changed

- `.dependency_sovereign_iso3()` (R/polities.R): for each crosswalk ISO3 that has
  a polity but no `area_code`, returns the ISO3 of a territory that shares its
  polity and does have a bucket -- its sovereign. Deterministic on both sides
  (most recent polity period per ISO3; lowest `area_code` where one polity spans
  several buckets, as ROW and the Sudan/Ethiopia splits do).
- `.read_land_areas(dependency_land = c("drop", "sovereign"))`: `"drop"` is
  exactly today's behaviour, `"sovereign"` relabels the dependency rows with the
  sovereign's ISO3 before the bridge merge, so they pick up the sovereign's
  bucket *and* label, and sums them into the sovereign's rows.

Rows are summed rather than stacked deliberately: `.fix_luh2_crop_collapse()`
assigns an interpolated value into every row matching an
`(area_code, year, Land_Use)` key, and it fires 890 times over the full range,
so a second row under one bucket would be repaired to twice the intended area.

## How I verified

**Default path moves no values.** I rebuilt the pre-change `.read_land_areas()`
from `origin/main`, evaluated it against the whep namespace, and compared on
1995-2005:

```
all.equal(old, new default): TRUE      identical: TRUE
nrow old/new:      30888 30888
sum Mha old/new:   149035.9928871944 149035.9928871944
sum C   old/new:   4716336.014268878 4716336.014268878
```

(Row order out of the threaded parquet read is not stable between calls on
`origin/main` either; the comparison also holds under a canonical sort.)

**Attribution is additive and lands only where it should.** Same years, with
`dependency_land = "sovereign"`: the warning drops to `-99` alone, row count is
unchanged at 30,888, total area rises by 1.193746 Mha, exactly the six
territories' raw LUH2 area, and only three buckets move:

| bucket | off (Mha) | on (Mha) | delta | dependencies' raw area |
|---|---|---|---|---|
| 67 Finland | 326.22357 | 326.62164 | +0.3980749 | 0.3980749 |
| 150 Netherlands | 36.14742 | 36.32287 | +0.1754446 | 0.1754446 |
| 229 United Kingdom | 245.38887 | 246.00909 | +0.6202265 | 0.6202265 |

FRA does not move: Saint Barthelemy has no LUH2 land. Buckets outside those
three are row-for-row identical (max abs diff 0), and no bucket gains a
duplicate `(area_code, area, year, Land_Use)` row (132 such rows before and
after, all pre-existing AUS grid fragments).

## Tests

New `tests/testthat/test_luh2_dependency_land.R`, 21 assertions. Checked against
`origin/main`'s `R/` with the new test file in place: 4 of the 4 test blocks
error out (`object '.dependency_sovereign_iso3' not found`,
`object '.attribute_dependency_land' not found`,
`unused argument (dependency_land = "sovereign")`), so none of it passes
vacuously. The `.read_land_areas()` block drives a 5-row fixture through a
mocked `.read_input()` rather than the 515k-row pin, because what is being
tested is which ISO3s survive the bridge, not the data.

Gates: `air format` clean, `lintr::lint_package()` with the CI config reports
`No lints found`, and `test_luh2_dependency_land.R`,
`test_luh2_land_area_repair.R`, `test_build_cbs.R`, `test_code_names.R`,
`test_energy_co2_extension.R`, `test_cow_to_lpjml_area_identity.R` all pass. No
roxygen touched (both helpers are internal, like their neighbours), so `man/` is
unchanged.

## Draft, because the modelling decision is still yours

This PR implements the plumbing and picks the *status quo* as the default, so
nothing published moves. It does not answer #407's actual question, which is not
mine to answer:

**Should a crown dependency's land go to its sovereign's polity?** Option 2 in
the issue, now one line away:

```r
dependency_land = c("sovereign", "drop")
```

The argument against is in the issue and I have not contradicted it: recovering
0.25% of `GBR` and 0.49% of `NLD` permanently redefines what those codes mean in
a land series. The argument for is the epic's own rule -- the crosswalk names a
polity for all six, and they are dropped only because the bridge is keyed by a
FABIO bucket rather than by identity. Option 3 (give the six their own
`polity_area_code`) cannot be done here; it needs an upstream crosswalk change.

Two things a reviewer should know before flipping it:

1. **The warning text is #382's, not fixed here.** With the default it still
   says "not found in polity_area_crosswalk" for the six, which is wrong about
   them; #382 splits the sentinel from the territories. I deliberately left
   those lines alone to avoid conflicting with it. With `"sovereign"` the six
   leave the warning entirely and only `-99` remains.
2. **`.read_luh2_cft()` in R/arable_permanent_land.R reads the same pin through
   its own bridge**, also keyed by `polity_area_code`, so it drops the six too.
   If the decision goes to `"sovereign"`, that reader needs the same treatment or
   the pre-1961 arable/permanent backcast will disagree with the land series.

Closes #407


lbm364dl:
Marked ready for review. Up to date with `main`, CI green.

Undrafted deliberately even though your open question stands. The reason: **this PR moves no published value by construction** — `dependency_land` defaults to `"drop"`, which is byte-for-byte today's behaviour. So merging it is safe regardless of how the modelling question is answered, and it converts that question from "someone must edit `.read_land_areas()`" into a one-line flag flip:

```r
dependency_land = c("sovereign", "drop")
```

The decision itself is still yours and still unanswered — should a crown dependency's land go to its sovereign's polity? — I have not pre-empted it. Two notes if you do flip it later: `.read_luh2_cft()` in `R/arable_permanent_land.R` reads the same pin through its own `polity_area_code`-keyed bridge and drops the same six, and the warning text still says "not found in `polity_area_crosswalk`" for territories the crosswalk does in fact know.

One tension worth flagging rather than resolving quietly: CLAUDE.md's multi-method rule says the default should be "the most rigorous available method", which would argue for `"sovereign"`. I kept `"drop"` because the choice here is a definition of what `GBR` and `NLD` mean in a land series rather than a question of rigour, and because a default that moves values is not mine to pick. If you read the rule as binding here, the flip is one line and no other code changes.

PR #507, opened 2026-08-04
Resolve synthetic fertiliser/manure country totals through the polity crosswalk

### LINKED ISSUE #464: Resolve synthetic fertiliser/manure country totals through the polity crosswalk
**Problem**
`.synthetic_n_country()` and `.manure_applied_country()` (`R/crop_soil_n2o_extension.R:145-178`)
read the raw `faostat-fertilizer-nutrients`/manure pins and take `area_code =
as.integer(.data[["Area Code"]])` directly — the literal FAOSTAT column, never joined against
`whep::polity_area_crosswalk`. `build_n_inputs()` then joins this against `crop_shares`, which IS
polity-keyed (descends from `get_primary_production()`), so any area whose raw code diverges from
its `polity_area_code` either silently fails to join or misattributes a country-year's synthetic-N
total.

**Evidence**
- `R/crop_soil_n2o_extension.R:145-178` — no crosswalk reference (grep confirmed 0 matches across
  this file, `R/n_balance_inputs.R`, `R/n_balance.R`).
- `R/n_balance_inputs.R:719-732` — `build_n_inputs()` joins the result against `crop_shares`.

**Target state**
Join `.synthetic_n_country()`'s/`.manure_applied_country()`'s output against
`whep::polity_area_crosswalk` (year-aware) immediately after reading the pin, renaming
`polity_area_code` onto `area_code`, mirroring `.aggregate_to_polities()`.

**Acceptance criteria**
- `build_n_inputs()`'s `country_totals` carries the same polity vocabulary as `crop_shares` before
  the join.
- Coverage of the 38 FAOSTAT-uncovered aggregate/non-territory codes is an explicit, documented
  exception, not a silent miss.

**Depends on**: none. **Decision needed**: drop unresolved aggregate codes vs. keep at
`resolution="polity"` only.

---

---

lbm364dl:
## What was wrong

`.synthetic_n_country()` and `.manure_applied_n_country()` (`R/crop_soil_n2o_extension.R`) read
the `faostat-fertilizer-nutrients` and `faostat-emissions-livestock` pins and took
`area_code = as.integer(.data[["Area Code"]])` -- the literal FAOSTAT column, never bridged through
`whep::polity_area_crosswalk`.

The crop shares those totals are split by descend from `get_primary_production()`, which
`.aggregate_to_polities()` keys on `polity_area_code`. So the `inner_join(by = c("year",
"area_code"))` in `.synthetic_n_inputs()` / `.manure_n_inputs()` -- and the same totals reused by
`build_n_inputs()`'s `.n_inputs_synthetic()` -- compared **two different vocabularies**. Every
FAOSTAT reporter whose code differs from its FABIO bucket fell out silently.

Measured on the real pins (1961-2023) on this branch, before touching any code:

| | value |
|---|---|
| FAOSTAT reporters in the synthetic-N total whose `area_code != polity_area_code` | **19** |
| ... which ones | Sudan 276 + South Sudan 277 -> **206**; Ethiopia PDR 62 -> **238**; 17 small territories (Aruba, Bermuda, Cook Is., Eq. Guinea, Eswatini, Faroe Is., French Guiana, Greenland, Guadeloupe, Martinique, New Caledonia, North Macedonia, Palau, Palestine, Reunion, Syria) -> **999** |
| country-years lost by the vocabulary mismatch, per total | **103** |
| joinable synthetic-N mass, summed over all years | 4,261,795,004 t N -> 4,271,752,119 t N (**+0.234%**) |
| joinable applied-manure mass, summed over all years | 1,461,141,053 t N -> 1,465,204,344 t N (**+0.278%**) |

The FABIO "rest of world" bucket 999 has **no FAOSTAT code at all**, so ROW cropland was receiving
zero synthetic fertiliser and zero applied manure. And post-2011 Sudan received nothing either:
FAOSTAT stops reporting 206 in 2012 and reports 276/277 instead, neither of which was in the crop
shares. The new end-to-end test returns `sum(result$impact_u) == 0` against the pre-change code
where it should be 5,894,460 -- 8 assertions in total fail without this change and pass with it.

## What I changed

One helper, `.n_country_to_polity()`, applied at the tail of both country-total readers. It bridges
`area_code` -> `polity_area_code` through `whep::polity_area_crosswalk` and sums the reporters that
share a bucket, exactly as `.aggregate_to_polities()` sums them. Two facts kept it this small:

- **No year predicate is needed.** `polity_area_code` is functionally determined by `area_code`: 0
  of the 550 crosswalk rows disagree. Checked against the year-aware `.add_polity_columns_dt()` path
  over the real pins, the two agree **to 0** on all 9,883 synthetic and 11,076 manure keys. (The
  issue's target state asked for year-aware; year-awareness selects a `polity_code`, which this
  aggregation key does not depend on. Same reasoning, and the same idiom, as the documented bridge
  in `R/arable_permanent_land.R`.)
- **The fold cannot double count.** No FAOSTAT year reports a predecessor together with its
  successors: the only multi-code polity-years across both pins are 276+277 under 206 and the ROW
  members under 999. 62 never coincides with 238.

## The open decision: drop unresolved aggregates

**Dropped, and documented in both the roxygen and the helper.** This is not a judgement call, it is
forced: I measured 40 aggregate/non-territory reporting codes across the two pins -- 5000 "World",
the 51xx-58xx continent/region rollups, 5707 EU-27, 5873 OECD, 5848/5849 Annex-I, 5817 NFIDCs, 420
Sub-Saharan Africa, and the "China" aggregate 351 that overlaps 41/96/128/214. Keeping them at
`resolution = "polity"` is impossible: **both** of `spatialize_country_n_to_crops()`'s resolutions
need a crop share, and a rollup has none of its own, so there is no output grain that could carry
them. Keeping them would double count the members they roll up. This costs no coverage -- they were
absent from the crop shares too, so the inner join already discarded them. The drop is now explicit
rather than silent.

Out of scope, as the issue notes: `inst/extdata/mueller_synthetic_n.csv` is keyed by `iso3c`, a
separate vocabulary. Untouched.

## How I verified

- After the change, **0** of the emitted area codes fall outside the polity vocabulary (was 40);
  999 is present in both totals; 206 has 12 post-2011 synthetic rows again (was 0).
- New tests: 3 in `test_crop_soil_n2o_extension.R` + `test_n_balance_inputs.R`, covering the helper
  against a fixture (whep ships no fertiliser/manure table -- both are remote pins), the
  `build_crop_soil_n2o_extension()` end-to-end join, and `build_n_inputs(resolution = "polity")`'s
  `country_totals`. Stashed the R change and confirmed **8 failures**; restored it, **0**.
- `test_crop_soil_n2o_extension.R` 24 pass, `test_n_balance_inputs.R` 75, `test_n_balance.R` 71,
  `test_n_balance_spatialize.R` 41, `test_n_balance_losses.R` 62, `test_n_percapita.R` 18,
  `test_footprint.R` 42, `test_read_raw_inputs.R` 3 -- all 0 fail / 0 error.
- `air format` on all three changed files; `lintr::lint_package()` with the exact CI config reports
  **zero lints**; `devtools::document()` run and `man/` committed; 0 non-ASCII bytes in the changed
  files.
- `test_exported_polity_columns.R` fails with `read_population` in the "without polity columns"
  census. That is the known `origin/main` red that #502 fixes, unrelated to this change.

## DRAFT: value magnitudes

**This moves published values, and no full-pipeline magnitude comparison was run.** The numbers
above are measured at the *country-total* level, and the crop-share key set they were joined against
was rebuilt the way `get_primary_production()` builds it rather than being that function's real
output -- a full-range `build_primary_production()` was off limits. So the direction and rough size
are measured (soil N2O and synthetic-N inputs go **up**, ~0.23-0.28% of joinable N mass, plus
whatever ROW's and post-2011 Sudan's newly-attributed N contributes), but the exact
`build_crop_soil_n2o_extension()` / `build_n_inputs()` deltas are not. Please run that diff before
merging.

Closes #464


lbm364dl:
Ran the `build_crop_soil_n2o_extension()` diff you asked for before merging.

Real pins, full range, `main` vs this branch:

| | main | this PR | |
|---|---|---|---|
| rows | 303,040 | 303,704 | **+664** |
| keys added / removed | — | **664 / 0** | nothing dropped |
| global `impact_u` | 4.337763e13 | 4.346631e13 | **+0.20%** |
| areas whose total moves | — | **3 of 194** | 191 bit-identical |

The 664 recovered keys land on exactly the three areas you named, and nowhere else:

| area | new keys | ratio |
|---|---|---|
| 206 Sudan | 356 | **1.546** |
| 999 Rest of World | 276 | **2.833** |
| 238 Ethiopia | 32 | **1.129** |

Those ratios look alarming out of context and are not: ROW was receiving **zero** synthetic fertiliser and zero applied manure because bucket 999 has no FAOSTAT code, and post-2011 Sudan the same because FAOSTAT stops reporting 206 in 2012. Getting a real number where there was none is the whole point of the PR. The new keys span 1961-2023 and carry 23% of the total move; the rest is the same country-years now joining against the right totals.

I also ran the invariant a sibling PR violated invisibly today (#563): **0 `(area_code, year)` pairs carry more than one polity label**, before *and* after — 11,247 → 11,261 pairs, none split. That is what distinguishes "recovered rows" from "split rows", and totals alone cannot tell them apart.

Full suite 5686 pass, 0 fail. Your out-of-scope note stands: `mueller_synthetic_n.csv` is still keyed by `iso3c` and untouched.

PR #509, opened 2026-08-04
fix(cropgrids): re-key the CROPGRIDS land pins to polity_area_code on read

### LINKED ISSUE #468: Key the CROPGRIDS land-use pin by polity_area_code, not raw FAOSTAT code
**Problem**
`data-raw/cropgrids_land.R` builds the `cropgrids-land` pin (inherited by `cropgrids-fallow-land`)
keyed on `regions_full$code` (raw FAOSTAT), even though `regions_full` already carries
`polity_area_code`. `build_cropgrids_land_extension()` joins this against `harvested`, which IS
`polity_area_code`-keyed. MEASURED: area 276 (Sudan) and 277 (South Sudan) both map to
`polity_area_code 206`, but the pin keeps them as raw 276/277, so neither ever matches `harvested`'s
area_code 206 — both lose their CROPGRIDS-specific multi-cropping ratio and silently fall back to
the global ratio.

**Evidence**
- `data-raw/cropgrids_land.R:70-73, 86-101`; `data-raw/cropgrids_fallow.R:32`.
- `R/crop_land_extension.R:264-310, 296-302, 319-322, 339-357` (silent fallback to `ratio_global`).
- MEASURED: 64/272 `regions_full` rows have `code != polity_area_code`.

**Target state**
In `data-raw/cropgrids_land.R`, replace `area_code = as.integer(code)` with `area_code =
as.integer(polity_area_code)` (already present on `regions_full`), dropping NA rows, letting the
existing `summarise(.by=c(area_code, item_cbs_code))` aggregate Sudan+South Sudan correctly.

**Acceptance criteria**
- Sudan and South Sudan both get the CROPGRIDS-specific ratio, not the global fallback.
- Pin is rebuilt and re-pinned; before/after ratio diff documented.

**Depends on**: none. **Decision needed**: the pin must be rebuilt from external CROPGRIDS source
files — a real external action, not just a code diff.

---

---

lbm364dl:
## What was wrong

`data-raw/cropgrids_land.R` keys the `cropgrids-land` pin on `regions_full$code` — the **raw
FAOSTAT** area code. `cropgrids-fallow-land` is built from that pin, so it inherits the keying. But
the `harvested` side of `build_cropgrids_land_extension()` comes from `get_primary_production()`,
whose `area_code` is a **`polity_area_code`** (`R/build_production.R:599`, `area_code =
polity_area_code`).

So every area whose two codes differ never matched, and fell silently through to the global per-item
ratio at `R/crop_land_extension.R:302-328` — losing exactly the thing the pin exists to provide, the
per-crop multi-cropping correction.

`regions_full`: **64 of 272** rows have `code != polity_area_code`. Sudan `276` and South Sudan `277`
both carry `polity_area_code 206`.

Read the live `cropgrids-land` pin (6269 rows, 183 areas) and joined it to the crosswalk:

| pin `area_code` | `polity_area_code` | rows | harvested Mha |
|---|---|---|---|
| 276 (Sudan) | 206 | 47 | 17.395 |
| 212 | 999 | 41 | 4.280 |
| 277 (South Sudan) | 206 | 51 | 1.528 |
| 154 | 999 | 37 | 0.333 |
| 209 (Eswatini) | 999 | 44 | 0.174 |
| 61 (Eq. Guinea) | 999 | 35 | 0.095 |
| 69 | 999 | 21 | 0.009 |

**7 areas, 276 of 6269 rows, 1.71% of pin harvested area** were mis-keyed. And `206` is **absent
from the pin entirely** (`206 %in% pin$area_code` -> `FALSE`), so all 51 Sudanese crop items were on
the global fallback. Ratio actually used vs. Sudan's own:

| `item_cbs_code` | harvested ha | ratio before (global) | ratio after (Sudan) | d ha |
|---|---|---|---|---|
| 2518 sorghum | 6,349,736 | 0.9297 | 0.9991 | +440,939 |
| 2517 millet | 2,408,738 | 0.8895 | 0.9990 | +263,958 |
| 2552 sesame | 3,112,221 | 0.9709 | 0.9987 | +86,412 |
| 2549 | 480,934 | 0.9297 | 1.0000 | +33,790 |
| 2535 | 96,175 | 0.8758 | 0.9999 | +11,941 |

Sudan is semi-arid and near-single-cropped, so the global multi-cropping correction was understating
its physical cropland.

## What I changed

`.read_cropgrids_land()` now re-keys and sums the pin to `polity_area_code` before returning it, via
a new `.cropgrids_to_polity_area()` helper. Both aliases get it, so the fallow pin is fixed the same
way.

**I deliberately did not take the issue's literal target state** (changing
`area_code = as.integer(code)` to `polity_area_code` in `data-raw/cropgrids_land.R`), because
re-keying on read is strictly better here and removes the blocker the issue flagged:

- **No pin rebuild is required.** The issue recorded "the pin must be rebuilt from external CROPGRIDS
  source files — a real external action". Doing the aggregation on read makes the *shipped* pin
  correct today, with no figshare download and no re-pin.
- **The source pin keeps CROPGRIDS' own reporting grain**, so Sudan and South Sudan stay separable.
  Pre-aggregating at build time would discard that permanently.
- **It matches the epic's own framing.** `polity_area_code` is an aggregation bucket, not an
  identity, so it is applied where the aggregation happens rather than baked into stored source data.
  This is the same shape as the existing `.cells_to_polity_area()`
  (`R/feed_intake_redistribute.R:1077`), which leaves `country_grid` raw and re-keys in the consumer,
  and as `.grassland_faostat_to_polities()`.
- **It is idempotent**, so a pin rebuilt on `polity_area_code` later passes through unchanged — the
  two approaches cannot conflict.

I added a comment in `data-raw/cropgrids_land.R` recording that the raw keying is deliberate and
where the aggregation happens, so nobody has to rediscover this.

## How I verified

Against the **live pin**, through the real `.read_cropgrids_land()`:

- 183 -> 178 areas, 6269 -> 6101 rows; `206` now present, `276`/`277` gone.
- `physical_ha` and `harvested_ha` totals **conserved exactly** — the re-key aggregates, it does not
  create or drop area.
- `276 + 277 == 206` for **all 51 items**.
- **Idempotent**: a second pass over the re-keyed table returns it unchanged. Structurally, all 266
  crosswalk area codes map their own `polity_area_code` to itself (0 non-identity targets).

Tests added to `tests/testthat/test_crop_land_extension.R`:

- `.cropgrids_to_polity_area` re-keys 276/277 -> 206, sums **before** the ratio is taken (weighted
  1239/1300, not the mean of 0.999 and 0.8), leaves identity-mapped areas alone, and conserves totals.
- Idempotence, including the `999` ROW bucket.
- `.read_cropgrids_land` applies the re-key for **both** aliases, with `whep_read_file` mocked. The
  pins are remote, so this is the fixture-based test of the changed line — **it fails on `main`** (6
  failures: `expect_equal(res$area_code, 206L)` gets `c(276L, 277L)`, for each alias).
- End-to-end: the same harvested row takes the global fallback `1739/2300` against a raw-keyed table
  and Sudan's own `1239/1300` against the re-keyed one.

Gates: `air format` clean; `lintr::lint_package()` with the exact CI linter config reports **"No
lints found"**; `devtools::document()` re-run and `man/` committed; `test_crop_land_extension.R`
**117 pass, 0 fail**. No non-ASCII introduced (the only non-ASCII in the file is pre-existing roxygen
prose). No doc example hits the network, so `--as-cran` examples are unaffected.

## Moves published values — opened as DRAFT

Yes. Using the pin's own harvested area as a **proxy** base:

| aggregation bucket | before | after | |
|---|---|---|---|
| 206 (Sudan combined), total | 18.0428 Mha | 18.9044 Mha | **+4.78%** |
| 206, the 23 of 51 items that change | 13.0741 Mha | 13.9356 Mha | **+6.59%** |
| 999 (ROW) | 3.8381 Mha | 3.8998 Mha | **+1.61%** |

55 `(area, item)` rows move. **This is a proxy magnitude only** — I did not run
`build_primary_production()` or a full-range build, so I have not measured the effect on the shipped
extension across all years, nor on downstream footprints. `build_fao_arable_fallow_extension()`
(`R/arable_permanent_land.R:538`) and `.hayr_attributed_fallow()` both consume this read path and
will move the same way; both stay internally consistent, since the two pins are re-keyed identically.
`attribute_fallow_to_crops()` is untouched, so `data-raw/cropgrids_fallow.R` — which works entirely
in raw FAOSTAT space against the raw landuse pin and raw gridded output — is unaffected.

Also of note: areas `212`, `154`, `209`, `61`, `69` now contribute a real CROPGRIDS ratio to bucket
`999`, which previously had none.

Closes #468. Part of the polity migration epic #458.


lbm364dl:
Ran the full-pipeline before/after you flagged as missing, and merging on it.

`build_cropgrids_land_extension()`, real pins, `main` vs this branch:

| | main | this PR | |
|---|---|---|---|
| rows | 705,581 | 705,581 | unchanged |
| `(year, area, item)` keys added / removed | — | **0 / 0** | no structural change |
| global `impact_u` | 1.478673e11 | 1.479162e11 | **+0.033%** |
| areas whose total moves | — | **2 of 194** | 192 bit-identical |

| area | ratio |
|---|---|
| 206 Sudan | **1.0559** |
| 999 Rest of World | **1.0159** |

Your proxy predicted +4.78% on 206 and +1.61% on 999 from the pin's own harvested area; the real extension gives **+5.59%** and **+1.59%**. Close enough that the proxy was doing its job, and now it does not have to.

The structural line is the one I would draw attention to: **0 keys enter or leave, and only two areas move.** A re-key that quietly split or dropped rows would show up there, and today that is not a hypothetical — a sibling PR did exactly that invisibly (#563), which is why this check now runs on everything that touches area codes.

Full suite 5673 pass, 0 fail. Merging.

Issue #386, opened 2026-07-28
14 of 17 regional classification columns in regions_full have no consumer


lbm364dl:
Found while working #382.

`regions_full` ships 17 regional classification columns. Counting references across `R/`, `data-raw/`, `tests/` and `vignettes/` (excluding the roxygen block in `R/datasets.R` that documents them):

| column | consumers |
|---|---|
| `region_krausmann` | 18 |
| `region_HANPP` | 4 |
| `ADB_Region` | 2 |
| `EU27`, `region_code`, `region_krausmann2`, `region_UN_sub`, `region_UN`, `region_ILO1`, `region_ILO2`, `region_ILO3`, `region_IEA`, `region_IPCC`, `region_labour`, `region_labour_agg`, `region_labour_mech`, `region_test` | **0** |

One of them is called `region_test`, which reads like a leftover rather than a classification anyone intends to ship.

They are not free. Each is a country grouping that has to stay consistent with the polity structure as upstream splits and merges polities — `test_region_classifications.R` (added in #382) now asserts none of them contradicts itself within a polity, so all 17 are load-bearing on that check even though 14 are read by nothing. And they are documented in `R/datasets.R`, so they carry a maintenance claim that they mean something current.

Options:

- **Drop the 14.** Breaking for anyone using them off-label, but they are undocumented-as-supported and unused in-tree. `region_test` should go regardless.
- **Keep, and say why.** If they exist for downstream analysis outside this package (plausible for `region_UN`, `region_IPCC`, `region_ILO*` — all standard published taxonomies), that is a fine reason, but it should be stated in the `@format` block so the next person does not re-litigate it.

Not doing either unilaterally: these are columns of a public dataset, so removal is the maintainer's call. Worth noting the *consistency* question is already settled — #382 measured all 17 and they agree within every non-aggregate polity, the only conflict being `ROW-1850-2023`, which spans continents by construction.

lbm364dl:
## Measured: the 14 unconsumed columns have one structured gap, not general incompleteness

Relevant to deciding this issue, since the question is whether these columns are safe to consume.

Six of them are **present-day classifications** — `region_UN`, `region_UN_sub`, `region_ILO1/2/3`, `region_IEA`, `region_IPCC` — and each is missing exactly the same five polities:

| polity | why |
|---|---|
| `F51-1947-1993` Czechoslovakia | dissolved |
| `F228-1945-1991` USSR | dissolved |
[...]

PR #510, opened 2026-08-04
Join the residue feed-use fraction on its real UN M49 sub-region key

### LINKED ISSUE #405: residue_feed_fraction.csv's region column holds UN sub-regions, so every region silently gets the 0.20 Global fallback
Found on #382 while sweeping `inst/extdata` files that define regions. Baselined there, not fixed, because the repair changes model outputs across most of the world.

## The defect

`inst/extdata/coefs/residue_feed_fraction.csv` has a column named `region_hanpp`. It does not contain HANPP regions. Its 17 named values are **UN M49 sub-regions** — Sub-Saharan Africa, Western Asia, South-Eastern Asia, Melanesia, Australia and New Zealand, and so on.

Measured against `whep::regions_full`:

| column | n | matches the file's 17 |
|---|---|---|
| `region_HANPP` | 8 | **0** |
| `region_UN_sub` | 18 | **14** |
| `region_krausmann` | 11 | 5 |
| `region_UN` | 6 | 1 |

`.residue_destiny_krausmann()` joins the file `by = "region_hanpp"`, and the pipeline supplies that column from `regions_full$region_HANPP` (`soil_carbon_inputs.R`: `region_hanpp = .data$region_HANPP`). Eight HANPP labels against seventeen UN sub-region labels match nothing — and the next line is

```r
feed_use_fraction = tidyr::replace_na(feed_use_fraction, global_feed)
```

so every row silently takes the `"Global"` default of **0.20**.

## Confirmed by running it, not by reading it

Six different regions through `.residue_destiny_krausmann()`:

```
region                        implied feed fraction
South and Central Asia        0.2
Sub-saharan Africa            0.2
East Europe                   0.2
Latin America and Caribbean   0.2
...                           0.2
```

Recovery rates from the *other* join varied correctly across the same rows, so the Krausmann join works. **Only this one is dead.**

## What it costs

The file spans **0.05** (Northern America, Oceania) to **0.45** (Southern Asia) — a ninefold range — and all of it collapses to one number. That shifts the feed-versus-burn split of crop residues almost everywhere, which propagates into the nitrogen and carbon balances.

## Proposed fix

Join on `region_UN_sub`, which the pipeline can carry because `regions_full` already has it. Three of the file's values need attention first:

- `South-Eastern Asia` → `South-eastern Asia` (pure case difference; this is why the overlap is 14 and not 15)
- `Caribbean` and `Oceania` — finer granularity than `regions_full` carries (which has `Latin America and the Caribbean`, and `Australia and New Zealand` / `Melanesia` / `Micronesia` / `Polynesia`). Either unused, or the source rows want aggregating.
- `regions_full` also has `Micronesia`, `Polynesia` and `RoW` with no row in the file; those would keep the Global fallback, which seems right.

I have not made this change because the coefficient values come from cited literature (Smil 1999, Lal 2005, Krausmann 2008, Erenstein 2014, McIntire 1992) and which regional taxonomy each belongs to is a scientific judgement, not a lookup — and the result moves published numbers.

## Baselined meanwhile

`tests/testthat/test_residue_feed_region_vocabulary.R` pins both the vocabulary mismatch and the resulting uniform 0.20, in both directions. It fails when the fix lands (asserting the uniformity is gone) and fails if the overlap changes shape for some other reason. A silent bug that nothing measures is how a ninefold coefficient range stays dead for as long as this one has.

lbm364dl:
Closes #405. Part of the polity migration epic #458.

**Draft, because it moves published values.** No magnitude comparison of the moved columns was run — see "What moves" below for what is measured and what is not.

## What was wrong

`inst/extdata/coefs/residue_feed_fraction.csv` named its region column `region_hanpp`, but its 17 named values are UN M49 sub-regions — `Sub-Saharan Africa`, `Western Asia`, `Melanesia`, `Australia and New Zealand`. `.residue_destiny_krausmann()` joined it `by = "region_hanpp"` against a column `.sci_crop_regions()` fills from `regions_full$region_HANPP`, whose 8 labels (`East Asia`, `West Europe`, `Sub-saharan Africa`, ...) share nothing with M49. The join matched nothing, and the next line

```r
feed_use_fraction = tidyr::replace_na(feed_use_fraction, global_feed)
```

handed every polity on earth the `"Global"` default of `0.20`. A published table spanning `0.05` to `0.45` was dead in full.

It stayed invisible because `replace_na()` leaves an unmatched region indistinguishable from a matched one: mass still conserves, the row count is right, and no column is ever `NA`.

## Reproduced before the change

Driving the split with the region pairs `.sci_crop_regions()` actually supplies, not a hand-built fixture:

```
              region_krausmann                region_hanpp  implied feed fraction
    North Africa and West Asia  North Africa and West Asia                    0.2
                     East Asia                   East Asia                    0.2
        South and Central Asia      South and Central Asia                    0.2
            Sub-saharan Africa          Sub-saharan Africa                    0.2
                   East Europe                 East Europe                    0.2
   Latin America and Caribbean Latin America and Caribbean                    0.2
     North America and Oceania   North America and Oceania                    0.2
                   West Europe                 West Europe                    0.2

distinct implied fractions: 0.2
```

11 of 11 pairs, all 8 HANPP regions, one distinct value. Vocabulary measured against every candidate column of `regions_full`: `region_UN_sub` matches 14 of 17 exactly (15 modulo one capital `E`), `region_HANPP` matches **0**.

## What I changed

The Krausmann bridge (`.residue_recovery_region()`, which translates `region_krausmann` -> `region_HANPP` through `regions_full`) could **not** be reused here, and that is why this was a judgement call rather than a copy: `region_UN_sub` -> `region_HANPP` is not 1:1, and the disagreements are legitimate rather than data errors. M49 puts Sudan in Northern Africa where HANPP puts it with Sub-saharan Africa; Greenland is M49 Northern America but HANPP West Europe. Any such translation would have to *choose*.

So the caller supplies the M49 sub-region directly instead:

- `residue_feed_fraction.csv`: column renamed `region_hanpp` -> `region_un_sub`. One value changed, `South-Eastern Asia` -> `South-eastern Asia`, to match `regions_full`'s spelling. No coefficient changed.
- `.sci_crop_regions()`: carries `region_un_sub` **alongside** `region_hanpp`. `region_hanpp` is kept because it is not spurious — `modern_variety_adoption.csv`'s `region_hanpp` column really does hold HANPP regions (8 of 8 match `regions_full`), and `calculate_crop_npp()` keys the modern-variety adoption share by it. Only the feed table was mislabelled.
- `.residue_destiny_krausmann()`: joins the feed table on `region_un_sub`; `krausmann_regional` now **requires** `region_un_sub` and aborts without it, instead of falling back to `0.20` for every row. That closes the hole rather than leaving a second way to re-open it.
- Six existing test files that built `krausmann_regional` inputs updated to the new column.

**The two rows I did not resolve, deliberately.** `Caribbean` and `Oceania` are coarser levels of the same M49 hierarchy and have no `region_UN_sub` counterpart, so they stay unused:

- `Caribbean` costs nothing: its `0.2` is identical to the `Latin America and the Caribbean` row that Caribbean areas do match. Asserted in the new test.
- `Oceania` at `0.05` would only ever apply to Micronesia and Polynesia, which have no row of their own. They keep the `0.20` Global fallback they already have today, so **this PR does not move their values either way** — but whether the coarse Oceania figure ought to cover them is a coefficient-authoring question I have left to you. Both rows are kept in the file (deleting published coefficients is not a bug fix) and pinned in the test with that reasoning.

Also out of scope and left alone: the region-per-area lookup still keys on the numeric `area_code` from `regions_full$code`, as `region_krausmann` and `region_hanpp` already did. Region columns as polity-derived attributes is #386's territory, not this fix's.

## How I verified

After the change, the same call that produced one value produces eight:

```
distinct implied fractions: 0.05 0.1 0.15 0.2 0.25 0.3 0.4 0.45
areas matched a region-specific row: 230 of 261 (88.1%)
areas on the Global fallback:         31  -- Micronesia 9, Polynesia 10, RoW 1,
                                            11 with no M49 sub-region
```

The file has no row for any of those 31, so `0.20` is the fallback rather than a gap.

End to end through the real chain, `.sci_crop_prod_wide() |> calculate_crop_npp() |> calculate_residue_destinies()`, confirming `region_un_sub` survives the NPP step:

```
 area_code    region_un_sub  residue_feed_dm_t  residue_burn_dm_t  implied
       231 Northern America              47.51             902.72     0.05
       100    Southern Asia             598.15             731.08     0.45
```

`.sci_npp_from_primary_prod()` on the repo's own fixture still returns positive `residue_soil_c_t` and the same six columns.

## What moves

`residue_feed_dm_t` and `residue_burn_dm_t` change, and with them `build_residue_feed_avail()` and the nitrogen balance's `used_residue_n_t` / `burnt_residue_n_t`. I did not measure those magnitudes.

What I *did* measure — as a test, not a claim — is the scope: `residue_soil_dm_t` does not change, and neither does `residue_feed_dm_t + residue_burn_dm_t`. The first is `residue_dm_t * (1 - recovery_rates)` and the second `residue_dm_t * recovery_rates`; neither mentions `feed_use_fraction`. So `build_soil_carbon_inputs()`'s residue carbon is untouched, and only the feed-versus-burn split of an unchanged removed total moves.

## Tests

New `tests/testthat/test_residue_feed_region_vocabulary.R`, 14 assertions in 5 blocks. Against the pre-fix code it reports **8 failures over 4 of the 5 blocks**. It pins:

- the vocabulary in both directions — 15 of 17 named rows are exactly a `regions_full$region_UN_sub` value, **0 of 17** are a `region_HANPP` value, so the mislabel cannot come back quietly;
- that the join is live, by driving `.sci_crop_regions()`'s real region pairs through the public function and asserting the full eight-value range rather than a single number;
- that three probe sub-regions receive their published fractions (`0.45`, `0.05`, `0.1`) with the recovery region held constant, so the fraction demonstrably comes from `region_un_sub` alone;
- the value-move scope algebraically (soil return and removed total invariant across two different feed fractions);
- that supplying `region_hanpp` alone now aborts.

Gates: `air format` on every touched file; `lintr::lint_package()` with the exact CI config reports **no lints**; `devtools::document()` run and `man/` committed. Test files run green: `test_residue_feed_region_vocabulary` 14, `test_residue_destiny` 16, `test_properties` 25, `test_coef_integrity` 42, `test_coef_tables` 6, `test_npp_bnf_pipeline` 7, `test_validation` 20, `test_n_balance` 71, `test_crop_npp` 35 — 0 failures, 0 errors.

`tests/testthat/test_exported_polity_columns.R` fails on `origin/main` for an unrelated reason (#502 fixes it) and is untouched here.


lbm364dl:
Measured the value movement on real data before merging. It behaves exactly as you described, including the part you said would not move.

Driving the real `.sci_crop_regions()` through `calculate_residue_destinies(method = "krausmann_regional")` with 1000 t of residue per area, so the resulting feed mass **is** the fraction each area receives — 261 areas:

| | main | this PR |
|---|---|---|
| distinct implied feed fractions | **5** | **17** |
| range | 0.14 – 0.18 | 0.035 – 0.405 |
| mean | 0.16004 | 0.21003 |

The 5 on `main` are not five feed fractions — they are recovery varying by region against a **constant** `feed_use_fraction` of 0.20 (0.8 × 0.20 = 0.16 mean). So the table really was dead in full, as you said.

The redistribution, and the reason this is safe:

| | ratio |
|---|---|
| `residue_feed_dm_t` | **1.3124** |
| `residue_burn_dm_t` | **0.9219** |
| `residue_soil_dm_t` | **1.0000**, `all.equal()` TRUE |

**Total per area conserved exactly** (261,000 → 261,000). Mass moves between feed and burn only, and the soil pathway is bit-identical — which is your claim that `build_soil_carbon_inputs()`'s residue carbon is unaffected, now measured rather than argued. 154 of 261 areas change.

I did not measure the propagation into `build_nitrogen_balance()`'s `used_residue_n_t` / `burnt_residue_n_t`; that is a gridded multi-hour build needing local rasters. Given feed and burn are the only movers and they conserve against each other, the N split moves by the same 1.31 / 0.92, but I have not run it and am not claiming it.

Your two deliberate non-decisions stand: `Caribbean` and `Oceania` stay unused, and Micronesia/Polynesia keep the 0.20 fallback — so this PR still does not move their values either way.

Full suite 5700 pass, 0 fail.

PR #511, opened 2026-08-04
Key the historical CBS proxy fills on the polity, not on an area name

### LINKED ISSUE #493: Key .fill_with_proxies() on polity_area_code instead of area name; decide what an aggregate's proxy means
Found while implementing #470 (PR #481).

## Problem

`.fill_with_proxies()` joins its GDP/population proxies **by area name**. Because names are not a stable key, `.canonicalise_gdp_pop_area()` exists to reconcile them — and it leaves **6 areas unfilled**: Syria, Equatorial Guinea, French Guiana, North Macedonia, Palestine and Eswatini.

All six share one cause: their FAOSTAT area folds into an aggregate polity's bucket (`999`), so there is no name to match. Relabelling them as the aggregate would be worse than leaving them empty — it would attribute one member's population to the whole bucket.

## The fix, and the decision inside it

Mechanically the fix is the epic's own rule: **stop joining proxies by name and key `.fill_with_proxies()` on `polity_area_code`.** That removes the canonicalisation layer entirely rather than extending it, since the numeric key is stable where the name is not.

But it exposes a question the name-join was hiding: **what is a Rest-of-World population?** The probable answer is "the sum over the bucket's members", which is the only reading under which a per-capita rate computed against it means anything. That is a methodological choice about aggregate proxies, not plumbing, so PR #481 did not make it.

If the summing rule is confirmed, the work is: sum member proxies to the bucket, key the join on `polity_area_code`, delete `.canonicalise_gdp_pop_area()`.

## Note

This **moves published values** — the six areas currently receive no proxy and would start receiving one. Needs a magnitude comparison.

Refs #458. Related: #419 (if the fold stops, five of the six get their own proxy directly and this shrinks to Sudan-style cases only).

lbm364dl:
Part of the polity migration epic #458. Closes #493.

**Draft, because this moves published values and no magnitude comparison against
`main` was run.** The pre-1962 historical CBS fills reach 43 polities they
previously missed for the population proxy and 55 for the agricultural-land
proxy. The only way to measure the magnitude is a full-range build; the
instruction for this task was not to run one.

## What was wrong

`.fill_with_proxies()` joined **both** of its proxies -- population for
`food`/`other_uses`/`processing`, agricultural land for `feed` -- with
`by = c("year", "area")`. Three independent name vocabularies met at that join:

| side | label vocabulary | example for FAO area 3 |
|---|---|---|
| the frame being filled | periodized `polity_name` (`.aggregate_to_polities()` renames it to `area`) | `Albania (1913-2025)` |
| the `gdp-population` pin | its own labels, keyed by ISO3 | `Albania` |
| `.read_land_areas()` (LUH2) | the crosswalk's **static** `area_name` | `Albania` |

Where the labels diverge the join finds nothing and the territory keeps its gaps,
with no warning.

## Evidence it reproduced BEFORE the change, on this branch

Counted against the vocabulary `.aggregate_to_polities()` actually emits (the
polity active at the 1961 back-cast anchor, since this fill only runs pre-1962):

- **57 of the pin's 196 area names — 8,263 rows, 27.8%** — are names no builder
  emits. Every one of them resolves on the polity key. Among them: `Albania`,
  `Poland`, `Romania`, `Hungary`, `Turkey`, `Russian Federation`, `Pakistan`,
  `Bangladesh`, `China, mainland`, `USSR`, `Czechoslovakia`.
- **96 of the LUH2 land labels — 41.7% of land rows** — likewise.
- Coverage of the frame's (year, polity) cells over 1850-1961: **13,664 of
  22,624 (60.4%)** had a population proxy. Over 1900-1902, **402 of 606 (66.3%)**
  had `agriland`.
- Functionally, on a fixture with each proxy in its native vocabulary (pin says
  `Albania`, LUH2 says `Albania`, frame says `Albania (1913-2025)`), `food = 100`
  in 1950 with gaps 1951-1955:

```
BEFORE (joined by area name):          AFTER (keyed on polity_code):
 year   food feed  pop agriland         year   food feed  pop agriland
 1950    100   10   NA       NA         1950    100   10 1000      2.0
 1951     NA   NA   NA       NA         1951    110   11 1100      2.2
 1952     NA   NA   NA       NA         1952    120   12 1200      2.4
 ...                                    ...
```

The `BEFORE` column was produced by installing `main`'s `.fill_with_proxies()`
into the loaded namespace, so it is the same frame through the old code.

## What I changed

Both proxies are resolved to a `polity_code` and joined on that. No name appears
in a join key any more.

- `.proxy_polity_key()` puts an ISO3-keyed proxy table on the polity key, using
  the same two helpers the builders use (`.iso3_to_fao_area_code()`,
  `.add_polity_columns_dt()`), so both sides agree by construction including the
  pre-1962 back-cast.
- `.read_land_areas_wide()` now emits `(year, polity_code, Cropland, Pasture,
  agriland)` instead of carrying a label.
- `.polity_code_from_labels()` recovers the frame's own polity identity from the
  `(polity_area_code, polity_name)` pair the rename flattened it into.
  `polity_name` is unique per `polity_code` across the crosswalk (**476 names,
  476 codes**), so this is a lookup on the crosswalk's own two columns, not a
  match between vocabularies.

Note that `polity_area_code` **alone** is not a sufficient key, which is worth
recording: bucket `206` carries two distinct frame labels at the 1961 anchor
(`Sudan` from FAO 206/276 and `South Sudan` from FAO 277), and both currently get
their own population from the pin. Keying on the bucket would have merged them.
Keying on `polity_code` keeps them apart. It is the only such bucket today.

## The decision I did not make

Aggregates reachable only by **folding other territories into them** keep no
proxy. Summing the six crosswalk members that fold into Rest of World (`999`) --
Equatorial Guinea, French Guiana, North Macedonia, Palestine, Eswatini, Syria --
would give the bucket a population that is neither one member's nor the real rest
of the world's, and a per-capita rate against it would mean nothing. That is the
methodological question in #493 and it stays open; until it is answered those
buckets are exactly as unfilled as the name-keyed join left them.

Aggregates that **report as themselves** do keep a proxy: the pin carries `RAFR`,
`RASI`, `REUR`, `RLAM`, `ROCE` and `BLX`, which *are* `Africa Other` ..
`Belgium-Luxembourg` rather than members of them, so nothing is being summed and
there is no choice to defer. My first attempt excluded aggregates wholesale and
lost those six; the `area_code == polity_area_code` carve-out (the same idiom
`R/energy_co2_extension.R` already uses) restores them, which is why the
after-measurement below has zero losses.

## How I verified

- Coverage, measured on the real pin and real LUH2 pin:

| proxy | window | before | after |
|---|---|---|---|
| population | 1850-1961, all (year, polity) cells | 13,664 / 22,624 (60.4%) | **18,480** (81.7%) |
| agricultural land | 1900-1902 | 402 / 606 (66.3%) | **567** (93.6%) |

  **43 polities gain a population proxy and none lose one; 55 gain agriland and
  none lose.** Strictly additive.
- No fan-out: 606 frame cells in, 606 out.
- Four new tests, and two existing fixtures updated to the new proxy contracts.
  Run against `main`'s definitions of the two changed functions they produce
  **6 failures**; on this branch `test_build_cbs.R` is **83 passing, 0 failing**.
  The tests use fixtures rather than shipped data because neither proxy table is
  exported -- both only exist inside a build -- and the frame they fill is
  assembled mid-pipeline.
- `test_polity_output_coverage.R`, `test_supply_use.R`, `test_feed_intake_build.R`,
  `test_toy_examples.R`, `test_build_integration.R`,
  `test_select_best_source_aggregates.R`: 0 failures.
- `air format`, and `lintr::lint_package()` with the exact CI linter config:
  no lints.

(`tests/testthat/test_exported_polity_columns.R` fails on `main` for unrelated
reasons; PR #502 fixes it. Untouched here.)


lbm364dl:
Ran the full-range comparison you said was the only way to measure this, and merging on it.

Two complete `get_wide_cbs()` builds, `main` vs this branch.

## The era split, which is the reassuring part

**1962 onwards is bit-identical** — same 924,543 rows, and `production`, `feed`, `food` all at ratio **1.00000000**. The change is confined to the era the fills run in, by construction and now by measurement.

## Pre-1962, where the fills run

Rows **1,080,724 → 1,114,801** (+34,077, +3.2%). Nothing dropped.

| column | ratio |
|---|---|
| feed | **1.0988** |
| processing | 1.0939 |
| domestic_supply | 1.0777 |
| food | 1.0677 |
| production | 1.0126 |
| seed | 0.9441 |
| import | 0.9777 |
| export | 1.3563 |
| stock_withdrawal | 1.3849 |

Full range that comes to feed **+3.97%**, food +2.00%, production +0.17%, stock_withdrawal +31.4%.

These are large, and they are the fix rather than a side effect: 57 of the pin's 196 names (27.8% of rows) and 96 of the LUH2 labels (41.7% of land rows) were names no builder emits, so those territories kept their gaps. Filling them adds mass. The redistribution across columns (seed and import *down*) is the balance re-solving, not rows appearing from nowhere.

## The integrity check that decided it for me

```
main : rows 2,005,267,  supply-use imbalances 48,235 (2.4054%),  max abs diff 2.520e+08
#511 : rows 2,039,344,  supply-use imbalances 48,235 (2.3652%),  max abs diff 2.520e+08
```

**Identical imbalance count and identical maximum residual**, while adding 34,077 rows. The fills land in balanced rows or fill both sides consistently; they do not manufacture a single new imbalance. Given #382's experience — where a relabelling conserved mass and still dropped 702,166 rows — that is the property I most wanted to see, and totals alone would not have shown it.

Full suite 5711 pass, 0 fail.

Your open question stands and I have not answered it: aggregates reached only by folding other territories into them (Rest of World) still get no proxy, because what an aggregate's proxy *means* is a modelling question.

Issue #489, opened 2026-08-04
biomass_coefs has two sources with divergent values: package data (68 cols) vs pin (41 cols), 12 shared columns disagree


lbm364dl:
## Problem

`biomass_coefs` is reachable by two paths that return **different data**, and both
are in use in `R/`:

| path | callers |
|---|---|
| `whep::biomass_coefs` (package data, `data/biomass_coefs.rda`) | `R/build_production.R`, `R/food_supply.R`, `R/feed_intake_redistribute.R` |
| `whep_read_file("biomass_coefs")` (remote pin, `whep_inputs.csv:34`, version `20250728T082553Z-f2fac`) | `R/n_prov_destiny.R:42, 1181, 1578`, `R/Typologies_Josette.R:112` |

MEASURED, both read on 2026-08-04:

```
package data : 421 rows x 68 cols
pin          : 421 rows x 41 cols
pin columns not present in the package data: 0   (the pin is the narrower schema)
shared columns: 41, of which 12 DISAGREE in value
```

The 12 disagreeing columns are `Equiv`, `Category`, `Product_kgDM_kgFM`,
`Residue_kgDM_kgFM`, `GE_residue_MJ_kg`, `Edible_portion`, `N_kgN_kgFM`,
`Product_kgN_kgDM`, `Product_kgC_kgDM`, `Residue_kgN_kgDM`,
`Residue_humified_kgC_kgC`, `Root_humified_kgC_kgC`.

Cell-level divergence is narrow but individually large:

| column | pkg non-NA | pin non-NA | rows differing | median pin/pkg on those |
|---|--:|--:|--:|--:|
| `Edible_portion` | 215 | 214 | 9 | 1.042 |
| `N_kgN_kgFM` | 258 | 258 | 4 | **0.492** |
| `Product_kgN_kgDM` | 362 | 362 | 2 | **0.238** |
| `Product_kgDM_kgFM` | 371 | 371 | 2 | 0.724 |

So a handful of commodities carry nitrogen or dry-matter coefficients that differ
by a factor of 2 to 4 depending on which path read them. `build_food_supply()`
and `create_n_prov_destiny()` are therefore using **different nutrition
coefficients for the same commodities**, with nothing declaring which is
authoritative.

## Why this matters for #361

`Edible_N_kgFM` is **absent from the pin schema entirely**. The empty `Edible_*`
block that #361 is about exists only in the 68-column packaged `.rda`. My
[...]

Issue #384, opened 2026-07-28
No gate detects a stale data/*.rda against its data-raw/ builder


lbm364dl:
Found while working #382.

`data/regions_full.rda` and `data/polities_cats.rda` had drifted from what `data-raw/harmonization_tables.R` produces. Re-running the builder changed 12 rows, and **8 of them had been resolving to polities that upstream had retired or superseded**:

| area | stale target | correct target |
|---|---|---|
| Angola | `AGO-1816-2025` (superseded) | `AGO-1975-2025` |
| Brazil | `BRA-1800-2025` (superseded, pre-Acre) | `BRA-1909-2025` |
| Djibouti | `DJI-1886-2025` (superseded) | `FRS-1977-2025` |
| Argentina, Belize, Greece, Iraq, Romania | superseded | live equivalents |
| Belgium-Luxembourg, Cabo Verde, South Africa, Suriname | no longer exist upstream | current splits |

Nothing detected this. The `.rda` files are committed build products, and no check asserts they match what their builder emits from the current inputs, so a stale copy is indistinguishable from a fresh one until someone happens to re-run the builder.

`tests/testthat/test_polity_prefix_naming.R` (added in #382) now catches the specific symptom for these two tables — it fails if `reporting_polity_code` holds a value absent from `whep::polities`. That is a narrow backstop, not the general check.

**What the general check would look like:** a CI step that re-runs each `data-raw/*.R` builder and compares the resulting object against the committed `.rda` by content (`all.equal` on the data, not bytes — `save()` output is not reproducible). Note that re-running `harmonization_tables.R` rewrites six unrelated `.rda` files whose content is identical, so a byte comparison would be pure noise; the comparison has to be on loaded content.

Blocked on: some builders need inputs outside the repo, so this likely needs to run per-builder with the network-free ones gated first.

Issue #131, opened 2026-07-22
Revisit CLAUDE.md agent instructions: shift focus from code-style rules to validation


lbm364dl:
The way we work with agents has changed: we mostly no longer read code line by line, but instead rely on checks, tests, and CI to validate changes. CLAUDE.md is still largely written as a code-style rulebook (e.g. the 25-line function limit), which no longer matches how PRs are actually reviewed and validated.

Proposal: revisit CLAUDE.md (and any other agent instruction files) to move the emphasis from prescriptive style micro-rules towards validation:

- What checks/tests must pass and how to run them cheaply.
- What evidence a PR should carry (regression locks, coverage of new paths, back-compat tests).
- Which style rules are actually load-bearing vs. which can be dropped or relaxed (candidate: the ≤25-line function guideline).

Context: came up while reviewing #127, where the only remarks were soft style nits that don't affect correctness.

---

## Update (2026-07-24): agents must separate mechanical fixes from science/methodology decisions

A second, more important axis surfaced while triaging the open PR backlog for merge priority. Agents naturally rank PRs by **engineering review effort** — small diff + green CI + one isolated file = easy to read, low conflict risk. That axis is **orthogonal** to **scientific decision content**: whether a change embeds a methodological choice — a coefficient, an allocation rule, a conservation/mass-balance policy, an A-matrix cap or other numeric default, a fail-loud-vs-continue behavior — that the domain expert must consciously own.

A one-line, green, isolated-file diff can still hide a major science decision. CI passing proves the code runs and regressions are locked; it does **not** prove the embedded methodological choice is the one the maintainer wants. Treating "low review effort" as "safe to rubber-stamp" is the failure mode to avoid.

CLAUDE.md (and agent triage/review behavior) should therefore require, whenever recommending or reviewing PRs:

- **Classify each PR** as a **mechanical fix** (objectively correct — code was broken, now works, no defensible alternative) vs a **science/methodology decision** (defensible alternatives exist; results change depending on the choice).
- **Surface the decision content explicitly**, independent of diff size or CI status, so the maintainer approves the *choice*, not just the code.
- Extend the existing "never guess reference values, cite a verifiable source" rule to the broader class of result-affecting choices: allocation rules, conservation policies, numeric caps/defaults, and behavioral policies (e.g. abort-on-NA).

[...]

PR #521, opened 2026-08-04
Stop the test suite reaching the network (fixes #490)

### LINKED ISSUE #490: Test suite reaches the network: test_n_prov_destiny.R fails R CMD check whenever saco.csic.es is down
## Problem

One test in the suite fetches a pin over the network, so `R CMD check` fails
whenever `saco.csic.es` is unreachable — regardless of the diff under test.

MEASURED on PR #454, commit `0c2352cb`, 2026-08-04: **3 of 9 checks failed**
(`ubuntu-latest (release)`, `windows-latest (release)`, `test-coverage`), each
with the same single failure:

```
── Error ('test_n_prov_destiny.R:1015:3'): .finalize_prod_destiny combines local, import, export flows ──
Error in `.get_cache_paths(file_info, file_alias, version, e)`:
  Could not fetch "biomass_coefs" from remote source.
x No local cached copy was found either.
i Connect to the internet and try again.
  Remote host "saco.csic.es" is not reachable.
Backtrace:
  1. └─whep:::.finalize_prod_destiny(...) at test_n_prov_destiny.R:1015:3
  2.   └─whep::whep_read_file("biomass_coefs") at whep/R/n_prov_destiny.R:1578:3

[ FAIL 1 | WARN 53 | SKIP 19 | PASS 4993 ]
```

`macos-latest`, `ubuntu-latest (devel)` and `ubuntu-latest (oldrel-1)` passed the
same commit, so the outage was partial across runners. The commit under test
changed only `validation/nourishment_axis.R`, which is in `.Rbuildignore` — so the
failure was provably unrelated to the diff, and re-running the failed jobs cleared
it.

## Why it matters

A red X that has nothing to do with the change trains reviewers to ignore CI, and
it blocks merges on a third party's uptime. It also contradicts the project's own
convention that tests be self-contained (CLAUDE.md: *"`test_commodity_balance_sheet.R`
uses small self-contained fixtures (no pins, no network)"*).

`.finalize_prod_destiny()` reads the pin unconditionally at
`R/n_prov_destiny.R:1578`; the test has no fixture injection point and no
`skip_if_offline()`.

## Target state

Either give `.finalize_prod_destiny()` a `data`/`biomass_coefs` injection argument
so the test can pass a fixture (the pattern the rest of the package uses — e.g.
`.food_supply_whep_native()`'s `data$biomass_coefs %||% whep::biomass_coefs`), or
have it read the packaged dataset. Note #489 first: the pin and
`whep::biomass_coefs` are **not** the same data, so a straight swap changes
results and needs that decision made first.

Failing that, `testthat::skip_if_offline()` on the affected test would at least
stop a remote outage from turning into a check ERROR.

## Acceptance criteria

- `devtools::test()` and `R CMD check` pass with no network access.
- No test in `tests/testthat/` reaches a remote host.

Related: #136 (move inputs off the Nextcloud pins board), #245 (offline cache
fallback can't locate `latest`/frozen-blank versions), #402 (tests that pass
locally and skip on CI).

### LINKED ISSUE #513: test_n_prov_destiny.R:1015 fetches a pin over the network, failing CI on unrelated PRs
`test_n_prov_destiny.R:1015` fails intermittently in CI on unrelated PRs. I have hit it **four times today** — on #476 (ubuntu-oldrel-1), #478 (windows-release), #503 and #505 (test-coverage) — none of which touch that code. Each cost a re-run.

It is not flakiness in the usual sense. The test has an **undeclared network dependency**.

## Cause

```
── Error ('test_n_prov_destiny.R:1015:3'): .finalize_prod_destiny combines local, import, export flows ──
Error in `.get_cache_paths(file_info, file_alias, version, e)`:
  Could not fetch "biomass_coefs" from remote source.
x No local cached copy was found either.
i Connect to the internet and try again.
Remote host "saco.csic.es" is not reachable.
Backtrace:
  1. └─whep:::.finalize_prod_destiny(trade_data, codes, soil, feed_shares)
  2.   └─whep::whep_read_file("biomass_coefs")
```

The test builds `trade_data`, `codes`, `soil` and `feed_shares` as tribble fixtures and passes all four in — but `.finalize_prod_destiny()` reaches out for `biomass_coefs` itself via `whep_read_file()`. So a unit test that looks fully fixtured actually needs `saco.csic.es` to answer.

## Why it looks flaky

It passes whenever the runner has a warm pin cache or the host is up, and fails when neither holds. **Six consecutive local runs pass** (`fail=0 error=0` each time) because my cache is warm, which is exactly why it is easy to dismiss as noise.

## Cost

Every failure looks like it belongs to whatever PR is in flight. I initially attributed it to two of my own PRs and only stopped doing so after seeing the same test fail on two different platforms in two unrelated PRs. That is the real expense: it trains people to re-run red CI instead of reading it.

## Fix, two options

1. **Pass `biomass_coefs` in as a fixture**, like the other four inputs. `.finalize_prod_destiny()` appears to accept a `data` list elsewhere in the package — if it does here, this is a few lines and makes the test genuinely offline.
2. **Skip when the pin is unavailable** — `testthat::skip_if_offline()` or a `tryCatch` around the read. Cheaper, but it silently reduces coverage on exactly the runners where the network is flaky.

Option 1 is the real fix; option 2 stops the bleeding.

Worth checking whether other tests in the suite have the same shape — a fully-fixtured-looking test whose function under test reads a pin internally. `whep_read_file` appearing anywhere under `tests/` or in a function called from a test with no `data` argument is the signature.

lbm364dl:
Fixes #490. Closes #513 (duplicate).

## What was happening

`test_n_prov_destiny.R:1015` looks fully fixtured — it builds `trade_data`,
`codes`, `soil` and `feed_shares` as tribbles and passes all four in — but
`.finalize_prod_destiny()` reached out for **two** pins of its own:

```r
biomass_coefs         <- whep_read_file("biomass_coefs")           # line 1577
pie_full_destinies_fm <- whep_read_file("pie_full_destinies_fm")   # line 1594
```

Both issues named only `biomass_coefs`, because that is the one that happened to
be fetched first and abort. `pie_full_destinies_fm` was a second, unreported
network dependency in the same function. There is no pin cache step in
`R-CMD-check.yaml`, so **every CI runner is cold** and both reads went to
`saco.csic.es` on every run. Seven observed failures across the two issues
(#454 ×3, #476, #478, #503, #505).

## Root cause: the reads only fed dead code

Neither pin was used for anything. Both existed solely to build
`shares_import_wide`, which was passed to `.split_import_consumption()` as a
third argument that **the function body never references**:

```r
.split_import_consumption <- function(
  local_vs_import,
  feed_share_rum_mono,
  shares_import_wide      # declared, never used; not even documented in roxygen
) {
```

Three independent confirmations that it is vestigial rather than something I
overlooked:

1. The roxygen block above it documents only two `@param`s.
2. `test_n_prov_destiny.R:787` already calls it with **two** arguments and
   passes — R's lazy evaluation never forces the missing third.
3. `git log -S shares_import_wide` returns exactly one commit (`6e98f4fd`,
   "Updates from grafs_plot changes") — it was introduced already unused.

So the fix is to delete the dead block rather than to fixture it. This is
**bit-for-bit output-preserving**: `pie_imports_n` and `shares_import_wide` were
computed eagerly, handed to a function that ignored them, and discarded.
`shares_import` (from `.calculate_consumption_shares()`, still live at what is
now line 1592) is a *different* variable and is untouched — worth stating
explicitly, since the two names differ by one word.

## Changes

- `R/n_prov_destiny.R`
  - `.finalize_prod_destiny()`: drop the `biomass_coefs` read, the ~60-line
    `pie_imports_n` pipeline and the `shares_import_wide` pivot. The function no
    longer touches the network, and the real pipeline makes two fewer remote
    fetches per run.
  - `.split_import_consumption()`: drop the unused third parameter.
- `tests/testthat/test_soil_carbon_inputs.R`
  - `.sci_read_manure` turnkey test: wrap in the repo's existing
    `tryCatch(...) + skip_if(is.null(...), "... pins unavailable")` idiom (as at
    `test_select_best_source_aggregates.R:91`). It already had `skip_on_ci()`, so
    it was never a CI blocker, but it did fail a local offline `devtools::test()`,
    which acceptance criterion 1 asks for. It is a real-data smoke test over
    `build_primary_production()`, so it cannot be made genuinely offline — the
    right treatment is a skip, not a fixture.
  - One deliberate tightening of that idiom: the handler **only** swallows errors
    matching `"Could not fetch|not reachable"` and re-`stop()`s anything else. A
    bare `error = function(e) NULL` would also convert a genuine break in
    `.sci_read_manure()` into a silent skip, which is how a network guard turns
    into lost coverage.

No test-side fixture for `.finalize_prod_destiny()` is needed: with the dead code
gone, the four tribbles the test already passes are the function's entire input.

## Verification

Both acceptance criteria were checked by **reproducing a cold CI runner
locally** — empty pins cache plus a dead HTTP proxy, so any remote read aborts:

```bash
XDG_CACHE_HOME=<empty dir> http_proxy=http://127.0.0.1:9 https_proxy=http://127.0.0.1:9 \
  Rscript --vanilla -e 'pkgload::load_all(); testthat::test_local()'
```

| | failures | skips |
|---|---|---|
| offline, before | **2** | 13 |
| offline, after | **0** | 14 |
| normal (warm cache), after | **0** | 8 |

Before, that run reproduced the exact CI error at `test_n_prov_destiny.R:1015:3`
and additionally surfaced `test_soil_carbon_inputs.R:583:3` — which is the audit
#513 asked for ("whether other tests in the suite have the same shape"). Those
were the only two. The other skips are env-var-gated local-raster tests, and
`read_example` in `test_input_files.R` resolves to a **local** board
(`.get_local_board()`), so it never touches the network.

The warm-cache run confirms the guarded test still genuinely executes and passes
when the pins *are* reachable — it is skipped only when they are not, not
disabled.

Other gates, all from the correct working directory:

- `air format .` — clean.
- `devtools::document()` — no `man/` change (both edited functions are `@noRd`).
- `lintr::lint_package()` with the four CI-disabled linters — **0 lints**.
- `rcmdcheck::rcmdcheck()` — **0 errors, 0 warnings, 2 notes**, and the notes are
  **byte-identical to an `origin/main` baseline** checked the same way, so this
  branch introduces none. (Both pre-existing: the `.sci_sum_components` NSE
  bindings and "All declared Imports should be used".) Checked against a clean
  `git archive` export, since an in-place check dies on the 37 GB gitignored
  `LPJmL_inputs/`.
- No `utils::globalVariables()` change needed: every symbol the deleted block
  used (`value_fm`, `prod_type`, `value_n`, `share`, `total`, `Value_destiny`) is
  still referenced elsewhere in `R/`, so no entry became stale.

## Note on #489 — which `biomass_coefs` is canonical

#490 flagged that this fix might depend on the #489 decision ("the pin and
`whep::biomass_coefs` are not the same data, so a straight swap changes results").
**It turns out not to, because the read was dead** — nothing here swaps a data
source, and no result changes.

Since the question was raised, I measured it anyway, comparing both against the
current upstream sheet (`afsetools/inst/extdata/Biomass_coefs.xlsx`, `Coefs`,
`skip = 1`, mtime 2026-06-17). Numeric columns compared with a relative
tolerance of 1e-6, so xlsx-vs-CSV float round-tripping is not miscounted as a
difference:

| column | pin vs pkg | **xlsx vs pkg** | xlsx vs pin |
|---|--:|--:|--:|
| `Category` | 31 | **0** | 31 |
| `Product_kgDM_kgFM` | 2 | **0** | 2 |
| `Residue_kgDM_kgFM` | 2 | **0** | 2 |
| `GE_residue_MJ_kg` | 3 | **0** | 3 |
| `Edible_portion` | 10 | **0** | 10 |
| `N_kgN_kgFM` | 4 | **0** | 4 |
| `Product_kgN_kgDM` | 2 | **0** | 2 |
| `Product_kgC_kgDM` | 4 | **0** | 4 |
| `Residue_kgN_kgDM` | 17 | **0** | 17 |
| `Residue_humified_kgC_kgC` | 18 | **0** | 18 |
| `Root_humified_kgC_kgC` | 21 | **0** | 21 |
| `Equiv` | 12 | 2 | 12 |

(cells differing, n = 421 rows)

**On every contested coefficient column, upstream agrees exactly with
`whep::biomass_coefs` and disagrees with the pin.** The package data is
canonical; the pin is a stale 2025-07-28 narrowed export. That supports the
direction already sketched on #489 — point the remaining
`whep_read_file("biomass_coefs")` call sites at `whep::biomass_coefs` and retire
the pin entry. **Deliberately not done in this PR**: those call sites
(`n_prov_destiny.R:42`, `:1181`, `Typologies_Josette.R:112`) are live, so
swapping them *would* change `create_n_prov_destiny()` output. That is a
methodological call for @eduaguilera, and it belongs in #489 rather than in a CI
fix. This PR removes one of the four call sites for free, as a side effect of
deleting dead code.

Two side findings for #489/#384, measured the same way:

- The packaged `.rda` is itself **stale against the current xlsx in 5
  below-ground columns** — `BG_Biomass_kgDM_ha` (92 cells), `Rhizodeposits_N_kgN_kgRootN`
  (25), `Root_Shoot_ratio` (24), `Root_kgC_kgDM` (23),
  `Rhizodeposits_mass_kgC_kgDM` (23). The pin and the `.rda` agree on all five,
  so upstream changed them after both were built. Independent of the
  canonicity question, and more evidence for the #384 stale-`.rda` gate.
- The empty `Edible_*` block is **upstream's own**, not a WHEP addition:
  `Edible_N_kgFM`, `NonEdible_kgN_kgFM` et al. are 0/421 non-NA in *both* the
  xlsx and the `.rda`. Relevant to #361.

## One thing to confirm

The deleted block computed per-item import destiny shares from the PIE table
(`share_Food` / `share_Other_uses` / `share_Feed`). It has never been wired into
anything, but it may be an **unfinished intent** rather than an accident — if
imports were meant to be split by PIE-derived destiny shares instead of by the
local demand gap that `.split_import_consumption()` currently uses, that is a
modelling change worth its own issue. I have deleted it as dead code and left
this note instead of guessing; it is recoverable from `6e98f4fd`. Happy to open a
follow-up issue if that was the plan.

## Checklist

- [x] `air format .`
- [x] `devtools::document()`
- [x] `rcmdcheck::rcmdcheck()` — no new errors/warnings/notes vs `origin/main`
- [x] `lintr::lint_package()` — 0 lints
- [x] Full suite green on a warm cache
- [x] Full suite green with **no network access** (the point of the PR)
- [x] No `man/` or `_pkgdown.yml` change needed (both edited functions are
      `@keywords internal` / `@noRd`)

### DIFF
--- a/.github/workflows/offline-tests.yaml
+++ b/.github/workflows/offline-tests.yaml
@@ -7,15 +7,27 @@
 # green exactly when the host happens to answer.
 #
 # This job removes the luck. It runs the suite with:
-#   1. an empty pins cache, so nothing can be served from a warm copy, and
-#   2. http(s)_proxy pointed at a dead port, so every remote read aborts.
+#   1. an empty pins cache, so nothing can be served from a warm copy,
+#   2. http(s)_proxy pointed at a dead port, so every remote read aborts, and
+#   3. CI unset, so `skip_on_ci()` tests run here too.
 # Any test that reaches out fails here, deterministically, on every run.
 #
-# The proxy vars are set on the test step ONLY -- the setup steps above it need
+# (3) is not incidental. A `skip_on_ci()` test is invisible to every GitHub
+# Actions job, this one included, so the guard would be blind to exactly the
+# tests most likely to reach the network -- the ones deliberately kept out of
+# CI because they are slow and real-data backed. That blind spot is not
+# hypothetical: it let a turnkey test through that ERRORed the r-universe
+# check for a week (r-universe does not set CI, so it ran the test for real,
+# built the full 1850-2023 series and died on a 16 GB allocation).
+#
+# Unsetting CI is safe because every remaining real-data test carries its own
+# local-file or env-var guard and skips on that instead.
+#
+# These env vars are set on the test step ONLY -- the setup steps above it need
 # real network access to install dependencies.
 #
 # Reproduce locally:
-#   XDG_CACHE_HOME=$(mktemp -d) http_proxy=http://127.0.0.1:9 \
+#   CI=false XDG_CACHE_HOME=$(mktemp -d) http_proxy=http://127.0.0.1:9 \
 #     https_proxy=http://127.0.0.1:9 Rscript -e 'devtools::test()'
 #
 # If this job fails alone, the fix is to give the offending test an offline
@@ -63,6 +75,9 @@ jobs:
           XDG_CACHE_HOME: ${{ runner.temp }}/empty-cache
           http_proxy: http://127.0.0.1:9
           https_proxy: http://127.0.0.1:9
+          # Runs the `skip_on_ci()` tests here as well -- see the header.
+          # testthat treats any non-"true" value as "not on CI".
+          CI: "false"
         run: |
           mkdir -p "$XDG_CACHE_HOME"
           Rscript -e 'devtools::test(stop_on_failure = TRUE)'
--- a/.github/workflows/offline-tests.yaml
+++ b/.github/workflows/offline-tests.yaml
@@ -1,0 +1,69 @@
+# Guards the invariant from #490: no test may depend on a remote host being up.
+#
+# A test that quietly fetches a pin passes on a warm cache and fails when the
+# host is down, so it shows up as an unrelated red X on whatever PR is in flight
+# and trains reviewers to re-run CI instead of reading it. The other jobs cannot
+# catch this: they have no pin cache, so they hit the network for real and are
+# green exactly when the host happens to answer.
+#
+# This job removes the luck. It runs the suite with:
+#   1. an empty pins cache, so nothing can be served from a warm copy, and
+#   2. http(s)_proxy pointed at a dead port, so every remote read aborts.
+# Any test that reaches out fails here, deterministically, on every run.
+#
+# The proxy vars are set on the test step ONLY -- the setup steps above it need
+# real network access to install dependencies.
+#
+# Reproduce locally:
+#   XDG_CACHE_HOME=$(mktemp -d) http_proxy=http://127.0.0.1:9 \
+#     https_proxy=http://127.0.0.1:9 Rscript -e 'devtools::test()'
+#
+# If this job fails alone, the fix is to give the offending test an offline
+# fixture (see CLAUDE.md), not to skip it and not to relax this job.
+
+on:
+  push:
+    branches: [main, master]
+  pull_request:
+
+name: offline-tests.yaml
+
+permissions: read-all
+
+# Cancel a superseded run on the same branch/PR so stale runs do not pile up.
+concurrency:
+  group: offline-tests-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  offline-tests:
+    runs-on: ubuntu-latest
+    # Backstop against a hung test. The suite is a few minutes; the headroom is
+    # for a cold dependency install of the geo stack.
+    timeout-minutes: 30
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: r-lib/actions/setup-r@v2
+        with:
+          use-public-rspm: true
+
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          extra-packages: any::devtools
+          needs: check
+
+      - name: Run the test suite with no network access
+        env:
+          # rappdirs::user_cache_dir("pins") resolves to $XDG_CACHE_HOME/pins on
+          # Linux, so an empty XDG_CACHE_HOME is an empty pin cache.
+          XDG_CACHE_HOME: ${{ runner.temp }}/empty-cache
+          http_proxy: http://127.0.0.1:9
+          https_proxy: http://127.0.0.1:9
+        run: |
+          mkdir -p "$XDG_CACHE_HOME"
+          Rscript -e 'devtools::test(stop_on_failure = TRUE)'
+        shell: bash
--- a/R/n_prov_destiny.R
+++ b/R/n_prov_destiny.R
@@ -1456,8 +1456,7 @@ create_n_nat_destiny <- function(example = FALSE) {
 #' @noRd
 .split_import_consumption <- function(
   local_vs_import,
-  feed_share_rum_mono,
-  shares_import_wide
+  feed_share_rum_mono
 ) {
   local_vs_import |>
     dplyr::left_join(
@@ -1575,7 +1574,6 @@ create_n_nat_destiny <- function(example = FALSE) {
   n_soil_inputs,
   feed_share_rum_mono
 ) {
-  biomass_coefs <- whep_read_file("biomass_coefs")
   grafs_prod_destiny_final <- .prep_final_ds(
     grafs_prod_item_trade,
     codes_coefs_items_full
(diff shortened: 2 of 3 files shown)


lbm364dl:
Re-checked against the reworked CLAUDE.md (#131), which lands two requirements this PR had only partly met.

## Classification: `mechanical`

No methodological decision. The deleted code's result was discarded by its only
consumer, so there is no alternative that could have been chosen differently. The
one science decision in the neighbourhood — which `biomass_coefs` table is
canonical — is deliberately **not** in this PR; it is now filed as #523.

## "No published value changes" — checked, not assumed

CLAUDE.md is explicit that this claim "must be checked, not assumed". My original
description argued it from the code (the value was computed, passed, ignored).
That is sound but it is an argument, not evidence. So I ran it:

`create_n_prov_destiny()` executed in full on both trees, sorted, compared.

| | `origin/main` | this branch |
|---|---|---|
| rows x cols | 4,304,529 x 8 | 4,304,529 x 8 |
| `sum(mg_n)` | 397675357.20889425278 | 397675357.20889425278 |

```
identical(before, after)          TRUE
all.equal(before, after, tol = 0) TRUE
max abs diff over numeric cols    0
```

Bit-for-bit identical across all 4.3 million rows. Not "no meaningful change" —
no change at all.

## Evidence for the fix itself

Also per CLAUDE.md ("a test that fails before the fix and passes after it"): the
failing-then-passing test is the existing `test_n_prov_destiny.R:1015` under the
cold-runner conditions in the PR body — 2 offline failures before, 0 after. It
fails on `main` offline and passes here, which is the same relationship a
purpose-written regression test would have. What this PR does *not* add is a
permanent guard that would fail if someone reintroduces a pin read inside a
function under test; the cold-cache recipe has to be run deliberately. If you want
that guard standing, the cheap version is a test asserting no `whep_read_file()`
call is reachable from the unit-test paths — say the word and I will add it here
rather than leaving it to be rediscovered.

## Follow-ups filed

- #522 — whether imports should be split by PIE-derived destiny shares. This is
  the intent the deleted block may have represented; filed so removing dead code
  did not silently erase a plan. `needs-expert`.
- #523 — repoint the three remaining `biomass_coefs` pin reads at the package data
  and retire the pin. Includes the full column-by-column measurement showing the
  package data is canonical. `needs-expert`, because it changes outputs.
- #524 — the packaged `biomass_coefs` is itself stale against its upstream sheet
  in 5 below-ground columns. Orthogonal to #523.
- #525 — two leftovers from the same `6e98f4fd` refactor: a
  `whep_read_file("")` placeholder default, and `.calculate_trade()` roxygen
  documenting three parameters it does not have. `mechanical`, good first issue.

Issue #523, opened 2026-08-04
Route remaining biomass_coefs reads at whep::biomass_coefs and retire the stale pin


lbm364dl:
Actionable follow-up to #489, which established that `biomass_coefs` has two live sources that disagree. This issue is the **fix**: pick the package data, repoint the remaining callers, delete the pin entry.

Verified 2026-08-04. #521 already removed one of the four pin call sites (it was dead code), so three remain.

## Which source is correct: the package data

Both copies were compared against the current upstream sheet
(`afsetools/inst/extdata/Biomass_coefs.xlsx`, sheet `Coefs`, `skip = 1`, mtime
2026-06-17), which is the origin of both. Numeric columns compared with a
**relative tolerance of 1e-6** — see the gotcha at the end, it matters.

Cells differing, out of 421 rows:

| column | pin vs pkg | **xlsx vs pkg** | xlsx vs pin |
|---|--:|--:|--:|
| `Category` | 31 | **0** | 31 |
| `Product_kgDM_kgFM` | 2 | **0** | 2 |
| `Residue_kgDM_kgFM` | 2 | **0** | 2 |
| `GE_residue_MJ_kg` | 3 | **0** | 3 |
| `Edible_portion` | 10 | **0** | 10 |
| `N_kgN_kgFM` | 4 | **0** | 4 |
| `Product_kgN_kgDM` | 2 | **0** | 2 |
| `Product_kgC_kgDM` | 4 | **0** | 4 |
| `Residue_kgN_kgDM` | 17 | **0** | 17 |
| `Residue_humified_kgC_kgC` | 18 | **0** | 18 |
| `Root_humified_kgC_kgC` | 21 | **0** | 21 |
| `Equiv` | 12 | 2 | 12 |

**On 10 of the 11 contested coefficient columns, upstream agrees exactly with
`whep::biomass_coefs` and disagrees with the pin.** `Equiv` is the only column
where the package data is not a perfect match, and even there it is 2 cells
against the pin's 12.

So `whep::biomass_coefs` (421 x 68) is canonical; the pin (421 x 41, version
`20250728T082553Z-f2fac`, dated 2025-07-28) is a stale narrowed export. Its 41
columns are a strict subset of the 68, so nothing is lost by dropping it.

## Work

1. Repoint the three remaining pin reads at the package data:
   - `R/n_prov_destiny.R:42` (in `create_n_prov_destiny()`)
[...]

PR #528, opened 2026-08-04
fix(n_prov_destiny): drop empty-alias default and stale roxygen

### LINKED ISSUE #525: n_prov_destiny.R: placeholder whep_read_file("") default and stale roxygen on .calculate_trade()
Two small leftovers in `R/n_prov_destiny.R` from the same refactor that produced
the dead code removed in #521 (`6e98f4fd`, "Updates from grafs_plot changes").
Both verified against current `main` on 2026-08-04. Neither affects output today;
both are traps for the next reader.

## 1. A default argument that reads the pin alias `""`

`R/n_prov_destiny.R:1179`, in `.convert_to_items_n()`:

```r
.convert_to_items_n <- function(
  grafs_prod_item_combined = whep_read_file(""),      # <-- empty alias
  codes_coefs_items_full = whep_read_file("codes_coefs_items_full"),
  biomass_coefs = whep_read_file("biomass_coefs")
) {
```

The first default is a placeholder that can never work: `whep_read_file("")`
aborts with `There is no file entry`. It is harmless in practice only because the
sole caller (`create_n_prov_destiny()`, line 87) always pipes the argument in. But
anyone calling the helper directly, or reading it to learn the contract, gets a
default that looks meaningful and is not.

Fix: drop the default so the argument is simply required.

## 2. Roxygen documenting three parameters that do not exist

`R/n_prov_destiny.R:1240-1253`. `.calculate_trade()` takes **one** argument:

```r
#' @param grafs_prod_item_n A dataframe with N values (MgN) by destiny.
#' @param pie_full_destinies_fm A data frame with destiny data.
#' @param biomass_coefs A data frame with biomass coefficients.
#' @param codes_coefs_items_full A lookup table with coefficients.
...
.calculate_trade <- function(grafs_prod_item_n) {
```

The last three are stale. The `@description` is also stale in the same way — it
promises a toggle the function does not have: *"National scaling can be activated
... It should be deactivated for provincial analysis."* There is no such argument.

`R CMD check` does not catch either, because the function is `@noRd`.

Fix: delete the three `@param` lines and correct the description.

Both are `mechanical` — no defensible alternative, no output change. Good
first-issue material.

Related: #521 (removed the dead code from the same refactor), #522 (the modelling
question that refactor left open).

RawNuke:
Fixes #525

Two small leftovers from the refactor in `6e98f4fd` (#521), both verified against current main on 2026-08-04.

## 1. Placeholder default `whep_read_file("")`

`.convert_to_items_n()` defaulted `grafs_prod_item_combined` to `whep_read_file("")`, which always aborts with "There is no file entry". Harmless in practice because the sole caller pipes the argument in, but any direct caller or reader gets a default that looks meaningful and is not. The default is dropped; the argument is now required.

## 2. Stale roxygen on `.calculate_trade()`

The roxygen block documented three parameters (`pie_full_destinies_fm`, `biomass_coefs`, `codes_coefs_items_full`) that the function does not take, and the `@description` promised a national-scaling toggle that does not exist. Both corrected.

## Verification

- `Rscript -e 'parse(file="R/n_prov_destiny.R")'` — syntax OK.
- Signature-vs-roxygen consistency verified: `.calculate_trade` documents only its real argument; `.convert_to_items_n` has no empty-alias default.
- No output change; no other file touched.

### DIFF
--- a/R/n_prov_destiny.R
+++ b/R/n_prov_destiny.R
@@ -1176,7 +1176,7 @@ create_n_nat_destiny <- function(example = FALSE) {
 #' @keywords internal
 #' @noRd
 .convert_to_items_n <- function(
-  grafs_prod_item_combined = whep_read_file(""),
+  grafs_prod_item_combined,
   codes_coefs_items_full = whep_read_file("codes_coefs_items_full"),
   biomass_coefs = whep_read_file("biomass_coefs")
 ) {
@@ -1240,13 +1240,9 @@ create_n_nat_destiny <- function(example = FALSE) {
 
 #' @title Consumption and Trade
 #' @description Calculation of consumption by destiny and trade
-#' (export, import). National scaling can be activated, for analysis for whole
-#' Spain. It should be deactivated for provincial analysis
+#' (export, import).
 #'
 #' @param grafs_prod_item_n A dataframe with N values (MgN) by destiny.
-#' @param pie_full_destinies_fm A data frame with destiny data.
-#' @param biomass_coefs A data frame with biomass coefficients.
-#' @param codes_coefs_items_full A lookup table with coefficients.
 #'
 #' @return A dataframe with consumption, exports, and imports in MgN.
 #' @keywords internal


lbm364dl (approved):
Both claims verified against current `main` rather than taken on trust.

**1. Empty-alias default.** `R/n_prov_destiny.R:1179` is indeed
`whep_read_file("")`, and it does abort (`There is no file entry with alias`).
Dropping the default is safe: the sole production caller at line 87 supplies the
argument as the piped LHS, and both tests (`test_n_prov_destiny.R:530`, `:555`)
pass all three positionally. Nothing reaches the default.

**2. Stale roxygen.** `.calculate_trade(grafs_prod_item_n)` takes one argument
and the block documented four, plus a `@description` promising a national-scaling
toggle that does not exist. `devtools::document()` produces no `man/` or
`NAMESPACE` diff, confirming `@noRd` leaves nothing to regenerate.

## Verification

Note that **this PR's CI never ran** (see below), so the following was done
locally on the PR merged into current `main`:

- merges cleanly into `main` (post-#521)
- `air format .` — no changes
- `lintr` on `R/n_prov_destiny.R` — no lints
- `devtools::document()` — no diff
- full suite, offline (`CI=false`, dead proxy, empty pin cache):
  `FAIL 0 | WARN 7 | SKIP 13 | PASS 5402`, 102 s

## The red check is not your fault

`format-suggest` failed at *checkout*, not on formatting: it runs on
`pull_request_target`, and `actions/checkout@v4` now refuses to check out fork
head code in that trusted context by default. That breaks the workflow for every
fork PR, and is being handled separately.

`R-CMD-check`, `lint`, `test-coverage` and `pkgdown` are all sitting in
`action_required` — fork PRs need a maintainer to approve workflow runs.
`offline-tests` did not run because it did not exist on this branch yet (it
landed with #521).

Thanks — this is a clean, well-scoped fix, and leaving the two neighbouring
`whep_read_file()` defaults alone was the right call (they are #523's territory).

lbm364dl:
Thanks for your contribution @RawNuke!

Issue #548, opened 2026-08-05
The no-LUH2-land warning misses the zero-land class: 9 areas have a land row of exactly 0 and lose their back-cast just as silently


lbm364dl:
`.warn_unmatched_land()` in `R/build_production.R` exists so that a silent loss
of back-cast production becomes visible. It catches areas with **no** LUH2 land
row. It does not catch areas with a LUH2 land row whose value is **zero**, which
fail identically: `fill_proxy_growth()` cannot grow a series from a zero proxy,
so their pre-1962 production is dropped just the same, and nothing says so.

Found while fixing #408 (the dissolved-federation class of the same warning).

## Measured on `origin/main` (60fc6610)

LUH2 land read for 1849-1965 via `.read_land_areas()`, summed over
`c3ann c3per c4ann c4per c3nfx` and `pastr range`:

```
areas with LUH2 land rows but ZERO pre-1962 cropland:  9
areas with ZERO pre-1962 cropland AND pasture:         9
```

Confirmed on a real `.read_production(1850, 1965)` build: **all 9 have zero
pre-1961 rows in the output**, and the warning names none of them.

Their combined share of 1961-62 FAOSTAT production tonnage is **0.0171%**, so
this is a diagnostic gap rather than a material one. The largest are:

| area_code | area | 1961-62 tonnes |
|---|---|---|
| 96 | China, Hong Kong SAR | 1,263,121 |
| 70 | French Polynesia | 526,488 |
| 83 | Kiribati | 188,696 |
| 196 | Seychelles | 95,631 |
| 132 | Maldives | 86,814 |
| 227 | Tuvalu | 11,840 |
| 148 | Nauru | 4,265 |

(The remaining two of the nine report no production tonnage in 1961-62 at all.)

Every one is a small island state or a city territory, so the zero looks like a
genuine property of LUH2 at 0.5 degrees rather than a bug — the cells carry no
crop or pasture fraction. The fix is therefore not to invent land, it is for the
warning to report the class, so a reader knows the series stops at 1961 by data
limitation and not by accident.

## Suggested fix

Extend `.warn_unmatched_land()` to a second bucket: areas that match a LUH2 land
row but whose proxy is zero across every pre-1962 year, reported separately from
the no-match bucket so the two causes stay distinguishable. Cheap, offline, and
testable with a `tribble()` fixture — the existing
`.extend_historical warns about areas with no LUH2 land match` test is the model.

Part of the polity migration epic #458.

PR #549, opened 2026-08-05
Reach a dissolved federation's LUH2 land through its successors, so 14.3% of early production stops vanishing

### LINKED ISSUE #408: 11.9% of 1961-62 production is not back-cast: the LUH2 bridge is keyed by modern ISO3 and four of the areas are dissolved federations
Found on #382 while working through the warning sites that touch area identity. Measured and pinned there; this is the piece that needs a decision.

## The gap

Pre-1962 production is back-cast onto LUH2 land, and the bridge from an area to its land is the area's **ISO3**. LUH2 is keyed by *modern* ISO3 while this project periodizes territory, so the bridge cannot reach a territory with no modern code. The build already warns:

```
Historical extension: {n} areas have no LUH2 land match; their pre-1962 production is not back-cast.
```

Measured on 1961–62, the first year the extension is asked for:

| | |
|---|---|
| production areas | 195 |
| LUH2-reachable | 187 |
| **not reachable** | **8** |
| **their share of production value** | **11.91%** |

8 of 195 areas sounds negligible. 11.91% of value is not.

## Two classes, only one of them documented

**Dissolved federations — the bulk of the 11.91%:**

| area | | why |
|---|---|---|
| 15 | Belgium-Luxembourg | no modern ISO3 |
| 51 | Czechoslovakia | no modern ISO3 |
| 228 | USSR | no modern ISO3 |
| 248 | Yugoslav SFR | no modern ISO3 |

**French overseas departments** — LUH2 folds *départements* into France, so their land is present but not separable: 69 French Guiana, 87 Guadeloupe, 135 Martinique, 182 Réunion. This class is already noted in the integration vignette.

## A shortcut that does not work

Worth recording so nobody re-derives it. The polities database has a `successor` field, so reconstructing a federation's land as the union of its successors' LUH2 land looks available. **It is not:** `successor` points at the next *period of the same family* — `F228-1945-1991` gives `F228-1991-1992`, not the fifteen republics. Measured:

```
area 15   Belgium-Luxembourg   successors=0   their iso3 in LUH2: 0/0
area 51   Czechoslovakia       successors=1   their iso3 in LUH2: 0/0
area 228  USSR                 successors=1   their iso3 in LUH2: 1/1
area 248  Yugoslav SFR         successors=2   their iso3 in LUH2: 0/0
```

## What would actually work

Go through the **polygons** rather than any code field. Each federation has a polygon in the polities GeoPackage, and LUH2 land is attributable to it by spatial intersection with the successor states' extents — which is the same dasymetric reallocation `build_constant_territory_series()` already implements in this package, pointed the other way.

That is a real piece of work rather than a mechanical fix, and it changes back-cast values for ~12% of early production, so it is a modelling decision. Options as I see them:

1. **Leave it, now that the size is pinned.** The warning fires and `test_luh2_backcast_reach.R` asserts the 8 areas and the ~12% share, so the gap cannot grow or shrink unnoticed.
2. Reconstruct federation land by spatial intersection with successor extents. Most faithful, most work.
3. Attribute the French departments to France's LUH2 land in proportion to their polygon areas — cheap, and covers the smaller class only.

I have not chosen, because 12% of a back-cast series is not something to change as a side effect of a warning sweep.

lbm364dl:
## What was wrong

`.extend_historical()` back-casts pre-1962 production from LUH2 land use, and the
bridge from an area to its land is `.read_land_areas()`'s ISO3 join
(`R/build_production.R:415-423`). LUH2 is keyed on **present-day** ISO3 while this
project periodizes territory, so an area whose territory is a dissolved
federation has no LUH2 land row, `fill_proxy_growth()` has nothing to grow, and
its pre-1962 series is dropped. The build already warned; the warning was ignored.

At base `60fc6610`, `.read_land_areas(years = 1849:1965)` against the real
`luh2-areas` pin plus `.read_fao_crop_liv(years = 1961:1962)`:

```
production areas 1961-62:  166
not LUH2-reachable:          4
Production tonnes 1961-62: total 12,707,597,516
                     unreachable 1,811,810,418  = 14.26%
Area harvested (ha) 1961-62 unreachable        = 16.49%
```

| area | polity | share of 1961-62 tonnes |
|---|---|---|
| 228 USSR | `F228-1945-1991` | **12.19%** |
| 51 Czechoslovakia | `F51-1947-1993` | 0.86% |
| 248 Yugoslav SFR | `F248-1991-1992` | 0.73% |
| 15 Belgium-Luxembourg | `BLX-1850-1999` | 0.48% |

And on a real `.read_production(1850, 1965)` build at base, the warning fires with
exactly those four, and the three big ones have **zero** pre-1961 rows in the
output while carrying 704.7 / 69.4 / 28.6 Mt in 1961 — a hard discontinuity at the
splice year, not a gentle taper.

### Two things the issue got wrong, corrected here

**The French overseas departments are not a live defect.** #408 lists 69 French
Guiana, 87 Guadeloupe, 135 Martinique and 182 Réunion as a second unreachable
class, for 8 unmatched areas out of 195. On current `main` they are not: all four
fold into `polity_area_code` 999 (Rest of World), and 999 *is* LUH2-reachable
because 62 area codes with an ISO3 land record share that bucket. The real
reproduction is **4 of 166 areas, 14.26%** of 1961-62 tonnes, not 8 of 195 /
11.91%. All eight-area arithmetic in the issue is stale.

**The `successor` shortcut does work.** #408 records it as a dead end ("`successor`
points at the next *period of the same family* — `F228-1945-1991` gives
`F228-1991-1992`, not the fifteen republics", measured as 1/1 in LUH2). On current
`whep::polities` that is false:

```
F51-1947-1993   -> CZE-1993-2025; SVK-1993-2025
F228-1945-1991  -> RUS UKR BLR EST LVA LTU MDA ARM AZE GEO KAZ KGZ TJK TKM UZB   (15)
F248-1991-1992  -> SVN-1992-2025; HRV-1992-2025; BIH-1992-2025; SCG-1992-2006; MKD-1991-2025
```

Walking it **transitively** — which is what the issue's measurement was missing —
resolves 3 of the 4 completely, because Yugoslavia reaches Serbia and Montenegro
only through `SCG-1992-2006`, three hops down. Against LUH2's 230-code ISO3
vocabulary:

```
F51-1947-1993    2  CZE SVK
F228-1945-1991  15  ARM AZE BLR EST GEO KAZ KGZ LTU LVA MDA RUS TJK TKM UKR UZB
F248-1991-1992   6  BIH HRV MKD MNE SRB SVN
BLX-1850-1999    0
```

So the expensive option 2 in the issue (polygon intersection with successor
extents) is not needed.

**Independent check that the union is the right territory.** Reconstructed 1961
cropland vs. what FAOSTAT itself reports for the same federation in the
`faostat-landuse` pin:

| area | LUH2 successor union | FAOSTAT "Cropland" 1961 | FAOSTAT "Arable land" 1961 | vs cropland |
|---|---|---|---|---|
| 51 Czechoslovakia | 5.414 Mha | 5.418 Mha | 5.330 Mha | **-0.07%** |
| 228 USSR | 240.41 Mha | 239.80 Mha | 235.40 Mha | **+0.26%** |
| 248 Yugoslav SFR | 7.341 Mha | 8.382 Mha | 7.691 Mha | -12.4% |

Two of three land within a quarter of a percent of FAOSTAT's own figure, which is
about as strong a confirmation as this reconstruction can get. The Yugoslav
shortfall is traceable and not a bug in the walk: **Kosovo is absent from LUH2
entirely** (`KOS-2008-2025` is in the successor chain, but no `XKX`/`KOS` code
exists in LUH2's 230), and LUH2 gives North Macedonia only 0.035 Mha of 1961
cropland, which is the same small figure whep already uses for area 154 in its own
right. Since `fill_proxy_growth()` consumes only the year-on-year *ratio*, a
constant level shortfall is largely inert.

## What I changed

**`R/polities.R`** — `.successor_iso3_map()` resolves a polity to the set of ISO3
codes a caller can actually reach, by walking the published `successor` relation
transitively and stopping each branch the moment it lands inside the caller's
vocabulary (so `RUS-1991-2014` terminates at `RUS` and does not expand into
`RUS-2014-2025`). Plus `.polity_successor_edges()` / `.polity_iso3_lookup()`.

**`R/build_production.R`** — `build_primary_production(federation_land =)`, threaded
through `.read_production()` into `.extend_historical()`:

* `"none"` — **default, current published behaviour.** Federations stay unmatched.
* `"successor_union"` — `.add_federation_land_rows()` adds, per unreachable bucket,
  a land series equal to the sum of its successors' LUH2 land.

Two design points worth stating:

* **The augmented table is local to `.extend_historical()`.** `.build_grassland()`
  reads the same `land_areas` in `.read_production()`; a federation grassland row
  there would double-count against its successors' own rows across all of
  1850-2023. There is a test pinning that the bridge does not mutate its input by
  reference, because data.table makes that a live hazard.
* **Only the year-on-year ratio of the proxy is used** by `fill_proxy_growth()`, so
  the level mismatch between a federation and its successors' present-day extent
  does not enter the result. LUH2 already reports every other area on modern
  boundaries at every year, so this is the same kind of aggregate, not a new one.

The warning now pluralises correctly (`1 area have` -> `1 area has`) and names the
argument that fixes the federation class.

## How I verified

Gates, all run, not predicted:

* `air format .` — binary, clean.
* `devtools::document()` — `man/build_primary_production.Rd` regenerated and committed.
* `lintr::lint_package(...)` with the four CI-disabled linters — **No lints found.**
* `devtools::test()` — full suite: **0 failed, 0 error, 5444 passed, 8 skipped**.
* pkgdown reference coverage — `man/*.Rd` minus `_pkgdown.yml` is empty (no new topic).
* No non-ASCII in any added line; `n_successors` added to `utils::globalVariables()`.

**The assertion count that fails without the change:** 11 new `test_that` blocks
(3 in `test_polities.R`, 8 in `test_build_production.R`). Ran them against a clean
`git archive origin/main` checkout with only the two test files swapped in:

```
failed: 0   error: 11   passed: 140
```

All 11 error at base — `.successor_iso3_map` / `.add_federation_land_rows` /
`.federation_land_bridge` not found, and `unused argument (federation_land = ...)`.
Two of the blocks assert the *status quo* (`status_quo$value` is all NA; the
"no LUH2 land" warning still fires), which is the reproduction locked in.

## Moves published values — why this is a draft

`"successor_union"` moves pre-1962 production, so it is **not** the default and
nothing moves unless asked. Magnitude measured by running the real
`.read_production(start_year = 1850, end_year = 1965)` both ways and diffing
1,200,298 tonnage rows:

| year | `"none"` t | `"successor_union"` t | change |
|---|---|---|---|
| 1850 | 2,220,163,829 | 2,529,103,111 | +13.92% |
| 1900 | 2,744,345,411 | 3,232,172,750 | +17.78% |
| 1930 | 3,282,464,252 | 3,895,758,256 | +18.68% |
| 1950 | 3,564,747,513 | 4,182,578,549 | +17.33% |
| 1960 | 4,086,029,682 | 4,879,910,128 | +19.43% |
| **1961** | 4,980,020,242 | 4,980,020,242 | **0.00%** |

* **33,633 rows change, on exactly three area codes** (228, 51, 248) and no others.
* **Zero rows change at year >= 1961.** No observed FAOSTAT value moves; only the
  back-cast window does.
* Unmatched areas drop from **4 to 1** (Belgium-Luxembourg alone).
* The 1961 splice discontinuity closes: USSR 1960/1961 goes from `0 -> 704.7 Mt`
  to `696.5 -> 704.7 Mt` (+1.17%); Czechoslovakia +0.42%; Yugoslav SFR +1.18%.

**This is a proxy, and labelled as one.** It is a real pipeline run, but truncated
to 1850-1965 rather than 1850-2023, so it does not exercise the post-1965 half of
the series — which the diff shows is untouched by construction, since only
year < 1962 rows are proxy-filled. No downstream builder (CBS, footprints) was
re-run.

## The open decision

**Should `federation_land = "successor_union"` become the default?** It is a
science decision, not a mechanical one, and the numbers above are the whole cost.
The case for: 12% of early-series production currently vanishes, and the 1961
discontinuity is an artefact of a lookup key, not of history. The case against:
a federation's territory is not identical to its successors' present-day extents,
and the reconstruction is a modern-boundary aggregate rather than a period one —
though that is exactly what LUH2 gives every other area anyway. Left as `"none"`
so this PR moves nothing without consent.

* **Did not invent Belgium-Luxembourg's successors.** Upstream publishes no
  `successor` and no inbound `predecessor` for `BLX-1850-1999`, so it stays
  unmatched (0.48% of 1961-62 tonnes) and the warning still names it. Filed as
  eduaguilera/whep-polities#138 — when upstream adds the relation, this code needs
  no change.
* **Did not implement the polygon-intersection option** (#408 option 2). It is
  unnecessary now that the transitive `successor` walk resolves 3 of 4, and it
  would embed a second, uncheckable territorial reconstruction.
* **Did not touch the French overseas departments**, because they are not broken —
  see the disproof above. If someone wants them separable from 999 that is #419 /
  #414 territory, not this.
* **Did not run `rcmdcheck`** (slow, ten agents on this machine) — the 5-platform
  check is left to CI, which is driven green.
* **Did not re-file the pre-1961 name-proxy join bug.** It is fixed on `main`:
  `.extend_historical()` joins on `area_code`
  (`R/build_production.R:2425`, plus the existing test "matches LUH2 land by
  area_code, not name"), and China (41) back-casts correctly — 168.6 Mt in 1850
  rising to 291.3 Mt in 1961.
* **Filed rather than widened**: #548, the same warning's blind spot for 9 areas
  whose LUH2 land row is exactly zero (0.017% of 1961-62 tonnes, so diagnostic
  rather than material).

Closes #408


lbm364dl:
CI: **10/10 green** on 1dc67085 (R-CMD-check on 5 platforms, lint, format-suggest, pkgdown, test-coverage, offline-tests).

One note for the record: `ubuntu-latest (release)` was `cancelled` on the first attempt, not failed — it stalled 35 minutes inside `setup-r-dependencies` and hit the job timeout, with `check-r-package` never starting. Re-run on the same commit and it passed. Infrastructure, not the diff; the other four platforms plus both suite-running jobs were green on the first attempt.

lbm364dl:
Verified and merging. **The default path is bit-identical to `main`**, which is what makes this safe to land despite the size of what it enables.

Two full-range `get_wide_cbs()` builds, `main` vs this branch with `federation_land` left at its `"none"` default:

```
rows                          2,039,344 -> 2,039,344
columns that move                     0 of 11
all.equal(before, after)           TRUE   (whole frame)
```

The two serialised frames are even the same byte size. So the +13.9% (1850) to +19.4% (1960) you measured is reachable only by opting in, and nothing published moves by merging this.

That is exactly CLAUDE.md's multi-method rule applied properly: the plumbing lands, the status quo stays the default, and the alternative becomes a one-argument choice backed by your measurement rather than a rewrite someone has to attempt later. Given that a sibling PR was reverted today for moving structure while appearing value-neutral (#563), a change that provably moves nothing by default is the shape I want more of in this epic.

Full suite **5762 pass, 0 fail**.

Merge note: `R/polities.R`, `R/utils.R` and `NEWS.md` all collided with work that landed today — `resolve_polity_label()` and the dead-polity filter in `R/polities.R`, two other agents' `globalVariables` entries in `R/utils.R`. All three were both-added; I rebuilt `R/polities.R` from the merge base rather than splicing the conflict halves, and parse-checked it, because splicing produced unparseable R earlier today.

Your open question is untouched and still yours: whether `"successor_union"` should become the default, i.e. whether a dissolved federation's pre-1962 production should be back-cast onto the union of its successors' LUH2 land. Belgium-Luxembourg remains unreachable either way, since upstream publishes no successor for it.

Issue #550, opened 2026-08-05
add_polity_code reads polity_end_year as inclusive while everything else documents it exclusive: 3 area-years land in a dissolved state


lbm364dl:
## The inconsistency

`polity_end_year` has two readings in this package and they disagree by one year.

**Exclusive** — the documented and computed convention:

- `R/table_mappings.R` (`polity_area_crosswalk` docs): "`polity_end_year` is exclusive."
- `R/polities.R:596` (`.area_year_polity_conflicts`): "`polity_end_year` is EXCLUSIVE, so a period covers `start:(end - 1)`", and it computes with `polity_end_year - 1L`.
- `data-raw/table_mappings.R` (`shadowed_by_map`): `polity_end_year - 1L >= map_year_start`, with a comment saying the comparison is deliberately mixed-convention because the polity end is exclusive and the map's `map_year_end` inclusive.

**Inclusive** — what the resolver actually does:

- `R/polities.R` `.add_polity_columns_dt()` joins on `join_end_year >= year`, so a row whose `year == polity_end_year` matches the period.
- `R/table_mappings.R` (`polities` docs) also says "`start_year`, `end_year`: Inclusive validity years for the row", contradicting the crosswalk docs in the same file.

## Measured effect

Over the 1961-2024 grid for all 266 crosswalk areas (`add_polity_code(backcast_anchor = -Inf)`), **7** area-years resolve at exactly `year == polity_end_year`. Checked against the year span the upstream FAOSTAT map declares for that area (inclusive on both ends), **3 of the 7 land one year past the map's span**, in a state that had already dissolved:

| area | resolves to | year | map span | verdict |
|---|---|---|---|---|
| 51 Czechoslovakia | `F51-1947-1993` | **1993** | 1961-1992 | wrong, over-extends 1 yr |
| 186 Serbia and Montenegro | `SCG-1992-2006` | **2006** | 1992-2005 | wrong, over-extends 1 yr |
| 248 Yugoslav SFR | `F248-1991-1992` | **1992** | 1991-1991 | wrong, over-extends 1 yr |
| 15 Belgium-Luxembourg | `BLX-1850-1999` | 1999 | 1961-1999 | correct |
| 151 Netherlands Antilles | `ANT-1961-2010` | 2010 | 1961-2010 | correct |
| 206 Sudan (former) | `SUD-1956-2011` | 2011 | 1961-2011 | correct |
[...]

Issue #517, opened 2026-08-04
Upstream publishes an authoritative FAOSTAT area to polity map and we ignore it, deriving prefixes instead


lbm364dl:
Part of epic #458. Found while refreshing the polities snapshot (#485), where it caused a live coverage gap.

## Problem

`whep-polities` publishes an authoritative FAOSTAT-area-to-polity map as part of its consumer contract:

```
data/final/faostat_area_polity_map.csv
  area_code, year_start, year_end, polity_code, source_label, iso3, match_route, confidence, rows_observed
  281 rows, 228 distinct area codes, ALL high confidence
  match routes: iso-equal 249, registry 18, manual-route 8, manual-replace 5, manual-span 1
```

**`WHEP_POLITIES_FAOSTAT_MAP` is referenced nowhere in this package.** Measured: zero matches across `R/` and `data-raw/`. So the map is published and ignored, and `data-raw/table_mappings.R` re-derives the same mapping by **string-prefix matching** — deriving a `polity_prefix` with `sub("-.*", "", polity_code)` and joining areas to it.

## Why that is not equivalent

A prefix heuristic answers a different question than a curated map, and three known defects all trace to it:

1. **Djibouti loses its mapping entirely.** In #485, once retired polities stop being resolution candidates, FAOSTAT area 72 resolves to nothing. It had been going through `DJI-1886-2025`, which upstream marks dead. Its live polities are `FRS-1884-1977` and `FRS-1977-2025` — the published map names both — but no prefix derived from `DJI` reaches `FRS`. The retired polity was masking an unreachable mapping.
2. **Subnational polities leak in.** `sub("-.*", "")` turns `MMR-LWR-1852-1885` into prefix `MMR`, so a subnational unit joins as though it were a period of Myanmar itself. Five crosswalk rows arrive this way, three of them typed `subnational`.
3. **And the leak is load-bearing.** `AZE-SSR-1920-1991` reaches the crosswalk *only* through that same collapse, so it is the sole reason FAOSTAT area 52 has any pre-1991 coverage. Fixing the regex naively deletes Azerbaijan's 1920-1990 mapping — which is why this cannot be fixed as a regex tweak and needs the map.

## What the map gives that the prefix cannot

- **Year ranges per mapping**, so `(area, year)` resolution is data rather than inference
[...]

PR #551, opened 2026-08-05
Re-sync polities against upstream eb02dcb (740 -> 749) and re-point the six guards that pinned the retired codes

### LINKED ISSUE #295: polities: aggregate recent-year drop, dead exact_start tiebreak, asymmetric partner polity_area_code

### LINKED ISSUE #530: Pending whep-polities re-sync: 10 retired polity codes will trip 5 test files
Filed proactively from the upstream side, because I created this debt and it should be tracked rather than discovered mid-regeneration.

## Nothing is broken today

whep `origin/main` embeds a **740-row** `data/polities.rda` carrying the **old** codes, and its tests pin those codes, so main is internally consistent:

```
RAFR-1850-2021   present      RAFR-1850-2025   absent
CAN-1948-2025    present      CAN-1949-2025    absent
ROW-1850-2023    present      ROW-1850-2025    absent
```

whep's tests reference whep-polities only in **comments** (issue links) and never read the upstream repo directly, so CI is unaffected until someone runs `data-raw/`.

## What changed upstream

Ten codes retired with redirects since that snapshot. Upstream is now **749 rows, 708 live, 41 dead**:

| retired | replaced by | why | upstream PR |
|---|---|---|---|
| `CAN-1886-1948` | `CAN-1886-1949` | Newfoundland acceded **31 March 1949**; CShapes dated the break to the 22 July 1948 *referendum*, so calendar 1948 was Newfoundland-excluded Canada | #92 |
| `CAN-1948-2025` | `CAN-1949-2025` | same; the old row double-counted Newfoundland's 398,084 km² for 1948 | #92 |
| `SER-2006-2008` | `SRB-2006-2008` | duplicate — two live Serbias for 2006-2007; the retired one was bound to Serbia **and Montenegro** | #86 |
| `MOR-1956-1958` | `MAR-1911-1958` | orphaned duplicate covering 1956-57 twice | #83 |
| `BLX-1921-1999` | `BLX-1850-1999` | duplicate of the reporting bucket; its `iso3` was `BEL`, which put a Belgium-Luxembourg row in *Belgium's* family | #117 |
| `RLAM-1850-2013` | `RLAM-1850-2025` | bucket stopped 11 years before FAOSTAT did | #127 |
| `RAFR/RASI/REUR/RNAM/ROCE-1850-2021` | `…-1850-2025` | same, 3 years each | #127 |
| `ROW-1850-2023` | `ROW-1850-2025` | same, 1 year across ~46 folded areas | #127 |

## What will trip, and why each is the guard working

On `main`, five files pin retired codes:

```
tests/testthat/test_polities.R          RAFR-1850-2021, ROW-1850-2023
tests/testthat/test_build_cbs.R         ROW-1850-2023
tests/testthat/test_footprint_paths.R   ROW-1850-2023
tests/testthat/test_datasets.R          ROW-1850-2023
data-raw/table_mappings.R               ROW-1850-2023  (a COMMENT only, not code)
```

`test_reporting_aggregate_coverage.R` on the `polities-integration` branch (#382) is the one whose failure is **most informative**: whep-polities#50 records that it guards the bucket end-years **bidirectionally**, so it is *designed* to fail when upstream fixes them. That is the intended signal, not a regression.

## The 72 area-years this recovers

The bucket extension exists to fix real data loss. Six regional "Other" buckets and Rest of World stopped before FAOSTAT stopped reporting them, so their most recent years resolved to `NA` — `RLAM` lost 2014-2024 entirely. After the re-sync those resolve.

## Suggested sequence

1. Run `data-raw/` to regenerate against upstream `main`.
2. Update the five files' expected codes. Four are one-line substitutions; `test_polities.R` pins two.
3. Update the stale comment in `data-raw/table_mappings.R` — the 31-area FABIO Rest-of-World fold now resolves to `ROW-1850-2025`. The fold itself is unchanged and still tracked as #419/#414.
4. Diff column **totals**, not row counts, against the pre-sync build. whep-polities#50's change adds coverage for 72 previously-unresolved area-years, so Rest-of-World and regional aggregates should *gain* rows — a change in totals here is expected and should be quantified rather than assumed neutral.

Point 4 matters because of eduaguilera/whep#444: a previous "changes no published numbers" claim on #382 was measured with row counts while every value was itself a row count, so it certified group sizes and was read as certifying values.

lbm364dl:
## What was wrong

Upstream `whep-polities` retired or superseded polity codes that this package's embedded snapshot still carried as live, and test files pinned the old codes. Nothing was broken on `main` — it is internally consistent, and #530 is correct that it embeds a **740-row** `polities.rda` — but the debt fired the moment anyone ran `data-raw/`.

**Both ends of the comparison are pinned to a git revision, not to a working checkout**, because upstream moved several times during this run and `data-raw/table_mappings.R` defaults `WHEP_POLITIES_GPKG` to the live checkout:

| end | upstream revision | `polities_database.gpkg` sha256 (first 16) |
|---|---|---|
| before | `ce7bc37` | `87eec2a6b569445e` |
| after | `eb02dcb` | `bc3642ac497cbb5d` |

Upstream `origin/main` is now `5ae7548`, whose `data/final/` artifacts are **byte-identical** to `eb02dcb`'s (same three sha256s) — the commits since then touched scripts and wiki pages, not the published database. So the after end is upstream `main` as of this PR, and the hash says so rather than a branch name that keeps moving.

**Three claims in the issue measured differently, and I report what I measured:**

| issue #530 says | measured |
|---|---|
| "**10** retired polity codes" (title); its own table lists 12 | **14** codes went live→dead relative to main's committed data: the table's 12 plus `MNE-1913-1915` (retired) and `PER-1825-1909` (superseded) |
| "**5** test files" pin retired codes | **6** test files break. It missed `test_polity_faostat_map.R` (a real code assertion) and `test_polity_resolution_uniqueness.R`; and it counted `data-raw/table_mappings.R`, which is comments only and cannot break |
| "the **72** area-years this recovers" | **88** area-years recover |

At base commit `2d34d02c`, the six affected test files are green:

```
polities                 pass=39 fail=0   polity_faostat_map            pass=32 fail=0
build_cbs                pass=80 fail=0   footprint_paths               pass=48 fail=0
datasets                pass=552 fail=0   polity_resolution_uniqueness  pass=10 fail=0
```

Then, at that same commit, regenerating `data-raw/table_mappings.R` + `data-raw/harmonization_tables.R` against pinned upstream `eb02dcb` and running the **whole** suite with no other change:

```
TOTAL pass= 5410  fail= 9  err= 1  skip= 9

test_build_cbs.R                      .resolve_hist_trade_polities drops pre-range aggregate rows
test_datasets.R                       CBS and FABIO area codes map to polity database rows
test_footprint_paths.R                origin_area keeps the legacy meaning of an area code   [ERROR]
test_polities.R                       add_polity_code maps area codes by year                [2]
test_polities.R                       add_polity_code does not extend aggregate rows outside their range
test_polities.R                       get_polity_geometries returns requested polygon rows
test_polity_faostat_map.R             the FABIO Rest-of-World fold still outranks the map
test_polity_resolution_uniqueness.R   the shipped crosswalk's conflicts are the known ones   [2]
```

### The drift decomposes into two upstream steps, and the first one is invisible to a row count

Rebuilding against each pinned revision in turn:

| vintage | `polities` rows | dead rows | crosswalk rows | area-year conflicts | unresolved area-years |
|---|---|---|---|---|---|
| main's committed `data/` | 740 | 27 | 598 | 86 | 191 |
| upstream `ce7bc37` | **740** | **31** | 596 | **0** | 191 |
| upstream `eb02dcb` | **749** | **41** | 596 | 0 | **103** |

The `polity_code` **set** is identical between main's committed data and `ce7bc37` — same 740 codes, not one added or removed. What changed is `wiki_status`: `ce7bc37` marks four more codes dead (`MNE-1913-1915`, `MOR-1956-1958`, `PER-1825-1909`, `SER-2006-2008`). So main's committed `polities.rda` was already a stale vintage of the same 740 rows, and **no row-count check could have detected it** — which is exactly the failure mode #530's point 4 and #444 warn about. That step alone removes the two conflicting pre-1961 crosswalk rows (area 170 `PER-1825-1909`, area 273 `MNE-1913-1915`) and takes the resolution-conflict count 86 → 0.

The second step, `ce7bc37` → `eb02dcb`, adds the 9 new codes, retires 10 more, and is what recovers the 88 area-years and relabels the buckets.

Object-level diff of the shipped data:

```
polities               740 -> 749 rows   9 added, 0 removed; dead 27 -> 41
polity_area_crosswalk  598 -> 596 rows   2 removed (both newly-dead pre-1961 duplicates)
regions_full            63 rows change reporting_polity_code
polities_cats           11 rows change reporting_polity_code
```

## What I changed

1. **Regenerated `data/polities.rda`, `data/polity_area_crosswalk.rda`, `data/regions_full.rda`, `data/polities_cats.rda`** against pinned upstream `eb02dcb`, via the documented `WHEP_POLITIES_GPKG` / `WHEP_POLITIES_FAOSTAT_MAP` contract.

   Eight further `.rda` files (`items_cbs`, `items_prod`, `items_full`, `animals_codes`, `cb_processing`, `crops_eurostat`, `liv_lu_coefs`, `primary_double`) came out byte-different but `identical()` in R — a re-serialisation artefact of the readr `spec`/`problems` attributes. I reverted those, so the `data/` diff contains only objects whose contents actually changed.

2. **Re-pointed the pinned codes so each guard still guards its thing.** Four are substitutions. Two are not:

   - `test_polities.R`'s aggregate out-of-range guard asked about **area 904 in 2021**. 2021 was outside `RLAM-1850-2013` and is *inside* `RLAM-1850-2025`, so a code substitution would have left the test asserting the opposite of what it exists for. It now asks about **2026**, the first year past the bucket's end, so the year tracks the bucket rather than an accident of the old vintage.
   - `test_polity_resolution_uniqueness.R` **pinned** two `(area, year)` conflicts, and its header said the pin "shrinks when upstream fixes Montenegro". Upstream fixed both, so the pin has shrunk all the way and the test now asserts the invariant at **zero** instead of enumerating exceptions. Measured history of that number: 199 conflicting area-years (prefix inference) → 86 (#517 adopted the upstream map) → **0** (this re-sync).

3. **Refreshed the stale prose in `data-raw/table_mappings.R`**: the dead-polity filter's "inert for the committed snapshot… excludes 27, retains 713" is now "excludes 41, retains 708" and load-bearing; the `MNE`/`PER` overlap it said no downstream filter could fix is now fixed upstream; and the `ROW-1850-2023` mention in the FABIO-fold comment is updated.

4. **Documented `polities$wiki_status`**, which this refresh makes load-bearing — a consumer must not resolve *to* a dead row, and the crosswalk is the table that draws the line.

## How I verified

Gates, all run, not predicted:

- `air format .` — binary, clean.
- `devtools::document()` — `man/polities.Rd` regenerated and committed.
- `lintr::lint_package(...)` with the four CI-disabled linters — **0 lints**. (It first flagged one new comment as `commented_code`; reworded.)
- `devtools::test()` full suite — **5421 pass, 0 fail, 0 error, 9 skip** (5410 / 9 / 1 / 9 before the test edits, on the same regenerated data).
- pkgdown reference completeness: 289 `man/` topics, all present in `_pkgdown.yml`, 0 missing.
- `rcmdcheck` deliberately left to CI (5 platforms; 10 agents share this machine).
- **The committed `data/` is byte-reproducible from the pinned revision.** Re-running both `data-raw/` scripts a second time against `eb02dcb` regenerated all four changed `.rda` files byte-identically to the commit (`git status data/` clean for them), and the four objects compare `identical()` in R. The eight `spec`-attribute files mentioned above came out byte-different again on that second run while staying `identical()` in R, which confirms their churn is a serialisation artefact unrelated to this change and correctly excluded.

**Assertion count that fails without this change: 10** (9 failures + 1 error) across 6 test files, listed above.

## Why this is a draft

The polity vintage changes underneath everything, so I measured the closest honest proxies at the crosswalk level. I did not run a full-range build — hours, plus multi-GB local rasters.

**What I measured, and it argues the change is value-neutral inside the default build range:**

- `polity_area_code`, the numeric bucket every matrix workflow aggregates on, is **unchanged for all 267 reporting areas**. No area code enters or leaves the crosswalk; bucket 999 keeps the same 62 members. That is the mechanism a #444-style shift would need, and it is absent.
- Resolving the full 266-area × 1961-2024 grid (17,024 area-years) with `add_polity_code(backcast_anchor = -Inf)`: unresolved **191 → 103**, i.e. **88 recovered, 0 lost**. The residual 103 are all deliberate: area 15 Belgium-Luxembourg after 1999 (25), area 151 Netherlands Antilles after 2010 (14), area 351 China the intentionally-unmapped statistical aggregate (64).
- Of the 88 recovered, **68 are calendar 2024** — 62 RoW-folded source areas (which include area 999 itself) plus one year each for the six regional buckets — and 2024 is outside `build_cbs`'s default `1850:2023` range. The other **20 are all on areas 901-906**, which are WHEP's own regional "Other" reporting labels appearing in no source dataset (they live in `regions.csv`, not `regions_full`, and carry no `fabio_code`). So no row a default build reads flips from dropped to kept — and dropping is the only way this could move a total, at `build_cbs.R:820`, `build_cbs.R:1225` and `read_raw_inputs.R:323`.
- **4,328** area-years get a different `polity_code` **string** for the same bucket (3,906 of them `ROW-1850-2023 → ROW-1850-2025`, 64 `CAN-1948-2025 → CAN-1949-2025`, the rest the regional buckets). A label change, not a quantity change.
- **Canada / Newfoundland, verified on the shipped geometries.** With the back-cast anchor off, calendar **1948** moved:

  ```
  BEFORE  1947 CAN-1886-1948   AFTER  1947 CAN-1886-1949
          1948 CAN-1948-2025          1948 CAN-1886-1949
          1949 CAN-1948-2025          1949 CAN-1949-2025
  ```

  `sf::st_area()` on those polygons: pre-accession 9,379,600 km², post-accession 9,774,537 km² — a **394,937 km²** correction for 1948. (The issue quotes upstream's 398,084 km² for Newfoundland; 394,937 km² is what the shipped geometries give.) The two pre-accession rows are the *same* polygon — `polygon_feature_year` 1886 and `polygon_area_km2` 9,553,560 in both — so the fix is about which calendar year belongs to which period, not about a polygon being redrawn. Under the default anchor (1961) a 1948 Canada row floors to the 1961 anchor territory either way, so only genuine historical sources resolved with the anchor off see this.

**What I did not measure:** any actual tonnage, area or emission total. No `build_primary_production()`, `build_commodity_balances()`, footprint, nitrogen or carbon build was run.

## Merge sequencing

**This PR establishes the polities vintage; #479 (`polity/389`) and #480 (`polity/459`) should regenerate their `data/*.rda` on top of it, not before it.** Both currently conflict with `main` on generated `data/` artifacts. Regenerating them against current upstream *before* this lands would fold this re-sync into their diffs and make the vintage change unattributable to either PR — the same conflation that left #444 unexplained. So: land this, then rebase those two and re-run their `data-raw/` step against the vintage recorded here (`eb02dcb`, gpkg sha256 `bc3642ac497cbb5d…`).

## On #444

I could **not** attribute any part of #444's feed +7% / production +2% to a specific retired row, and I do not think this re-sync is that mechanism. #444 localised its shift to added rows at areas 999, 229 and 11 with `production == feed`; this re-sync touches neither 229 nor 11, adds no row inside the default range, and leaves every `polity_area_code` identical — which #444 had already rejected as a hypothesis ("both crosswalks have the same 267 area codes and the same 62 members of bucket 999"), and my measurement independently reproduces that. #444's residual suspects were `regions_full`, `polities`, `polities_cats`, `cow_to_lpjml.csv` and `urban_n_reference` on the pre-#517 `polities-integration` branch, whose crosswalk was still prefix-derived; the map-based crosswalk on `main` is a different object. #444 stays open and unexplained; this PR does not claim to close it.

What is worth recording is that the 88 recovered area-years are the *same class* of effect #444 was worried about — previously-dropped rows becoming resolvable — just an instance landing outside the default range. If anyone widens a build to 2024, those 68 area-years — 62 RoW-folded source areas and the seven buckets — will newly be retained rather than dropped, and RoW totals will rise. That is an improvement, not neutrality, and should be stated as one.

- **Did not touch upstream.** Every territorial fact here was decided upstream and is already published; whep's job is to consume it. I read `/home/usuario/whep-polities` at pinned revisions and wrote nothing to it.
- **Did not lift the FABIO Rest-of-World fold.** 31 map-covered areas still resolve to the RoW bucket rather than the real polity the map names. Tracked as #419/#414 and out of scope; `test_polity_faostat_map.R` still pins the 62 folded areas so it cannot move by accident.
- **Did not fix #295's aggregate-fallback logic.** The upstream bucket extension removes the *symptom* for these seven buckets — that issue names exactly the loss I measured, "area 904 loses 2014-2023, areas 901/902/903/905/906 lose 2022-2023" — but the fallback still hard-excludes aggregates, so the next bucket that ends early will lose its recent years again. That code fix belongs to #295 / PR #302 and I stayed off it.
- **Did not fix the `polity_end_year` convention bug** found while checking bucket boundaries: the resolver joins on `end_year >= year` (inclusive) while the crosswalk docs, the conflict detector and the crosswalk build all treat it as exclusive, putting 3 area-years in a state that had dissolved (1993 Czechoslovakia, 2006 Serbia and Montenegro, 1992 Yugoslav SFR). Filed as **#550** with measurements rather than widening this PR.
- **Did not weaken any guard.** Every pinned number that could stay, stayed (62 folded areas; the 7 prefix-fallback areas `351, 901-906`; `CAN-1800-1866`'s pinned polygon-status defect); the two that had to move now test the property rather than the old vintage, and one tightened from a pin to zero.
- **Did not check upstream's succession graph as a finding.** 12 live polities point their `successor` at a dead row and `BRL-1945-1949` points at a nonexistent `EBL-1949-1990`; I checked upstream's `validate_references.py` before filing anything and both are already handled there — dead targets are warnings by design, and the dangling link is in `validate_references_baseline.txt`. Not re-filed.
- **Did not run `rcmdcheck`** locally, per the machine-sharing note.

## The open decision

**Nothing methodological in the code** — the vintage is upstream's decision and this PR only consumes it. One call for the owner, and it is why this is a draft:

> Is crosswalk-level evidence enough to merge a polities vintage bump, or does a vintage bump require a real build comparison first? Everything measurable says nothing numeric moves inside `1850:2023`, but "nothing moves" is a claim about a full build and I did not run one. #444 exists precisely because a similar claim was accepted on a proxy that could not have detected the movement. If you want this priced properly, the cheapest sufficient run is `get_wide_cbs()` column totals over the full range, before and after, on a machine with the pins.

Closes #530


lbm364dl:
Reviewed the sequencing against the PRs I brought up to date today. **Your merge-order note is right and I have made the one PR that would have broken it independent of the vintage instead.**

## The collision this would have caused

`tests/testthat/test_polities.R` is touched by both this PR and #302, and the conflict was semantic rather than textual. #302 asserts that aggregate reporting areas are *not* extended past their period, using area 904 in **2021** — outside `RLAM-1850-2013`, **inside** your `RLAM-1850-2025`. Merging in either order would have flipped one of them into asserting its own negation. You caught the identical trap in your own edit to that file and moved to 2026; I have gone one step further on #302's side and made both of its guards read the bucket's end year off the crosswalk, asking about `end - 1` and `end + 1`.

Verified both ways: 44 pass / 0 fail on main's 740-row vintage, and the same two tests pass with your `polity/530` data overlaid (area 904 then reads `RLAM-1850-2025`, end year 2025). So the two branches can now land in either order.

## Confirming your new finding from the other side

Your two-step decomposition matches what I hit independently while regenerating #480's data. I needed a vintage that reproduces `main`'s committed **598-row** crosswalk so #480's delta would be attributable to its own change, and `ce7bc37` gave 596 — exactly the `wiki_status` step you describe, invisible to a row count because the code set is identical. `eadf1db` is the revision that reproduces 598, and that is what #480 is currently regenerated against.

**So after this lands, #480 and #479 regenerate against `eb02dcb` as you recorded.** Concretely:

- **#480** (`polity/459`) — regenerate `polity_area_crosswalk`, `regions_full`, `polities_cats`. It also touches `test_datasets.R` and `test_polity_faostat_map.R`, which you edit too, so expect a textual merge there.
- **#479** (`polity/389`) — this PR is what **unblocks its alias-map regeneration**. It currently aborts on exactly three targets missing from the 740-row vintage: `CAN-1886-1949`, `CAN-1949-2025`, `ROW-1850-2025`, all three of which arrive with you. Its `mueller_synthetic_n` coverage bound should also lift from 4832 back above 4999, because the five spanning rows causing the ambiguity (`BRA-1800-2025` beside `BRA-1909-2025`, same shape for ARG/AGO/GRC/IRQ) are among your retirements. I left that as a lower bound so it passes either way.

## On your open question

For what it is worth from the crosswalk side: your evidence looks sufficient to me for the *structural* claim, and it agrees with what I measured on #480 — `polity_area_code` unchanged for all areas, no bucket membership change. That is the invariant that decides whether any published number can move, since every build keys on the numeric bucket. The residual risk is the 4,328 area-years whose `polity_code` string changes, and that only reaches a number through a consumer joining on the string rather than the code. Still the owner's call, and the `get_wide_cbs()` totals run you name is the right cheapest check.

Also noting #550 as a good catch — the inclusive/exclusive `end_year` boundary landing 3 area-years in dissolved states is the kind of thing that would have been read as a rounding detail.

lbm364dl:
Correcting the value-neutrality claim in this body with the full-build number, which did not exist when it was written.

This PR's argument was that the re-sync is neutral inside the default build range, from crosswalk-level proxies: `polity_area_code` unchanged for all 267 areas, 88 area-years recovered and 0 lost, 68 of them in calendar 2024 which is outside `1850:2023`. All of that is correct and I reproduced it.

It is not the whole effect. Two full-range `get_wide_cbs()` builds — `main` at `6e5af760` against its own tree with this PR's four `.rda` files swapped back to `3f479f7a`'s, so the polity data is the only differing input — measure:

| | before | after | ratio |
|---|---|---|---|
| rows | 2,019,253 | 2,005,267 | 0.9931 |
| `feed` | 9.0636e11 | 9.0501e11 | 0.9985 |
| `other_uses` | 9.2130e10 | 9.1787e10 | 0.9963 |
| `stock_withdrawal` | 1.4828e11 | 1.4651e11 | 0.9881 |

Globally small. Per area not: the whole −13,986 rows fall in **1850–1960**, at 126 rows a year for all 111 years, in **four** areas — 84 Greece (−28.5% of pre-1961 `feed`), 202 South Africa (−58.6%), 7 Angola (−98.3%), 72 Djibouti (−100%) — plus 249 Yemen moving +54.0% with no row change. The other 190 areas are unchanged to the row.

Those four are exactly the four whose `polity_name` this re-sync periodized (`Greece` → `Greece (1947-2025)`, and so on). `area` — the polity name — is still a join key in the pre-1961 chain, and `.resolve_hist_trade_polities()` labels it with `area_name` while everything else labels it with `polity_name`, so renaming a polity deletes that area's back-cast. 68 buckets are already mismatched and every future re-sync can add one silently. Filed as #584.

Nothing here argues the re-sync was wrong — the new vintage is the correct one and the names it introduces are better. It argues that the neutrality claim needed a build to make, and that the mechanism which made it false is worth fixing before the next re-sync. Measured in #587, which also adds an offline guard so the coverage half of this question is answerable in a second.

Issue #563, opened 2026-08-05
The polity-level ROW un-fold splits one bucket into several area labels, so .aggregate_to_polities() stops summing it


lbm364dl:
Filed from the revert of #480 (PR #561), with the measurement, so the un-fold can be redone correctly rather than rediscovered.

## The defect

`.aggregate_to_polities()` groups by `area` — the polity **name** — as well as `area_code`. So un-folding a FABIO Rest-of-World member at the *polity* level, while leaving `polity_area_code` at `999`, splits the bucket's rows without moving any value:

```r
raw <- tibble::tribble(
  ~year, ~area_code, ~item_cbs_code, ~element,     ~unit,    ~value,
  2010L, 212L,       2511,           "production", "tonnes", 300,   # Syria
  2010L, 999L,       2511,           "production", "tonnes", 40,    # RoW itself
  2010L, 40L,        2511,           "production", "tonnes", 7
)
whep:::.aggregate_to_polities(data.table::as.data.table(raw), item_cbs_code)
```

| with the fold (main today) | with #480's un-fold |
|---|---|
| `999  Rest of World  340` | `999  Syrian Arab Republic  300` |
| | `999  Rest of World       40` |

Mass conserves — 347 in, 347 out — so no conservation check sees it.

## Why it matters

Two known, expensive failure modes:

- **`area` is a join key.** #382 measured this exact shape and withdrew the change: four inner joins use `c("area", "area_code")`, and creating a second `area` vocabulary for one bucket dropped **702,166 rows** there.
- **Duplicate `(area_code, year, item)` keys** are what fed the `dcast()` `length()` fallback in #425 — the defect that turned every commodity-balance value into a row count, and whose corrupted baseline produced #419's phantom 13.7x feed inflation (disproved in #555 at 1.0000).

## Why it was not caught by the obvious checks

`polity_area_code` is unchanged for all 266 reporting areas, no area enters or leaves a bucket, and bucket membership is identical. Every one of those measurements passes. **The numeric bucket is what values are summed into; `area` is what they are grouped by.** Un-folding the identity splits the group while leaving the bucket alone, so the value-neutrality checks are all blind to it.

It was caught only by #555's `.aggregate_to_polities() sums a fold and reports it`, which asserts bucket 999 comes to 340.

## Target state

[...]

Issue #544, opened 2026-08-05
polity_area_crosswalk's mapping_status conflates a curated match with a prefix-guess fallback


lbm364dl:
Recovered from the now-closed PR #305 so the idea is not lost with it.

## What

`polity_area_crosswalk$mapping_status` uses one label, `matched`, for outcomes that carry very different confidence. `data-raw/table_mappings.R` resolves an area through several branches — the upstream curated FAOSTAT map, an `area_iso3c` prefix, the `reporting_polity_code` prefix, a `manual_area_prefixes` override, and the FABIO rest-of-world bucket — and a consumer reading the shipped dataset cannot tell which branch produced a given row.

That matters because the branches are not equally trustworthy. `main`'s own comments record what prefix inference gets wrong when it is the only authority:

- prefix `BLX` reached `BLX-1921-1999` for FAOSTAT area 15, a row that was itself a duplicate of the reporting bucket and carried `iso3 = BEL`, putting a Belgium-Luxembourg row inside *Belgium's* family (retired upstream, see #530);
- `sub("-.*", "", polity_code)` collapses `MMR-LWR-1852-1885` onto prefix `MMR`, and three of the five overlapping-period crosswalk defects in #458 enter that way — while `AZE-SSR-1920-1991` reaches the crosswalk *only* through the same quirk, so naively fixing the regex silently removes Azerbaijan's 1920-1990 coverage.

A curated hit and a prefix guess are different epistemic objects, and today they are indistinguishable in the published data.

## Suggested target state

Split `mapping_status` so the branch is recoverable — PR #305 proposed `matched_curated` / `matched_fallback` / `row` / `unmapped`, which is a reasonable vocabulary. Then the two questions this unblocks become answerable by filtering rather than by re-deriving the build:

- how many areas still depend on prefix inference at all (if the answer is small, the fallback can be retired);
- whether any area whose two candidate branches *disagree* is currently resolved by the weaker one.

## Notes

- Contract-only change to a published dataset's `mapping_status` column — it adds resolution rather than moving a mapping, so no `polity_code` or `polity_area_code` should change. That should be asserted, not assumed.
- Any consumer filtering on `mapping_status == "matched"` needs updating; PR #506 (`.dependency_sovereign_iso3()`) and #504 both read the crosswalk, so check them.
- Worth doing alongside or after #530, since the upstream re-sync changes which rows exist.

Part of the polity migration epic #458. Recovered from #305.

PR #575, opened 2026-08-05
fix(regions): stop calling four of the Rest-of-World folds a FABIO convention

### LINKED ISSUE #556: regions_full marks Syria, Eswatini, North Macedonia and New Caledonia as CBS reporters AND folds them into FABIO's Rest of World
`regions_full` says two contradictory things about five reporting areas: that
they have their own commodity balance sheet, and that FABIO folds them into
Rest of World.

```
> regions_full |> filter(fabio_code == 999, cbs) |> select(code, iso3c, name)
   code iso3c  name
    999   ROW  RoW
    209   SWZ  Eswatini
    154   MKD  North Macedonia
    153   NCL  New Caledonia
    212   SYR  Syrian Arab Republic
```

`cbs == TRUE` is the flag that says the area reports its own balance sheet;
`fabio_code == 999` is the flag that sends its rows into bucket 999. Area 999
itself is legitimately both. The other four are territories, and one of them is
the **single largest contributor to the fold**: Syria carries 24,426
`faostat-production` rows and 21,053 `faostat-fbs-new` rows, more than any other
folded area (measured at `main` 60fc6610, see #555). 57 of the 62 areas with
`fabio_code == 999` have `cbs == FALSE`, so these five are the anomaly, not the
pattern.

Upstream disagrees with the fold too. `whep-polities`'
`faostat_area_polity_map.csv` names a real polity for each, at `iso-equal`
route and `high` confidence:

| area | upstream polity | rows_observed |
|---|---|---|
| 212 | `SYR-1946-1967` / `SYR-1967-2025` | 113,504 |
| 209 | `SWZ-1894-2025` | 180,663 |
| 154 | `MKD-1991-2025` | 115,222 |
| 153 | `NCL-1800-2025` | 161,177 |

## The question

Is `fabio_code == 999` a **statement about FABIO's region list** (FABIO does not
enumerate this country, so it has nowhere else to go) or a **data error in
`regions_full`**? The two lead to different fixes:

* If FABIO really does not enumerate Syria, the fold is correct for
  FABIO-compatible matrix output and the fix is #120's — defer the collapse to
  the matrix boundary and keep per-country grain in the base tables.
* If `regions_full` is simply wrong for these areas, the fix is a one-cell
  correction per area and no design change is needed.

**I could not settle this offline.** Deciding it needs FABIO's own region list
(`io_codes.csv` from the FABIO release, which `inst/scripts/compare_fabio.R`
downloads into `FABIO_DIR`); nothing in the repo carries it, and I am not going
to guess whether a 191-country database enumerates Syria.

## Why it matters now

#555 re-measured the consequence of lifting the fold on a full-range
`get_wide_cbs()`: **at most 1.2% on any global CBS total** (`stock_addition`),
under 0.1% for feed, production and processing. The 13.7x feed inflation
recorded in #419 does not reproduce — it came from the `dcast()` duplicate-key
trap fixed by #429. So the cost side of this decision is now small and the
decision is unblocked; what is missing is this fact.

Reproduce the measurement with `options(whep.unfold_rest_of_world = TRUE)`
(added in #555).

Related: #120 (defer the collapse to the matrix boundary), #414 (the same
problem for bucket 206), #467, #544 (`mapping_status` cannot express "folded").

lbm364dl:
## What was wrong

`regions_full` says two contradictory things about four reporting areas, and
#556 could not settle which side was wrong because it needed FABIO's own region
list, which is not in the repo.

`cbs == TRUE` says the area reports a commodity balance sheet of its own.
`fabio_code == 999` is the value `polity_area_code` is **derived from** in
`data-raw/table_mappings.R` (`polity_area_code = if_else(!is.na(fabio_code),
fabio_code, area_code)`), so it is not merely a fact about FABIO — it is the
instruction that sums every row the area reports into Rest of World. Four areas
carry both flags at once, and `folded_reporting_areas()` labelled all of them
`fold_kind == "fabio_rest_of_world"`, i.e. asserted that FABIO puts them there.

**It is available, and FABIO does not.** The finding, with three independent
confirmations:

1. `io_codes.csv` of the FABIO v1.1 release (Zenodo record 2577067 — the file
   `inst/scripts/compare_fabio.R` already downloads into `FABIO_DIR`) has 24,000
   rows = **192 areas x 125 commodities**. Areas **153 New Caledonia, 154 North
   Macedonia, 209 Eswatini and 212 Syrian Arab Republic each have their own
   125-row block**, distinct from area 999 `RoW`.
2. The FABIO source repository (<https://github.com/fineprint-global/fabio>)
   does the fold in `replace_RoW(cbs, codes = regions[current == TRUE, code])`
   — everything **not** in that code list becomes 999/`RoW`. All four are
   `current == TRUE` in its `inst/regions_full.csv`.
3. The 192 codes that same file flags `cbs == TRUE` are **exactly** the 192
   area codes of `io_codes.csv` (`identical()`, 0 in either difference), which
   is what makes `cbs` the right in-repo discriminator: WHEP's `cbs` set is a
   superset of FABIO's 192, so `cbs == FALSE` implies FABIO does not enumerate
   the area either.

So `fabio_code == 999` for these four is **a statement WHEP makes, not one
FABIO makes**. It is a data error in `regions_full` — but correcting it moves
published values, so the label is what changes here, not the number.

At `origin/main` `2759be3e`, offline, on the shipped package data.

The contradiction, exactly as #556 reports it (`regions_full`):

```
   code iso3c  name                  cbs   fabio_code
    999  ROW   RoW                  TRUE          999
    209  SWZ   Eswatini             TRUE          999
    154  MKD   North Macedonia      TRUE          999
    153  NCL   New Caledonia        TRUE          999
    212  SYR   Syrian Arab Republic TRUE          999
```

62 areas carry `fabio_code == 999`: **57 `cbs == FALSE`, 5 `cbs == TRUE`** (999
itself legitimately both). The rule `fabio_code` actually follows is exact and
its exceptions are readable — of 202 `cbs == TRUE` areas, **195 have
`fabio_code == code`** and 7 do not:

| code | area | fabio_code | why |
|---|---|---|---|
| 62 | Ethiopia PDR | 238 | successor fold |
| 276 | Sudan | 206 | successor fold |
| 277 | South Sudan | 206 | successor fold |
| 153 | New Caledonia | 999 | **#556** |
| 154 | North Macedonia | 999 | **#556** |
| 209 | Eswatini | 999 | **#556** |
| 212 | Syrian Arab Republic | 999 | **#556** |

And **0** of the 58 `cbs == FALSE` areas keep their own code (57 get 999, China
351 is `NA`), so `cbs` separates the two populations cleanly.

`folded_reporting_areas()` at base returned 72 rows, **61 labelled
`"fabio_rest_of_world"`** — including the four FABIO does not fold — and 11
`"successor_state"`.

FABIO's list, read at base from `io_codes.csv`:

```
      area_code                 area
14126       153        New Caledonia
14251       154      North Macedonia
19376       209             Eswatini
19751       212 Syrian Arab Republic
23876       999                  RoW
n regions: 192
```

## What I changed

Nothing that moves a number. `regions_full` is untouched.

* **`folded_reporting_areas()` gains a third `fold_kind`,
  `"cbs_reporter_folded"`** (4 areas), so `"fabio_rest_of_world"` (57) means
  what it says. The discriminator is the crosswalk's own `cbs` column, which is
  now a required input; the FABIO evidence above is recorded in the roxygen
  `@section What FABIO's own region list says` with the DOI.
* **`options(whep.unfold_rest_of_world = "cbs_reporters")`** promotes only those
  four — the narrower experiment #556 sets up: lift exactly the folds FABIO does
  not make, leave the 57 it does. `TRUE` and `"all"` remain the all-61 mode,
  `FALSE`/`"none"` the default. An unrecognised value now aborts instead of
  silently behaving like `FALSE`.
* **`.rest_of_world_members()` is now the single predicate** both call sites
  use. `.polity_crosswalk()` and `.iso3c_area_code_lookup()` each spelled the
  same four-clause condition out; two tables stating one fold, and only one of
  them being rebuilt, is precisely the failure mode #419 records.
* **The `fabio_code` rule and its seven exceptions are pinned by a test** in
  `test_datasets.R`, and documented on `regions_full` in `R/datasets.R`.
* One stale doc value fixed in passing: `read_population()` said 999 resolves to
  `ROW-1850-2023`; since #551 it is `ROW-1850-2025`.

## How I verified

Gates, all run locally with `Rscript --vanilla`:

* `air format .` (binary), `devtools::document()`, `man/` committed.
* `lintr::lint_package()` with the four CI-disabled linters: **No lints found.**
* `devtools::test()` full suite: **FAIL 0 | WARN 11 | SKIP 8 | PASS 5717 |
  ERROR 0**. The warnings and skips are pre-existing (fold warnings the tests
  assert, env-var-gated raster tests).
* `man/*.Rd` against `_pkgdown.yml`: 0 missing.

**Assertions that fail without the change.** Running the new/changed tests
against `origin/main`'s `R/` (tests from this branch, sources from base):
**11 failed assertions plus 1 error, across 5 tests** —
`folded_reporting_areas() names every fold and its kind` (1),
`the four CBS reporters folded into 999 are named (#556)` (4),
`folded_reporting_areas() needs the cbs flag` (1),
`the unfold switch can lift only the CBS reporters` (5), and
`an unrecognised unfold mode aborts instead of folding` (error —
`.unfold_rest_of_world_mode()` does not exist at base).

Being honest about the rest: the new `test_datasets.R` invariant
(`regions_full sets fabio_code from cbs, with seven exceptions`) **passes at
base too**. It is a regression pin on the shipped data, not a bug
demonstration — its job is that the four cannot leave or be joined silently.

**Structural checks, not just "no number moved"** — the #480 lesson:

* `.polity_crosswalk()` is `identical()` before and after (all 26 columns, not
  just `polity_area_code`).
* `folded_reporting_areas()` returns **the same 72 rows**, and
  `identical()` on `c(area_code, area_name, area_iso3c, polity_area_code,
  polity_code, polity_name)`. Only `fold_kind` differs, on **exactly 4 rows**:
  153, 154, 209, 212. No key added, none removed.
* **One `area_code` -> one `area` label** out of `.aggregate_to_polities()`, the
  guard added for #563: driven directly on the live 212/999/40 shape, bucket 999
  comes out as a single row, `999 | Rest of World | 340`, and
  `tapply(out$area, out$area_code, n_distinct)` is `1,1`. The full
  `test_read_raw_inputs.R` also passes.
* `.iso3c_to_area_code("SYR")` is still `999L` on the default path, and the test
  pinning the default crosswalk against the committed dataset still passes.

## Moves published values

**No.** `regions_full`, `fabio_code`, `polity_area_code`, every polity label and
the crosswalk are byte-identical on the default path, verified with
`identical()` above. The only behaviour changes are the `fold_kind` label, a new
opt-in option value, and an abort on an unrecognised option value.

* **Did not correct `regions_full$fabio_code` for the four.** That is the fix
  #556 anticipates, and the fact is now settled in its favour — but it moves
  published values and is the owner's call, not mine. See the open decision.
* **Did not regenerate `polity_area_crosswalk.rda`.** It is derived from
  `fabio_code`, so correcting the CSV means re-running
  `data-raw/table_mappings.R`, which also rewrites `polities.rda` from live
  upstream and would drag the #485 snapshot refresh into this PR.
* **Did not vendor FABIO's `io_codes.csv`.** It is a 24,000-row third-party
  release file under CC BY-NC-SA; `inst/scripts/compare_fabio.R` already
  downloads it from Zenodo, so the finding is reproducible in one line rather
  than duplicated in the repo.
* **Did not touch bucket 206.** FABIO v2 folds 206 into 276 (`area_merge(prices,
  orig = 206, dest = 276)`), the opposite direction from WHEP's 276/277 -> 206.
  That is #414's question, not this one.
* **Did not touch `whep-polities`.** Upstream already names a real polity for
  all four at `iso-equal`/`high` confidence; the fold is imposed downstream.
* **Did not make the sensitivity option relabel the promoted rows.** It moves
  `polity_area_code` but not `polity_code`, so a promoted area comes out under
  its own code still labelled "Rest of World" — pre-existing behaviour of the
  `TRUE` mode from #555, and the mirror image of #480, which moved the label
  and not the code. It does not split any bucket (verified in all three modes),
  but it is why only correcting `fabio_code` is the finished un-fold.
* **Did not investigate `seed`.** It moves +0.68% here against +0.04% for the
  superset of 61 in #555, which is non-monotonic and above #420's noise floor.
  Filed rather than widened into this PR.

## The open decision

**#556's question is now answered: `fabio_code == 999` is wrong for 153, 154,
209 and 212.** FABIO enumerates all four. What remains is whether to act on it,
and that is a value-moving choice:

1. **Correct the four cells** — `fabio_code = code`, rebuild `regions_full.rda`
   and the crosswalk. **Cost measured on this branch** (two full-range
   `get_wide_cbs()` builds, 1850-2023, ~7 min each; full table in the comment
   below): **at most 1.14%** on any global total (`stock_addition`), 0.68% on
   `seed`, under 0.05% on production, feed, food, processing, trade and
   domestic supply. The baseline reproduces #555's `main` column to every
   printed digit. Structurally: 4 area codes added (exactly 153/154/209/212),
   0 removed, and one `area_code` -> one `area` label still holds.
   Benefit: 130,103 observed `faostat-production` rows stop being attributed to
   Rest of World for territories that do have their own balance sheet.

   Worth knowing before choosing: **nearly the whole cost of lifting the fold
   for all 61 is these four.** `stock_addition` 0.98862 for the four against
   #555's 0.9878 for all 61. The remaining 57 contribute a rounding error.
2. **Leave the fold and keep the honest label**, which is what this PR ships.

**What this implies for #459 and #563.** #563 lists three ways to redo the
reverted un-fold; this finding says which one applies to these four. Option 3
(un-fold at the *numeric* level) is not merely one alternative — for these four
it is the **only coherent one**, because `polity_area_code` is computed from
`fabio_code`, so correcting the factual column *is* the numeric un-fold. Options
1 and 2 (relabel or regroup while leaving `polity_area_code` at 999) are exactly
what #480 did and #561 reverted: `.aggregate_to_polities()` groups by
`polity_name`, so `999 Rest of World 340` splits into `999 Syrian Arab Republic
300` + `999 Rest of World 40`. **#459 should not un-fold these four at the
polity level alone.** Either both keys move or neither does; the
`"cbs_reporters"` option added here moves both, which is why it is the right
shape for the measurement.

Also worth noting for #544: `mapping_status` still says `"matched"` for all four
— it cannot express `"folded"`, let alone "folded against the source WHEP cites
for the fold".

Closes #556


lbm364dl:
## Measured: what lifting the fold for **just these four** costs

Two full-range `get_wide_cbs()` builds (1850-2023) on this branch, same session
semantics, ~7 min each, `options(whep.warn_polity_folds = FALSE)`:

```r
# baseline
get_wide_cbs()
# the four FABIO does not fold
options(whep.unfold_rest_of_world = "cbs_reporters"); get_wide_cbs()
```

| column | main | `"cbs_reporters"` | ratio | #555's all-61 ratio |
|---|---|---|---|---|
| production | 3.007430e12 | 3.006691e12 | 0.99975 | 0.9995 |
| feed | 9.050144e11 | 9.049936e11 | 0.99998 | 1.0000 |
| food | 3.143402e11 | 3.143458e11 | 1.00002 | 0.9999 |
| processing | 1.767704e12 | 1.766908e12 | 0.99955 | 0.9993 |
| seed | 5.031688e10 | 5.065738e10 | **1.00677** | 1.0004 |
| other_uses | 9.178742e10 | 9.176250e10 | 0.99973 | 0.9996 |
| import | 1.113820e11 | 1.113530e11 | 0.99974 | 0.9997 |
| export | 7.913412e10 | 7.917792e10 | 1.00055 | 1.0002 |
| domestic_supply | 3.146834e12 | 3.146315e12 | 0.99984 | 0.9996 |
| stock_addition | 3.935775e10 | 3.890993e10 | **0.98862** | 0.9878 |
| stock_withdrawal | 1.465130e11 | 1.463583e11 | 0.99894 | 0.9989 |

**The baseline column reproduces #555's `main` column to every printed digit**
on all eleven, which is the cross-check that the two measurements are
comparable.

Two things the comparison settles:

1. **Lifting the fold for these four costs at most 1.14%** (`stock_addition`),
   0.68% on `seed`, and under 0.05% on production, feed, food, processing,
   trade and domestic supply.
2. **Nearly the entire cost of lifting it for all 61 is these four.**
   `stock_addition` moves 0.98862 for the four against 0.9878 for all 61, so
   the remaining 57 areas contribute a rounding error. The "cost" side of the
   #419/#556 decision is essentially a decision about Syria, New Caledonia,
   North Macedonia and Eswatini — the four FABIO enumerates anyway.

`seed` is the one column where the four move more than all 61 did
(+0.68% vs +0.04%); worth a look before acting, since a promotion should not
increase a global total by more than promoting a superset of it does.

### Structure, not just totals

* **4 area codes added, 0 removed**: 195 -> 199 distinct `area_code`, the added
  ones being exactly `153, 154, 209, 212`. Rows 2,005,267 -> 2,028,421.
* **One `area_code` -> one `area` label** out of `.aggregate_to_polities()` in
  every mode — the #563 guard, driven on a 212/999/64/40 shape:

  | mode | output |
  |---|---|
  | default | `40 Chile 7`, `999 Rest of World 351` |
  | `"cbs_reporters"` | `40 Chile 7`, `212 Rest of World 300`, `999 Rest of World 51` |
  | `TRUE` | `40 Chile 7`, `64 Rest of World 11`, `212 Rest of World 300`, `999 Rest of World 40` |

  `max(labels per area_code) == 1` in all three. No bucket splits.

**A caveat about that third column, which predates this PR.** The option moves
`polity_area_code` but not `polity_code`, so a promoted area comes out under its
own code carrying the label *"Rest of World"*. That is the exact mirror image of
#480, which moved the label and not the code — and it is why neither is the
finished un-fold. Only correcting `regions_full$fabio_code` moves both at once,
because `polity_area_code` is derived from it. As a sensitivity switch this is
harmless (the invariant holds, nothing splits); as a production mode it would be
wrong, which is what the warning on every crosswalk read says.

### The noise floor these numbers sit on

#420 measures `get_wide_cbs()` as non-reproducible across sessions by ~0.148%
on rows, with **values moving in the fourth decimal** (`export` -0.008%, `food`
-0.003%, `feed` -0.001%, `production` -0.000%), and reports `processing` and
`stock_addition` as bit-identical across every sample. So:

* `stock_addition` -1.14%, `seed` +0.68%, `processing` -0.045% and `production`
  -0.025% are two to three orders of magnitude above that floor -- real.
* `feed` -0.0023% is only about twice the observed `feed` noise, and `food`
  +0.0018% is within it. Treat those two as "did not move".
* The +1.15% row count (2,005,267 -> 2,028,421) is well outside the 0.148% row
  noise, and is accounted for by the four added area codes.

Issue #565, opened 2026-08-05
balance_coefficients.R reads polity_end_year as inclusive: 282 (iso3, year) pairs silently claim a dissolved polity


eduaguilera:
A fourth site reads `polity_end_year` as inclusive, and unlike #550 this one is **silent** on the majority of its disagreements.

`data-raw/balance_coefficients.R:135-136` filters `to_year >= one_year` against `polity_end_year`, while the column is documented — and used almost everywhere else — as **exclusive**.

## Measured

Over `whep::polity_area_crosswalk`, comparing the inclusive read against the exclusive one:

| outcome | (iso3, year) pairs |
| --- | --- |
| inclusive read resolves to **two** polities → the function aborts | 296 |
| inclusive read silently claims a **dissolved** polity | **282** |
| **total disagreements** | **581** |

The 296 are loud and would be noticed. The **282 are not**: the boundary year resolves to the interval that ended on it, so a coefficient is attributed to a polity that no longer existed in that year, with no error and no warning.

## Why it is worth a separate issue from #550

#550 covers `add_polity_code()`. This is a different call site with a different blast radius: `balance_coefficients.R` is a `data-raw/` script, so fixing it **regenerates packaged `.rda` data**, and the resulting value movement needs measuring in its own right rather than being folded into an unrelated change. That is why it was left alone when the three live fenceposts in `R/` were fixed on the polycell branch (commit `f7675039`, which corrected `R/constant_territory.R`, `R/spatialize_compartments.R` and the `R/table_mappings.R` documentation that had said "Inclusive").

For context on the magnitude the convention carries elsewhere: fixing the two live `R/` sites moved **1,912 polity-years across 349 polities**, of which **1,103 moved off a `retired` or `superseded` label onto a live one**. It was not only relabelling — on a worked example (`build_constant_territory_series()` for Greece at 1881) the total value went from **114.85 to exactly 100.00**, because the dissolved epoch was attracting donor imputation while cells outside its smaller extent were being dropped entirely.

## A fifth, currently inert

`tests/testthat/test_datasets_balances.R:395` re-derives Spain's polity with `polity_end_year >= yr`. Harmless today — ESP is the single interval `ESP-1800-2025` and 0 of the 10 benchmark years disagree — but it encodes the wrong convention and would misresolve if ESP were ever split.

## Note on the convention itself

[...]

PR #577, opened 2026-08-05
fix(polities): read polity_end_year as exclusive everywhere

### LINKED ISSUE #550: add_polity_code reads polity_end_year as inclusive while everything else documents it exclusive: 3 area-years land in a dissolved state

lbm364dl:
## What was wrong

`polity_end_year` had two readings in this package and they disagreed by one year.

**Exclusive** — the documented and computed convention, and the one upstream
publishes:

- `R/table_mappings.R` (`polity_area_crosswalk` docs): "`polity_end_year` is
  exclusive."
- `R/polities.R` `.area_year_polity_conflicts()`: "a period covers
  `start:(end - 1)`", computed with `polity_end_year - 1L`.
- `R/polities.R` `resolve_polity_label()`: `live <- year >= start & year < end`.
- `data-raw/table_mappings.R` (`shadowed_by_map`): `polity_end_year - 1L >=
  map_year_start`, with a comment saying the mixed convention is deliberate.
- Upstream `whep-polities`: `scripts/validate_code_year_agreement.py` states
  "`end_year` is exclusive", its README says "a polity `end_year` is EXCLUSIVE",
  and issue eduaguilera/whep-polities#131 is open on exactly this seam
  ("This database defines `end_year` as **exclusive**: `CIV-1893-1900` covers
  1893-1899").
- The published data agrees: a successor's `start_year` equals its
  predecessor's `end_year`. `F51-1947-1993` hands over to `CZE-1993-2025` and
  `SVK-1993-2025`; `SUD-1956-2011` to `SDN-2011-2025` and `SSD-2011-2025`;
  `F248-1991-1992` to `SCG-1992-2006`; `SCG-1992-2006` to `SRB-2006-2008` and
  `MNE-2006-2025`.

**Inclusive** — what the resolver actually did: `.add_polity_columns_dt()` joined
on `join_end_year >= year`, so a row whose `year == polity_end_year` matched the
period. The `polities` `@format` block in the same file as the crosswalk docs
also called the years "Inclusive validity years", contradicting them.

At `origin/main` (2759be3e), over the 1961-2024 grid for all 266 crosswalk areas
(`add_polity_code(backcast_anchor = -Inf)`, 17,024 area-years), **7** resolve at
exactly `year == polity_end_year`:

```
  area_code             area_name year    polity_code polity_end_year mapping_status
1        15    Belgium-Luxembourg 1999  BLX-1850-1999            1999        matched
2        51        Czechoslovakia 1993  F51-1947-1993            1993         manual
3       151  Netherlands Antilles 2010  ANT-1961-2010            2010        matched
4       186 Serbia and Montenegro 2006  SCG-1992-2006            2006        matched
5       206        Sudan (former) 2011  SUD-1956-2011            2011         manual
6       228                  USSR 1991 F228-1945-1991            1991         manual
7       248          Yugoslav SFR 1992 F248-1991-1992            1992         manual
```

Rows 2, 4 and 7 are a year in a state that had already dissolved, and none of
them said so: `mapping_status` read `"matched"`/`"manual"`, not `"out_of_span"`.
That confirms the issue's table exactly.

**What separates the wrong three from the right four is measurable, not a
judgement call.** The crosswalk carries `map_year_end`, the last year upstream
declares the area *reports* under that period, inclusive. Across the 245
`upstream_map` rows, `polity_end_year - map_year_end` is:

| delta | rows | reading |
|---|---|---|
| 1 | 240 | `polity_end_year == map_year_end + 1` — exclusive |
| 0 | 4 | 15 Belgium-Luxembourg, 151 Netherlands Antilles, 206 Sudan (former), 228 USSR |
| 2 | 1 | 157 Nicaragua (open period, map just stops at 2023) |

So 240 of 245 rows state the exclusive convention outright. The four delta-0 rows
are the four the issue calls "correct by coincidence": their last *reported* year
equals their polity's exclusive end. The three wrong ones have map spans that
stop a year earlier (51: 1961-1992, 186: 1992-2005, 248: 1991-1991).

The same inclusive reading was in `build_constant_territory_series()`, which
selects source and target polities with `end_year >= yr`. On the shipped
snapshot that makes **238** polities carry a polygon in 1993 instead of 236,
Czechoslovakia sitting on top of Czechia and Slovakia; over 1850-2024 it
over-counts on **114 of 175 years**, **453** extra active polity-years in total
(3.97 per affected year). Each grid
cell is assigned to exactly one source and one target, so the dissolved
predecessor was capturing the cells its successors should have received.

## What I changed

1. `.add_polity_columns_dt()` joins on `join_end_year > year`.
2. New `.polity_join_end_year()` computes that bound as
   `max(polity_end_year, map_year_end + 1)`. The territorial span is the rule;
   the upstream map, which is the authority on reporting years, widens it where
   it declares a reported year past the polity's end. That is the 4 delta-0 rows
   and nothing else — a reported year is never dropped for being one past a
   polity's end, and two of those four (15, 151) would otherwise have gone to
   `NA` outright, because the nearest-period fallback deliberately skips
   aggregates.
3. The fallback's `year_distance` is exclusive-consistent: a row at
   `join_end_year` is 1 year past the period, not 0.
4. `build_constant_territory_series()`'s `.active()` uses `end_year > yr`.
5. The `polities` `@format` line no longer says "Inclusive"; the crosswalk docs
   and the `ref_year` docs state the half-open interval and the map-span
   widening.

Guarded: a crosswalk without `map_year_end` (a mock or a caller-supplied one)
falls back to the territorial span alone rather than erroring — that path is
exercised by `test_polity_folds.R`'s mocked crosswalk.

## How I verified

**The three wrong resolutions, after:** 51/1993, 186/2006 and 248/1992 now report
`"out_of_span"`. They still resolve to the same polity through the nearest-period
fallback, so no tonnage is dropped — the misattribution is now visible instead of
silent, which is what `"out_of_span"` exists for. The four map-declared years
(15/1999, 151/2010, 206/2011, 228/1991) keep `"matched"`/`"manual"`.

**Full-grid diff, 1850-2024 x 266 areas = 46,550 area-years**, before vs after,
same base grid both times:

| | anchor 1961 (default) | anchor -Inf |
|---|---|---|
| rows in / out | 46,550 -> 46,550 | 46,550 -> 46,550 |
| columns | identical | identical |
| resolved (non-NA `polity_code`) | 46,336 -> 46,336 | 46,225 -> 46,225 |
| `polity_area_code` moved | **0** | **0** |
| `polity_code` moved | 1 | 2 |
| `mapping_status` moved | 3 | 16 |

The one `polity_code` move under the default anchor is area 273 Montenegro in
1962, `MNE-1913-1918` -> `MNE-2006-2025`: a nearest-period stand-in
(`out_of_span`) both before and after, now landing on the nearer period under
the exclusive distance. The second, under `-Inf`, is area 107 in 1901,
`CIV-1893-1900` -> `CIV-1902-1932` — the very year upstream #131 names as
falling in no polity, so this is the exclusive metric picking the nearer of two
stand-ins.

**Structural checks, not just "no number moved" (#480/#563):**

- Driving the real `.aggregate_to_polities()` over the same 1961-2024 grid:
  12,953 rows in and out, total value 16,921 in and out, **one** cell differs —
  the area-273 label, `"Montenegro (1913-1918)"` -> `"Montenegro"`.
- One `area_code` -> one `area` label: `area_code`s carrying more than one label
  are **45 before, 45 after**; `(year, area_code)` keys carrying more than one
  label are **64 before, 64 after**, all of them bucket 206 (the Sudan fold of
  whep#414, pre-existing and untouched).
- `tests/testthat/test_read_raw_inputs.R`'s bucket-split guard: passes.
- `.area_year_polity_conflicts()`: **0** rows before and after. The 4 widened
  rows create no overlap — no other period of the same area covers the widened
  year (checked explicitly, 0 for each).

**Tests.** Reverting only the `R/` changes and keeping the new tests gives
**5 failures/errors**:

- `test_constant_territory.R:228` and `:229` — the dissolved `OLD` takes all 100
  and the successors get nothing (`"OLD"` instead of `c("NEW_L", "NEW_R")`;
  `100` instead of `c(50, 50)`).
- `test_polities.R:188` — the existing stand-in invariant, updated to the
  exclusive reading, catches 3 unflagged boundary years.
- `test_polities.R:204` — the three dissolved states read
  `"manual"`/`"matched"`/`"manual"` instead of `"out_of_span"`.
- `test_polities.R:246` — `.polity_join_end_year()` does not exist.

**Gates.** `air format .` (binary), `devtools::document()` (3 `man/` files
refreshed), `lintr::lint_package(...)` -> "No lints found", pkgdown
`man/*.Rd` vs `_pkgdown.yml` comm check -> empty, `devtools::test()` full suite
`[ FAIL 0 | WARN 11 | SKIP 8 | PASS 5751 ]`. `rcmdcheck` left to CI as
instructed.

## Moves published values

**No published value moves.** `polity_area_code` — the bucket every matrix
workflow aggregates on — is unchanged on all 46,550 area-years under both
anchors, the resolved-row count is unchanged, and `.aggregate_to_polities()`
conserves both its row count and its total. What moves is one bucket *label*
(area 273 in 1962) and three `mapping_status` values, from a status that claimed
a real period hit to one that admits a stand-in.

`build_constant_territory_series()` does move numbers when `ref_year` or a data
year is a hand-over year, which is the point: the dissolved predecessor no longer
takes its successors' grid cells. It is not called anywhere in the package's
pipelines, so nothing downstream shifts.

- **I did not extend the convention fix to upstream.** The four delta-0 rows are
  arguably an upstream inconsistency — a polity whose exclusive end equals the
  last year its area reports — and the clean fix would be upstream ending those
  four periods at `map_year_end + 1`. Until upstream decides that,
  honouring `map_year_end` here is the change that loses no reported year.
  Filed upstream as eduaguilera/whep-polities#164, with the 240/4/1 table and
  the three options, rather than widening this PR.
- **I did not touch `.area_year_polity_conflicts()`** to make it aware of the
  map-span widening. It reports 0 conflicts today and the 4 widened rows create
  none, so there is nothing to fix; a future vintage could change that.
- **I did not run any gridded build** (`build_nitrogen_balance()`,
  `build_carbon_balance()`, `build_water_balance()`): they need multi-GB local
  rasters and hours. None of them resolve polities at a period boundary in a way
  this could reach that `polity_area_code` invariance does not already cover.
- **I did not re-measure through a full `build_commodity_balances()`**, for the
  same reason. `.aggregate_to_polities()` is the seam every CBS/production read
  passes through, and it is driven for real above.

## The open decision

`build_constant_territory_series(ref_year = 2025)` now aborts with "No polities
with a polygon are active in `ref_year` = 2025", because the vintage's open
periods carry 2025 as their exclusive end and therefore stop at 2024. That is
correct under the convention and no WHEP data year reaches 2025, but it is a
behaviour change for a caller who passed the horizon year. The alternative would
be to treat the maximum `end_year` in the supplied table as an open sentinel that
does cover its own year — which reintroduces a second reading of the same field,
so I did not. Say if you would rather have the sentinel.

Closes #550

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,31 @@
 # whep (development version)
 
+* `polity_end_year` / `end_year` is now read as **exclusive** everywhere, which
+  is the convention upstream `whep-polities` publishes: a successor's
+  `start_year` equals its predecessor's `end_year`, and 240 of the 245
+  FAOSTAT-map rows in `polity_area_crosswalk` carry
+  `polity_end_year == map_year_end + 1`. `add_polity_code()` used to join on
+  `polity_end_year >= year`, so a period answered for one year past its end.
+  Over the 1961-2024 grid for all 266 crosswalk areas that put **7** area-years
+  on their period's end year, of which **3 landed in a state that had already
+  dissolved** and still read `"matched"`/`"manual"`: 1993 Czechoslovakia
+  (`F51-1947-1993`), 2006 Serbia and Montenegro (`SCG-1992-2006`) and 1992
+  Yugoslav SFR (`F248-1991-1992`). Those three now report `"out_of_span"`. The
+  other four are years the upstream map explicitly declares the area reports
+  (`map_year_end`, inclusive), and the resolver keeps them: a reported year is
+  never dropped for being one past a polity's end. **No published value moves**:
+  over 1850-2024 x 266 areas, `polity_area_code` is unchanged on every row, the
+  resolved-row count is unchanged (46,336 with the default back-cast anchor),
+  and the only `polity_code` that moves is area 273 Montenegro in 1962, from
+  `MNE-1913-1918` to `MNE-2006-2025` -- a nearest-period stand-in either way,
+  now landing on the nearer period. `build_constant_territory_series()` reads
+  the same convention, so a dissolved polity no longer sits on top of its
+  successors in the hand-over year (238 polities carried a polygon in 1993 on
+  the old reading against 236, and 453 extra active polity-years over
+  1850-2024), where each grid cell goes to exactly one target and the
+  predecessor was capturing the ones its successors should have received. Note
+  that `ref_year = 2025` now aborts: the vintage's open periods carry 2025 as
+  their exclusive end, so they stop at 2024.
 * `polities` and `polity_area_crosswalk` are re-synced against upstream
   `whep-polities` at `eb02dcb` (740 rows to **749**), which retired or superseded
   **14** codes this package had been treating as live and published a replacement
--- a/R/constant_territory.R
+++ b/R/constant_territory.R
@@ -34,7 +34,10 @@
 #'     in `year` and carry a polygon).
 #'   - `value`: numeric value (summed if a polity appears more than once).
 #' @param ref_year Integer. Target boundaries are the polities active in this
-#'   year (`start_year <= ref_year <= end_year`).
+#'   year. `end_year` is exclusive (see [polities]), so "active" means
+#'   `start_year <= ref_year < end_year` and a polity does not answer for the
+#'   year its successor takes over. The same reading selects the sources of
+#'   each data year.
 #' @param polities An `sf` of polity polygons with `polity_code`, `start_year`,
 #'   `end_year` and geometry. Defaults to [get_polity_geometries()].
 #' @param covariate `NULL` (uniform density, i.e. area weighting) or a function
@@ -130,8 +133,15 @@ build_constant_territory_series <- function(
   polities <- polities[!sf::st_is_empty(polities), ]
   polities <- sf::st_make_valid(sf::st_transform(polities, crs_equal_area))
 
+  # `end_year` is EXCLUSIVE (see [polities]), so a period covers
+  # `start_year:(end_year - 1)`. Reading it inclusively made a polity and its
+  # successors all active in the hand-over year -- 238 polities carry a polygon
+  # in 1993 on that reading, Czechoslovakia on top of Czechia and Slovakia, and
+  # 453 extra active polity-years over 1850-2024. Since `.assign_polity()` gives
+  # each cell exactly one source and one target, the dissolved predecessor was
+  # capturing the cells its successors should have received.
   .active <- function(yr) {
-    polities[polities$start_year <= yr & polities$end_year >= yr, ]
+    polities[polities$start_year <= yr & polities$end_year > yr, ]
   }
 
   target <- .active(ref_year)
--- a/R/polities.R
+++ b/R/polities.R
@@ -96,18 +96,49 @@
 # while the areas whose map span stops earlier (51 Czechoslovakia 1993, 186
 # Serbia and Montenegro 2006, 248 Yugoslav SFR 1992) no longer answer for a year
 # their polity had already ended in.
-.polity_join_end_year <- function(polity_end_year, map_year_end) {
+.polity_join_end_year <- function(polity_end_year, map_year_end, is_open) {
   territorial <- data.table::fifelse(
     is.na(polity_end_year),
     Inf,
     as.numeric(polity_end_year)
   )
+  # EXCLUSIVE AT A SUCCESSION, INCLUSIVE AT THE OPEN END.
+  #
+  # A boundary between two epochs belongs to the successor, which is what the
+  # exclusive reading buys. But a still-open interval has nothing after it, so
+  # there is no double-count to prevent and excluding its terminal year simply
+  # deletes a year. Measured on the shipped snapshot: 227 live polities end at
+  # 2025 and a strictly exclusive rule left NONE of them covering 2025, so every
+  # current-year row degraded from `matched` to `out_of_span` -- resolved only by
+  # the nearest-period fallback, which is the pathology this epic removes.
+  #
+  # Openness is detected by ABSENCE OF A SUCCESSOR, not by comparing the end year
+  # to the table maximum. The year test re-introduces the double-count for any
+  # polity whose last interval ends at the maximum AND has a successor, and the
+  # maximum itself moves (#530 took the table from 740 rows to 749). Measured:
+  # 244 live polities have no successor, 227 of them end at 2025, and ZERO live
+  # polities ending at 2025 have one -- so the two agree today and the successor
+  # test is the one that keeps agreeing.
+  # `fifelse()` will not recycle its test, so a scalar `is_open` (which is what a
+  # caller testing one period naturally passes) has to be widened here.
+  is_open <- rep_len(as.logical(is_open), length(territorial))
+  territorial <- data.table::fifelse(is_open, territorial + 1, territorial)
   reported <- data.table::fifelse(
     is.na(map_year_end),
     -Inf,
     as.numeric(map_year_end) + 1
   )
   pmax(territorial, reported)
+}
+
+# Which polity codes upstream declares nothing succeeds. Read from `polities`
+# rather than the crosswalk because succession is a fact about the polity, and
+# the crosswalk does not carry the relation.
+.open_polity_codes <- function() {
+  p <- polities
+  succ <- p$successor
+  open <- is.na(succ) | !nzchar(trimws(succ))
+  unique(p$polity_code[open])
 }
 
 .add_polity_columns_dt <- function(
@@ -169,7 +200,11 @@
           -Inf,
           as.numeric(polity_start_year)
         ),
-        .polity_join_end_year(polity_end_year, get("map_year_end")),
+        .polity_join_end_year(
+          polity_end_year,
+          get("map_year_end"),
+          polity_code %in% .open_polity_codes()
+        ),
         area_name,
         area_iso3c,
         polity_area_code,
--- a/R/polities.R
+++ b/R/polities.R
@@ -76,6 +76,40 @@
   out
 }
 
+# Exclusive upper bound of the years a crosswalk period covers, i.e. the period
+# runs `polity_start_year:(polity_end_year - 1)`.
+#
+# `polity_end_year` is EXCLUSIVE. That is what `whep-polities` publishes -- a
+# successor's `start_year` equals its predecessor's `end_year` (F51-1947-1993
+# hands over to CZE-1993-2025 and SVK-1993-2025; SUD-1956-2011 to SDN-2011-2025
+# and SSD-2011-2025), and 240 of the 245 FAOSTAT-map rows in the crosswalk carry
+# `polity_end_year == map_year_end + 1L` against an inclusive `map_year_end`.
+# It is also what `.area_year_polity_conflicts()`, `.polity_area_years()`,
+# `resolve_polity_label()` and the crosswalk build already compute with.
+#
+# `map_year_end` is the last year upstream declares the reporting area reports
+# under this period, inclusive, and the map is the authority on reporting years.
+# In the four rows where it reaches past the territorial span it wins, so a
+# reported year is never dropped for being one past a polity's end: four areas
+# whose last reported year equals `polity_end_year` (15 Belgium-Luxembourg 1999,
+# 151 Netherlands Antilles 2010, 206 Sudan (former) 2011, 228 USSR 1991) keep it,
+# while the areas whose map span stops earlier (51 Czechoslovakia 1993, 186
+# Serbia and Montenegro 2006, 248 Yugoslav SFR 1992) no longer answer for a year
+# their polity had already ended in.
+.polity_join_end_year <- function(polity_end_year, map_year_end) {
+  territorial <- data.table::fifelse(
+    is.na(polity_end_year),
+    Inf,
+    as.numeric(polity_end_year)
+  )
+  reported <- data.table::fifelse(
+    is.na(map_year_end),
+    -Inf,
+    as.numeric(map_year_end) + 1
+  )
+  pmax(territorial, reported)
+}
+
 .add_polity_columns_dt <- function(
   data,
   code_col = "area_code",
@@ -116,6 +150,11 @@
   if (!is.null(year_col) && year_col %in% names(dt)) {
     lookup <- .polity_crosswalk(include_unmapped = include_unmapped)
     lookup <- lookup[!is.na(area_code)]
+    # A caller-supplied or mocked crosswalk need not carry the upstream map's
+    # reporting years; without them the territorial span is the only bound.
+    if (!rlang::has_name(lookup, "map_year_end")) {
+      lookup[, "map_year_end" := NA_integer_]
+    }
     lookup <- lookup[,
       c(
         "area_code",
@@ -130,11 +169,7 @@
           -Inf,
           as.numeric(polity_start_year)
         ),
-        data.table::fifelse(
-          is.na(polity_end_year),
-          Inf,
-          as.numeric(polity_end_year)
-        ),
+        .polity_join_end_year(polity_end_year, get("map_year_end")),
         area_name,
         area_iso3c,
         polity_area_code,
@@ -178,7 +213,7 @@
       on = .(
         area_code,
         join_start_year <= year,
-        join_end_year >= year
+        join_end_year > year
       ),
       allow.cartesian = TRUE
     ]
@@ -232,12 +267,15 @@
         !is.na(polity_code) & get("lookup_polity_type") != "aggregate"
       ]
       if (nrow(fallback_matches) > 0L) {
+        # `join_end_year` is the exclusive upper bound, so the last year a
+        # period covers is `join_end_year - 1` and a row at `join_end_year`
+        # itself is already one year past it.
         fallback_matches[,
           "year_distance" := data.table::fcase(
-            year < join_start_year ,
-            join_start_year - year ,
-            year > join_end_year   ,
-            year - join_end_year   ,
+            year < join_start_year   ,
+            join_start_year - year   ,
+            year >= join_end_year    ,
+            year - join_end_year + 1 ,
             default = 0
           )
         ]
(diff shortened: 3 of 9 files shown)


eduaguilera:
We have collided — the same fix is committed on `edu/polycell-spatial-support` as `f7675039`, touching `R/constant_territory.R`, `R/polities.R`, `R/table_mappings.R` and `tests/testthat/test_constant_territory.R`, which is your file list almost exactly. Happy to drop ours in favour of this PR; posting the measurements first because two of them may not be in yours, and one is a gap this PR would leave open.

## The defect loses mass, not only labels

The relabelling is the visible half. The mechanism underneath is worse: on a boundary year both epochs come back active, and `.assign_polity()`'s "keep first match on overlap" hands every cell to whichever sorts first in `sf` row order. That ordering is systematic — measured, **the dissolved predecessor sorts first in 315 of 315 cases**, because `X-a-b` always precedes `X-b-c` lexically. So the successor gets **no row at all**, and cells lying outside the predecessor's smaller extent get `tgt = NA` and are **dropped**.

End to end through `build_constant_territory_series()` on real polygons, Greece at 1881:

```
before: three labels, total value 114.85   (100 covered + 14.85 donor imputation
                                            attracted by the dissolved epoch)
after : one label,   total value 100.00
```

90.1 of the 100 had been sitting on dissolved or superseded labels. So a fix here is not value-neutral in the way a pure relabelling would be, and the exclusive bound alone does not get you all of it — the same-polity dedupe is what removes the double-activation that causes the drop.

## Measured movement, decomposed

| effect | polity-years | polities |
| --- | --- | --- |
| exclusive bound alone | 665 | 349 |
| same-polity dedupe alone | 1,263 | 22 |
| **both** | **1,912** | **349** |

**1,103 of the 1,912 move from a `retired`/`superseded` label onto a live one.** Named: `RUS-1991-2014` → `RUS-2014-2025` and `UKR` likewise at 2014; USA at 1803/1848/1867/1959; CAN 1870–1885, sixteen consecutive years leaving the **retired** `CAN-1866-1886`; EGY 1821–1884 leaving the superseded `EGY-1800-1899`. Largest movers by count: BLZ 140, ARG 127, BRA 123, GRC 113, IRQ 94, EGY 84, ROU 84, BLX 79, PRY 70.

## The gap a strictly-exclusive read leaves

**All 229 still-open intervals end at 2025**, so under a uniformly exclusive rule **2025 has no active polity at all** — 2024 becomes the last covered year and `ref_year = 2025` aborts. We hit this immediately after making the bound exclusive.

The reading we settled on is **exclusive at a succession, inclusive at the open end**: a boundary between two epochs belongs to the successor, but a still-open interval covers its terminal year, because nothing succeeds it and no double-count is possible. Worth checking whether this PR has the same hole — and if you adopt the same rule, the trap is that "inclusive when `end_year` is the table maximum" re-introduces the double-count for any polity whose last interval ends at the maximum *and* has a successor. Detect still-open by absence of a successor, not by the year, especially with #551 moving the table to 749.

## Two more sites, and one caveat

- **`data-raw/balance_coefficients.R:135-136`** — filed as #565. 581 disagreeing `(iso3, year)` pairs: 296 abort loudly, **282 silently claim a dissolved polity**. Left out of our branch because fixing it regenerates packaged `.rda`.
- **`tests/testthat/test_datasets_balances.R:395`** re-derives ESP inclusively. Inert today (one interval, 0 of 10 benchmark years disagree) but it encodes the wrong convention.
- **Caveat on the tie-break**: we asserted "the named successor always wins" and it failed on 8 cases. **22 (family, start-year) pairs share a start year**; in 7 the loser is a `superseded`/`retired` duplicate the tie-break correctly passes over, but `MNE-1913-1915`/`MNE-1913-1918` are both live. Worth an explicit exception list rather than a blanket assertion.

Also flagging that the four pre-existing `test_constant_territory.R` blocks use deliberately **non-adjacent** epochs (SRC 1850–1950, targets 1990–2025), so they pass under either convention — which is why this survived. An adjacent-epoch fixture is what makes the test able to fail.

Tell me which way you want it: we drop `f7675039` and rebase onto this, or you take the measurements and we keep the branch-local fix until this merges.

lbm364dl:
Thank you — **the 2025 hole was real and this PR had it.** Reproduced exactly as you described, fixed with your rule, pushed as `5250cbcc`.

## Confirming your prediction

Before the fix, on this branch vs `main`:

| area | 2024 | 2025 (main) | 2025 (this PR) |
|---|---|---|---|
| 203 ESP | matched | matched | **out_of_span** |
| 229 GBR | matched | matched | **out_of_span** |
| 41 CHN | matched | matched | **out_of_span** |
| 100 IND | matched | matched | **out_of_span** |

Live polities covering 2024: **227**. Covering 2025: **0**. `add_polity_code()` masked it behind the nearest-period fallback, so nothing aborted — every present-day row just quietly stopped being a real period hit. That is the same pathology the epic keeps removing, so it would have been a genuine regression.

## Implemented your rule, including the trap you flagged

Exclusive at a succession, inclusive at the open end — and **openness detected by absence of a successor, not by the year**. Your warning about `end_year == max` was the useful part: the maximum moves (#530 just took the table 740 → 749), and the year test breaks for any polity ending at the maximum that *does* have a successor.

Measured on the shipped snapshot: **244 live polities have no successor, 227 of them end at 2025, and zero live polities ending at 2025 carry one.** So the two predicates agree today; a test now pins *why* they keep agreeing rather than asserting the coincidence.

Verified both directions: 2025 is `matched` again, and 51 in 1993, 248 in 1992, 186 in 2006 all stay `out_of_span` — the handover year belongs to the successor. Full suite **5779 pass, 0 fail**.

## On sequencing

**Please drop `f7675039` and take this**, if you are still happy to — the open-end rule is now in and the file lists overlap almost exactly, so keeping both would only mean resolving them against each other later.

Two things from your comment I have deliberately *not* folded in, so they do not get lost:

- **The mass loss.** Your Greece-1881 measurement (114.85 → 100.00, with 90.1 of the 100 sitting on dissolved labels, and the predecessor sorting first in **315 of 315** cases) is a stronger result than anything in this PR, and it says `.assign_polity()`'s same-polity dedupe is doing work the exclusive bound alone does not. If that half is on your branch, it is worth landing on top of this rather than being re-derived.
- **`data-raw/balance_coefficients.R:135-136`** (#565), where 282 pairs silently claim a dissolved polity. Out of scope here because it regenerates packaged `.rda` — agreed it should be its own PR.

One correction on your caveat: `MNE-1913-1915`/`MNE-1913-1918` are **no longer both live**. #551 landed the 740 → 749 re-sync and upstream has retired `MNE-1913-1915`, so that pair is down to one live polity. Worth re-measuring your 22 shared-start-year pairs against the new vintage before building an exception list — some of the 8 failures may have resolved themselves.

eduaguilera:
Agreed on all three points. **We will drop `f7675039` and take this**, and land the dedupe half on top rather than re-deriving it.

Your measurement is the better version of the rule, and worth stating for the record: **244 live polities have no successor, 227 of them end at 2025, and zero live polities ending at 2025 carry one** — so the year test and the successor test agree on today's snapshot, and pinning *why* they agree is exactly right. A test that asserts the coincidence would go green until the day the table changes shape, which is the failure mode we have hit three times on our branch.

## What we will land on top

The same-polity dedupe, since you have confirmed it does work the exclusive bound alone does not. The mechanism, so it is on the record here too: on a boundary year both epochs come back active and `.assign_polity()`'s "keep first match on overlap" hands every cell to whichever sorts first in `sf` row order — and that ordering is systematic, **the dissolved predecessor sorting first in 315 of 315 cases**, because `X-a-b` always precedes `X-b-c` lexically. The successor gets no row, and cells outside the predecessor's smaller extent get `tgt = NA` and are dropped. So the bug loses mass as well as relabelling it, which is why Greece at 1881 came to **114.85 rather than 100.00** — 14.85 of donor imputation attracted by a dissolved epoch, with 90.1 of the legitimate 100 sitting on dissolved or superseded labels.

Decomposed, so the two halves stay attributable: the exclusive bound alone moves **665 polity-years across 349 polities**; the dedupe alone moves **1,263 across 22**; together **1,912 across 349**, of which **1,103 move off a `retired`/`superseded` label onto a live one**.

## Thank you for the MNE correction — it lands on more than the caveat

You are right, and it invalidates more of ours than the exception list. Our **8-item shared-start-year exception list** was measured against the 740-row vintage, and if `MNE-1913-1915` is retired then at least one of the two genuinely-live cases has resolved itself. We will re-measure the 22 shared-start-year pairs against 749 before rebuilding the list — and per your own point, pin the property with an enumerated exception rather than a blanket assertion, so it cannot silently grow.

More usefully: **#551 landing unblocks a re-pin we had deliberately deferred.** Our producer's census figures — coverage classes, the terra-measured piece list, the S-A2 exception list, the long-edge table — were all measured on 603 rows and then on 740, and we held off re-deriving them precisely because the table was still moving. With 749 in, that re-pin can happen once instead of three times.

One thing we will re-check rather than assume: our producer currently reports **86 self-overlapping polity-years across 2 families** under the live filter (`MNE` 1913–1914, and `PER-1825-1909` coexisting with `PER-1825-1884` + `PER-1884-1909`, all three live `draft` — 84 years of double coverage). If the re-sync retired the MNE row, the Peru triple may or may not have moved with it. We will report the number on 749 either way.

lbm364dl:
Merged as `254c5668`. 5798 pass, 0 fail.

`f7675039` is now safe to drop, and the dedupe half is the next thing — your Greece-1881 decomposition makes the case on its own, and having the two halves attributable (665 polity-years across 349 polities for the bound; 1,263 across 22 for the dedupe; 1,912 across 349 together, 1,103 moving off a dead label) is exactly what makes it reviewable separately.

On the Peru triple: worth checking against 749 before you re-measure the census, because #560 changed what "live" means for the resolver in a way that may interact. `resolve_polity_label()`'s inference routes were indexing retired and superseded rows where the crosswalk build had always filtered them — so a family with a live `draft` row coexisting with its own finer split would have been ambiguous to the label route and invisible to the crosswalk. `PER-1825-1909` alongside `PER-1825-1884` + `PER-1884-1909`, all three live `draft`, is precisely that shape. It was one of the six ISO3 families whose ambiguity #560 resolved — but it resolved it by excluding *dead* rows, and if all three Peru rows are genuinely live upstream then the overlap is a real upstream defect that no downstream filter can fix, the same class as the MNE pair before it was retired.

So the 84 years of double coverage may well survive the re-sync. If your producer still reports it on 749, that is worth filing upstream rather than absorbing — I did not file it myself because your measurement is the one with the polygon evidence.

Issue #461, opened 2026-08-04
Feed crop/livestock spatialize the fractional cell-coverage grid, not the centroid grid


lbm364dl:
> [!IMPORTANT]
> **Delivered by the polycell work on branch `edu/polycell-spatial-support` (@eduaguilera)** — the consumer-side half of #460. Do not start.
>
> **The "decision needed" in this issue is settled: `polycell_id`, not majority-fraction overwrite.** Majority-fraction is centroid assignment under another name — it re-creates the very failure this issue opens with (a polity smaller than a cell gets nothing), just with a different tie-break. The polycell costs **+6.3% rows** (68,527 vs 64,438), which buys the guarantee that aggregating polycells to a polity changes no absolute value and no quantity crosses a border it does not belong to.
>
> **Corollary, and a separate acceptance criterion:** that guarantee only holds if *every* allocation is keyed on the polycell, never computed on the cell and apportioned afterwards. Computing from cell-average drivers and splitting after the fact hands a neighbour a share of whatever the cell produced.
>
> Measurements from that work: the centroid pin is genuinely one polity per cell (58,791 rows = 58,791 unique cells, no share column, so `.normalize_country_grid()` defaults `cell_area_frac = 1`). The two grids disagree **in both directions** — 5,651 cells in the fractional grid and not the centroid one, and **4 cells the other way** (`(-75.75, 22.25)` a/c 12, `(98.25, 11.25)` a/c 28, `(131.75, -4.75)` a/c 101, `(-80.75, -33.75)` a/c 40) which are carbon-path rows today, so a naive switch silently deletes them. Small polities are worse than "zero cells": French Polynesia 18 polycells, Kiribati 9, Micronesia 4, Maldives 1, all with zero LUH2 terrestrial area, so each draws a full cell and over-counts land by roughly 13×, 34×, 18× and 10×.
>
> **This will expose polygon-quality problems underneath it** — `MDV-1800-2025` measures 24.4 km² against an official 300 km², `MCO-1800-2025` 19.7 km² against 2.08 km². Those belong upstream in `whep-polities`, not here.
>
[...]

PR #579, opened 2026-08-05
fix(spatialize): let the shared cell grid be keyed on the polity bucket

### LINKED ISSUE #460: Route the shared cell↔area grid (build_cell_polity/build_cell_polity_fraction) through the polity crosswalk

lbm364dl:
## What was wrong

`build_cell_polity()` produces the shared cell/area grid every gridded consumer
keys on — `build_water_balance()`, `get_soc_climate_drivers()`,
`build_ag_land_support()`, `build_n_deposition()`, `build_urban_n()`,
`spatialize_country_n_to_crops()`, `aggregate_grass_to_polity()`, the HWSD
clay/pH path. It reads `cell_polity_fraction.parquet` verbatim.

That parquet is rasterized from present-day polygons through
`inst/extdata/regions.csv`, so its `area_code` is a **raw reporting-area code**,
not the `polity_area_code` **bucket** every polity-keyed national table in whep
is aggregated on. Where the two vocabularies disagree, the join between the grid
and national data is silently empty *on both sides*: the grid cells get nothing
and the national total is placed nowhere.

At `origin/main` (`6e5af760`), against the real deployed parquet at
`WHEP_POLITY_FRACTION_PATH` and real `get_primary_production()` output.

The parquet measures **62,808 rows / 58,791 cells / 182 area codes**. **12 of
those 182 codes are not a `polity_area_code`**, covering **819 cells**:

| grid `area_code` | area | bucket it belongs to | cells |
|---|---|---|---|
| 62 | Ethiopia PDR | 238 Ethiopia | 411 |
| 277 | South Sudan | 206 Sudan (former) | 248 |
| 212 | Syrian Arab Republic | 999 Rest of World | 94 |
| 205 | Western Sahara | 999 | 18 |
| 154 | North Macedonia | 999 | 17 |
| 61 | Equatorial Guinea | 999 | 13 |
| 209 | Eswatini | 999 | 12 |
| 299 | Palestine | 999 | 9 |
| 153 | New Caledonia | 999 | 6 |
| 6 | Andorra | 999 | 4 |
| 125 | Liechtenstein | 999 | 1 |
| 192 | San Marino | 999 | 1 |

The set of grid codes that no 2010 national cropland row can reach is **exactly
that same set of 12** — the two directions are the same defect. Measured on real
2010 harvested area from `get_primary_production()` (189 national codes,
1,383.21 Mha world total):

```
19 of 189 national codes cannot be placed on any grid cell
21,094,469 ha = 1.525% of world 2010 cropland
  238 Ethiopia    15,297,195 ha   (grid carries 62, national carries 238)
  999 RoW          5,672,558 ha   (grid carries the 10 members individually)
  + 17 island/microstate codes with no cell at all, 124,716 ha total
```

So **the whole of Ethiopia's cropland and the whole of Rest of World are
unspatializable today**, and nothing warns. The remaining 17 are the separate
zero-cell problem #461 tracks; this PR does not claim them.

Honest caveat found while measuring: `regions.csv` **today** maps `ETH -> 238`
and `SDN -> 276`, so 11 of the 12 reproduce from the current producer while
Ethiopia's `62` is a stale artifact of the parquet on disk (built from an older
lookup). Both are real in production, because that file is what every consumer
reads — and both are healed reader-side.

Also corrected: the "CONFIRMED FACTS" header of `R/n_balance_spatialize.R`
asserted the parquet has 68,527 rows. It has 62,808; 68,527 is the pre-#381
vintage. The literal is replaced with the reason not to assert one.

## What I changed

`build_cell_polity()` gains `area_key = c("grid", "polity_area")`.

- **`"grid"` (default)** — today's behaviour bit-for-bit, plus a `cli::cli_warn()`
  naming the off-bucket codes, their cell count, and the argument that fixes it.
- **`"polity_area"`** — resolves each code to its bucket through
  `polity_area_crosswalk` and **re-sums `polity_frac` within `(lon, lat,
  area_code)`**, so a cell holding two areas of one bucket (Sudan/South Sudan)
  collapses to one row instead of duplicating the key. Codes absent from the
  crosswalk keep their own code, so a gap stays visible rather than being
  dropped.

It reads the crosswalk through `.polity_crosswalk()`, not the shipped table, so
`options(whep.unfold_rest_of_world = TRUE)` reaches the grid too and the grid and
the national tables cannot disagree about where a Rest-of-World member's rows
belong. A code mapping to two buckets aborts rather than picking a winner.

## How I verified

Gates, all on this branch:

- `air format .` (binary), then `devtools::document()` — `man/` committed.
- `lintr::lint_package(...)` with the four disabled linters: **No lints found.**
- `devtools::test()` full suite: **5712 passed, 0 failed, 0 errors**, 8 skipped.
- pkgdown: only `man/build_cell_polity.Rd` changed, an existing `_pkgdown.yml`
  entry — no new topic, nothing to add.
- No new `utils::globalVariables()` entries needed (`lon`, `lat`, `area_code`,
  `polity_frac`, `polity_area_code` are all already declared). No non-ASCII.

Assertions that fail without the change: reverting only
`R/n_balance_spatialize.R` and re-running the file gives **1 failure + 4 errors,
43 passing**; with the change, **53 passing, 0 failed, 0 errors**.

**Structural checks, not just totals.** The epic is explicit that a totals diff
proves nothing here, and #480 proved "no number moved" is not evidence. Driving
the real function over the real deployed parquet, `"grid"` vs `"polity_area"`:

| | `"grid"` | `"polity_area"` |
|---|---|---|
| rows | 62,808 | 62,774 (34 Sudan border rows collapse) |
| **cells** | 58,791 | **58,791, byte-identical set** |
| area codes | 182 | 172 |
| per-cell `sum(polity_frac)` | exactly 1 | exactly 1 |
| one row per (cell, code) | yes | yes |
| off-bucket codes | 12 | **0** |
| total land | 13,205,203,638 ha | **13,205,203,638 ha** |

Per-`code` **L1 distance between the old and new normalised cell-share vectors**
(the discriminating measurement the epic asks for): **15 of 184 codes move, 169
are exactly 0.0**. The 15 are the 12 off-bucket codes (L1 = 1.0, they cease to
exist as keys), their two destinations 238 and 999 (L1 = 1.0, they come into
existence), and 206 (L1 = 0.4956, it absorbs South Sudan).

**Change in distinct cells receiving a non-zero allocation**: the 12 go to 0;
`238` gains 411, `999` gains 175 (= 4+13+1+6+17+1+18+12+94+9, the members are
disjoint so nothing double-counts), `206` goes 667 -> 881.

**One `area_code` -> one `area` label**: the #480 guard in
`tests/testthat/test_read_raw_inputs.R` passes (7/7). It is unaffected by
construction — this PR touches neither `.aggregate_to_polities()` nor any
`polity_name`/`area` label, and adds no name column to the grid.

What it buys, end to end on real 2010 data: national cropland the grid cannot
place goes from **21.09 Mha (1.525%) to 0.12 Mha (0.009%)**.

## Moves published values

**No — not on the default path.** `area_key = "grid"` is the default and
reproduces today's output bit-for-bit; the only change on that path is a warning.
Nothing downstream moves without an explicit `area_key = "polity_area"`.

The alternative *does* move published values for every gridded consumer at once,
which is exactly why it is opt-in rather than the default.

- **I did not add `polity_code` / `reporting_polity_*`**, which the issue's
  original acceptance criteria asked for. Epic #458 withdrew that framing and it
  is right to: a bucket is not a polity. `999` holds up to 17 territories
  simultaneously and `206` holds Sudan and South Sudan together, so no polity
  code string is recoverable from this year-less grid. Writing one would write a
  false identity. That belongs to the cell x polity x validity-interval unit the
  epic tracks, not a column added here.
- **I did not touch `build_cell_polity_fraction()`** in
  `inst/scripts/prepare_spatialize_all.R`. Its `area_code` is geographically
  faithful — the NaturalEarth polygon really is Syria, not Rest of World —
  and changing the producer means regenerating and redeploying the pin, with no
  test able to see it. Resolving in the reader is testable, reversible, and
  additionally heals the stale `62`/`206` codes already baked into the file on
  disk.
- **I did not add a `method_*` output column**, which CLAUDE.md's multi-method
  rule would normally require. `data$cell_polity` is a shared object that
  `.wb_attach_polity()` joins wholesale into multi-million-row gridded tables; a
  stray character column would propagate into gridded outputs. Flagging this as a
  conscious deviation — the choice is instead surfaced by the argument, the
  documented `@section`, and the warning.
- **I did not attempt any full-range gridded build** (`build_nitrogen_balance()`,
  `build_carbon_balance()`, `build_water_balance()`) — hours and multi-GB local
  rasters, out of scope per the epic.
- **I did not duplicate the polycell work.** Note for the record: branch
  `edu/polycell-spatial-support` **does not exist** in `eduaguilera/whep` (no
  local ref, no remote ref, no PR, and `build_polycell_support` appears nowhere
  in the tree). `polycell_id` support does already exist in the *engine*
  (`.normalize_country_grid()`, `R/spatialize_compartments.R:39`), unused by this
  grid. Nothing here conflicts with adopting a polycell table later: this changes
  which numeric key the existing grid carries, and a polycell table would replace
  the grid wholesale.

## The open decision

**Should `area_key = "polity_area"` become the default?**

Arguments for: it is the only key that can join to whep's polity-keyed national
data at all, and today's default silently strands 1.525% of world cropland
including all of Ethiopia.

Argument against, and why I did not flip it: it re-attributes 15 codes' cells in
one commit across ~10 unmigrated gridded consumers, making per-consumer
attribution impossible — the exact failure mode #480/#563 documented. Flipping it
should be its own PR, per consumer or with a full gridded before/after.

A second, smaller decision underneath it: the fold sends Syria, Palestine,
Eswatini and seven others into `999`, which is faithful to the current national
vocabulary but is not a territory. If #419's unfold is ever adopted, this grid
follows it automatically (it reads `.polity_crosswalk()`), which is the intended
coupling — but it is worth stating that the coupling exists.

Closes #460


eduaguilera:
Flagging a collision and a methodological tension, from the polycell spatial-support branch (`edu/polycell-spatial-support`), which is mid-migration on exactly this object.

## The collision

`data$cell_polity`'s contract is load-bearing for that branch. It is a **shared object** passed to `build_n_deposition()`, `build_urban_n()`, `spatialize_country_n_to_crops()`, `build_water_balance()`, `read_soil_ph()`, `read_soil_hydraulic()` and the carbon clay path, which is why the migration ships a **superset shim**: `build_polycell_support()` emits `polity_frac` and `cell_area_ha` as derived columns that reproduce today's crosswalk **bit-for-bit**, so every unmigrated consumer is provably unchanged while consumers move one at a time.

That claim is asserted with `identical()`, not a tolerance, and it was re-measured after merging 138 commits of `main`: 68,527 rows both sides, `identical(whole table)` TRUE, fingerprint 68,527 / 64,438 / 191, `min(polity_frac)` identical to `1/36`. **If this PR changes what `build_cell_polity()` returns, that assertion breaks and the shim has to be re-derived against the new contract.** Not an objection — just something we need to sequence rather than discover.

## The tension worth settling deliberately

This PR re-keys the shared grid onto `polity_area_code`, the aggregation **bucket**. We went the other way for the polycell table and it is worth putting the reasoning side by side, because both cannot be right for the same column.

Epic #458's own measurement is the argument against the bucket as an identity: **544 distinct `polity_code` collapse into 201 `polity_area_code`**, bucket **999 holds up to 17 territories simultaneously**, and bucket **206 holds Sudan and South Sudan simultaneously 2011-2024**. #458 states it directly — *"`polity_area_code` is not a polity."*

So for a table whose entire purpose is correct territorial attribution, keying on the bucket bakes an irreversible fold into every gridded consumer downstream: once a cell is labelled 206, nothing later can tell whether its nitrogen belongs to Sudan or South Sudan. We chose to have `build_polycell_support()` carry the real periodized `polity_code` and let consumers convert at their own boundary — so the fold still happens, but it happens visibly at the join rather than invisibly in the grid.

The counter-argument for this PR is real too: the join to national tables is empty on both sides today, and a bucket-keyed grid fixes that immediately. I am not arguing against fixing it. The question is whether the grid should carry **both** — the bucket for the join, the polity code for attribution — so that the fold is recoverable rather than baked in.

Related: **#582** ("gridded land-use/livestock output carries two disagreeing territorial keys") is the same seam from the output side, so whatever is decided here should settle that too.

## Offer

Whichever way it goes, I can supply from our side: the measured shim contract (what exactly changes if `area_code` becomes the bucket), and the per-consumer contract delta we took across #487 — all five example-bearing consumers gained exactly four columns via a trailing `.add_reporting_polity_columns()`, with computations untouched. Say the word and I will post the diff.

lbm364dl:
Holding this one rather than merging, and marking it draft. Your objection is the stronger argument and it is not mine to overrule.

## Why I am not merging it

The epic states the rule you are invoking, in its own words: **"`polity_area_code` is not a polity."** Its measurements are the reason — 544 distinct `polity_code` collapse into 201 buckets, 999 holds up to 17 territories simultaneously, 206 holds Sudan and South Sudan 2011-2024. For a table whose entire purpose is territorial attribution, keying on the bucket makes that fold irrecoverable at every gridded consumer downstream, which is a different and worse thing than performing the fold visibly at a join.

That is not hypothetical today. A sibling PR (#591, for #563) has just measured the bucket-206 fold **corrupting data in both directions** in the published production series — Sudan's 40.8 M goats dropped in 2018, sugar cane doubled in 2019 — precisely because two territories share one bucket. Baking that bucket into the grid key would put the same ambiguity somewhere much harder to unwind.

Your "carry both" proposal — bucket for the join, `polity_code` for attribution — resolves the tension without giving up the fix, and it is the shape I would want. But it changes `build_cell_polity()`'s contract, which is exactly the `identical()` shim assertion you say has to be **sequenced rather than discovered**. So the sequencing call is yours.

## What I would find most useful

Yes to the offer, please: **the measured shim contract** — what exactly changes if `area_code` becomes the bucket — and the per-consumer contract delta from #487. With those, whoever picks this up can implement "carry both" against a known baseline instead of rediscovering it.

## Meanwhile

The defect this PR reports is real and still open — the join to national tables is empty on both sides today. I have left the PR open and un-merged rather than closed, so the reproduction and the measurement are not lost. **#582** should be settled in the same decision, as you say.

For context on the rest: 20 polity PRs merged today and `main` is at 5779 passing. #577 has taken your open-end rule (exclusive at a succession, inclusive at the open end, detected by absence of a successor) and is green. This one is the only PR of the batch I am holding on a design question rather than a measurement.

eduaguilera:
Here is what you asked for. The #591 goats-and-sugar-cane result is the decisive evidence — that is the same fold corrupting published series in both directions, and it settles the design question better than the argument from the epic did.

## The measured `data$cell_polity` contract, today

`build_cell_polity()` returns `lon, lat, area_code, polity_frac, cell_area_ha`. Measured on the deployed parquet, and **re-measured after merging 138 commits of `main` — byte-identical across the merge**:

```
rows 68,527 | cells 64,438 | area_codes 191
per-cell sum(polity_frac): min = max = 1 (2 cells at 0.99999999999999989, max |sum-1| = 1.11e-16)
min(polity_frac) identical() to 1/36  ->  TRUE      (subcells = 6L, not the 1/144 often assumed)
cells by polity count: 60,513 / 3,764 / 158 / 3
```

The shim asserts `identical()` on the whole table, not a tolerance. So "what changes if `area_code` becomes the bucket" is answerable exactly:

**Rows whose `area_code` would change value**, from our polycell run against `polity_area_crosswalk`:

- **41 cells / 12.18 Mha** fold two distinct `polity_code`s onto **206** — the Sudan/South Sudan case #591 just caught in production.
- **3,568 polycell rows / 280.70 Mha (2.114% of territory)** have **no `polity_area_code` at all**: Greenland 213.2 Mha, Western Sahara 26.8, Syria 18.7, French Guiana 8.4. Under a bucket key these become `NA` rather than folding.
- 16 live-2015 polities carry `NA area_code`, and **9 deployed area codes receive no polycell**: Fiji 66 (unrepairable polygon, now fixed upstream in whep-polities#147), plus NCL 153, MKD 154, SWZ 209, SYR 212, GNQ 61, SDN 276, SSD 277, PSE 299 — the #459 set.

Two footprint facts that bear on "carry both": **1,373 cells are crosswalk-only and 1,401 are polycell-only**. So a bucket column added to the current grid is not simply a relabelling of the same rows.

## The per-consumer contract delta from #487

Measured pre-merge snapshot vs merged tree. All five example-bearing consumers gained **exactly four columns** and a reordering (`year, area_code, <4 polity cols>, lon, lat, …`), via a single trailing `.add_reporting_polity_columns()` — **computations untouched**:

| consumer | before | after | added |
|---|---|---|---|
| `build_n_deposition()` | 7 | 11 | `polity_area_code`, `reporting_polity_code`, `reporting_polity_name`, `reporting_polity_has_geometry` |
| `build_urban_n()` | 6 | 10 | same 4 |
| `build_water_balance()` | 18 | 22 | same 4 |
| `build_carbon_balance()` | 13 | 17 | same 4 |
| `read_luh2_landuse()` | 7 | 11 | same 4 |
| `build_ag_land_support()` | absent | 11 | new on `main` (`7701ebb2`) |
| `spatialize_country_n_to_crops()` | 4 | **unchanged** | none — no `example` formal, so outside #487's census |

One consequence worth having: **`build_urban_n()` additionally routes `area_code` through `.manure_territory_to_area_code()`**, which is a behaviour change rather than an added column, and it is what broke our characterisation fixture.

Also: **the polycell exports are invisible to #487's gate** — `test_exported_polity_columns.R` enumerates exports with an `example` formal, and none of our eight has one.

## On sequencing, since you left the call to me

Our branch can absorb a `build_cell_polity()` contract change; it just has to be a **re-derivation with the shim's `identical()` re-established against the new contract**, not a surprise. So: land "carry both" whenever suits you, and I will re-derive. What I would ask is that the bucket arrive as an **added** column rather than a **replaced** `area_code`, so the raw code stays recoverable — that is the same asymmetry #582 is complaining about, and it keeps the fold visible at the join instead of irrecoverable in the grid.

If it helps, our producer already carries `polity_code` per cell, so a cell-level bucket↔polity mapping is derivable from it directly rather than through the crosswalk — happy to hand that over if it saves you the join.

lbm364dl:
Done — `grid_area_code` is added rather than `area_code` replaced, exactly as you asked. Pushed; full suite **5810 pass, 0 fail**.

Under `area_key = "polity_area"` the output now carries the parquet's own code in `grid_area_code`, joined with `+` where a cell's areas collapse into one bucket. Where they collapse I join rather than pick, because picking is the silent half of the same problem. The `"grid"` default is untouched and still reproduces the parquet bit-for-bit, so **nothing moves by merging this** — your shim's `identical()` assertion holds unless a caller opts in.

Your measurements changed my read of the risk here, in particular the **3,568 polycell rows / 280.70 Mha (2.114%) with no `polity_area_code` at all** — Greenland 213.2, Western Sahara 26.8, Syria 18.7, French Guiana 8.4. Under a bucket key those go `NA` rather than folding, which is a different failure from the one this PR set out to fix and would have been easy to miss. And **1,373 cells crosswalk-only against 1,401 polycell-only** settles that a bucket column is not a relabelling of the same rows — worth stating in the PR body, which I have not yet rewritten.

Two things I am leaving to you rather than guessing:

- **When to re-derive the shim.** The contract only changes for opt-in callers, so there is no forcing function; land it whenever the re-derivation suits your branch.
- **`build_urban_n()` routing `area_code` through `.manure_territory_to_area_code()`** — you flag that as a behaviour change rather than an added column, and it broke your characterisation fixture. That is not from this PR and I have not touched it, but it is the one item in your #487 table that is not "four columns and a reorder", so it probably deserves its own issue rather than living in a comment thread.

Also noting your point that **the polycell exports are invisible to #487's gate**, since `test_exported_polity_columns.R` enumerates exports with an `example` formal and none of your eight has one. That is a real hole in a test I have been treating as a census — same class as the guard I added in #566 turning out to be too narrow to catch the live bucket-206 split. Worth widening the enumeration to all exports and skipping the ones that cannot be exercised, rather than silently omitting them.

Leaving this in draft for you to sequence.

lbm364dl:
Merging, having verified it does what you asked rather than assuming it.

Both conditions, checked against `main` on a fixture with a folded cell (276+277 into 206), a plain area, and a Rest-of-World member:

**1. The default is untouched.** `area_key = "grid"` returns the same columns in the same order with the same values as `main` — `lon, lat, area_code, polity_frac, cell_area_ha`, no new column. So your shim's `identical()` assertion holds unchanged, and nothing re-derives until a caller opts in.

**2. The raw code is added, not replaced.** Under `"polity_area"`:

```
   lon   lat area_code polity_frac grid_area_code
 40.25 20.25       203           1            203
 30.25 10.25       206           1        276+277
 50.25 30.25       999           1            212
```

`area_code` carries the bucket, `grid_area_code` carries the parquet's own code, and the folded cell keeps **both** members rather than one being picked. `polity_frac` re-sums to 1 across the fold. That is the asymmetry #582 complains about, not reproduced here.

Full suite **5955 pass, 0 fail**.

Merging now rather than holding, because you said "land it whenever suits you, and I will re-derive" and the opt-in default means there is no forcing function on your timing — the re-derivation is needed only when a consumer actually switches.

Two things from your comment still open, and both are yours rather than blocked on this:

- **`build_urban_n()` routing `area_code` through `.manure_territory_to_area_code()`** — a behaviour change rather than an added column, and the one item in your #487 table that is not "four columns and a reorder". It broke your characterisation fixture and it is not from this PR. Probably wants its own issue.
- **The polycell exports being invisible to #487's gate**, since `test_exported_polity_columns.R` only enumerates exports with an `example` formal. That is the second test-coverage hole found today by the same mechanism — my own guard in #566 used a bucket that could never fail it. Both are cases of a check that looks like a census and is not.

Thanks for the shim contract and the #487 delta — the **3,568 polycell rows / 280.70 Mha with no `polity_area_code` at all** was the number that changed my read of the risk here, and the **1,373 crosswalk-only vs 1,401 polycell-only** cells settled that a bucket column is not a relabelling of the same rows.

Issue #171, opened 2026-07-22
gapfilling: order-dependence and smoothing edge-case robustness in fill_linear


lbm364dl:
## What

Several robustness gaps in `fill_linear`/`gapfilling.R`, all silent or crash-on-edge:

- **`:118-142`** the ungrouped `fill_linear` path never sorts by `time_col` (the grouped path does `setkeyv(dt, c(by, time))`). Carry-backward/forward and `first_valid`/`last_valid` are *positional*, so unsorted input (e.g. 2020 first) carries a late value into earlier years and `zoo::rollmean` averages non-adjacent years. Inconsistent with the grouped path.
- **`:305-329`** `.fill_linear_by_group` crashes when smoothing leaves no valid anchor: `c(1, NA, 2, NA, 3)` with window 3 → `sv` all-NA → `first_v <- valid[1L]` is `NA` → `if (fill_backward && first_v > 1L)` is `if (NA)` → hard error. The ungrouped path guards this (`:415-418`); the grouped path doesn't.
- **`:1184-1235`** `.fg_growth_aggregate`: weights are shifted *after* filtering to non-NA growth rows, so the "previous year weight" is the weight of the previous *surviving* row (can be years away) whenever the proxy series has holes; the first surviving row per group is also silently excluded.
- **`:169-176`** trust-based sort skip: if `attr(data, ".whep_sorted_by")` matches, the code stamps `sorted` without verifying — a caller that reordered the data between calls gets LOCF/NOCB over the wrong order.

## Fix

Sort the ungrouped path by `time_col`; guard the empty-anchor case in the grouped smoothing path; shift weights before dropping NA-growth rows; add a cheap `.is_sorted_by` verification instead of trusting the attribute. Add order-invariance and smoothing-window edge fixtures.

---
_From a static read-only code audit (2026-07-22). Line references from source reading, not verified by running — confirm with a test before fixing. Related to #131._

PR #606, opened 2026-08-06
fix(gapfilling): repair the weighted proxy growth aggregation

### LINKED ISSUE #324: gapfilling: weighted proxy growth ("var[weight]" / "var:group[weight]") has a setnames mismatch and a cross-group weight shift
Surfaced while fixing #150 (which made the `var:group` grouped-proxy syntax reachable).

Two independent pre-existing bugs live in the **weighted** branch of `.fg_growth_aggregate()` in `R/gapfilling.R`. They affect the already-documented simple weighted form `"gdp[weight]"` as well as the newly-reachable grouped weighted form `"var:group[weight]"`:

1. **`setnames(c("V1","V2"))` mismatch** — data.table auto-names the aggregated columns `g_val` / `o_val`, not `V1`/`V2`, so the rename targets the wrong columns.
2. **Weight `shift()` crosses the finer boundary** — the weight lag is grouped by the coarse group, so it crosses the finer (per-series) boundary and drops the first row's weight.

Not fixed in PR #322 (scoped to the #150 reachability defect; the unweighted `var:group` path is tested and correct there). These need their own fix + tests on the weighted aggregation path. Related to #131.

lbm364dl:
## What was wrong

`fill_proxy_growth()` documents two weighted proxy forms — `"var[weight]"` and
`"var:group[weight]"`. Neither worked, and behind the crash the weighting
itself was wrong.

**1. `setnames()` mismatch (a hard abort).** The weighted branch of
`.fg_growth_aggregate()` returned `list(g_val, o_val)` from a `{}` block and
then renamed `c("V1", "V2")`. data.table names such columns after the symbols
in the `list()`, not `V1`/`V2`, so *every* weighted spec aborted:

```
> fill_proxy_growth(data, value, proxy_col = "gdp[pop]", .by = "country")
Error in data.table::setnames(result, c("V1", "V2"), c(growth_col, obs_col)) :
  Items of 'old' not found in column names: [V1, V2]. Consider skip_absent=TRUE.
```

Confirmed on `origin/main` for both `"gdp[pop]"` and `"gdp:region[pop]"`.

**2. The weight lag was wrong twice over.** With (1) patched locally, the
weighted numbers came out wrong. The lag was taken inside
`.fg_growth_aggregate()`, i.e. *after* `.fg_growth_calc_individual()` had
dropped every row with no individual growth rate, and it was grouped by the
**coarse aggregation group** rather than by the individual series. Three
consequences, all reproduced:

- the first member of each group got `w = NA` and was excluded from the
  weighted mean entirely (and from `n_obs`);
- the next member picked up the **previous member's** weight — a cross-series
  leak;
- a member with an interior gap in the proxy was weighted by the **wrong
  year**, because `shift()` stepped over the dropped rows.

Evidence, with only the `setnames` line patched so the numbers are visible.
Fixture: region `eu` = ESP (`gdp` 100/110/121, `pop` 10) + FRA (`gdp` flat,
`pop` 90), region `am` = USA alone; `proxy_col = "gdp:region[pop]"`,
`.by = "country"`.

| series | year | on main (`setnames` patched) | correct |
| --- | --- | --- | --- |
| ESP | 2001 | 50.00 | 50.50 |
| ESP | 2002 | 50.50 | 51.005 |
| FRA | 2001 | 80.00 | 80.80 |
| FRA | 2002 | 80.80 | 81.608 |

The `eu` weighted growth should be `(0.10·10 + 0.00·90) / 100 = 0.01` in both
years; it came out as FRA's `0.00` in 2001, because ESP's weight was `NA` and
FRA had inherited ESP's `pop = 10`. The ungrouped `"gdp:[pop]"` form was off
by more: growth `0.18` in 2001 against the correct `11/150 ≈ 0.0733`. With an
interior gap in ESP's `gdp`, the 2004 weight was taken from 2001 (`pop = 2`)
instead of 2003 (`pop = 4`), giving `50.83` against the correct `51.90`.

## What changed

`R/gapfilling.R`

- `.fg_growth_aggregate()` names its aggregated columns explicitly in both
  branches, so the rename can no longer miss.
- The previous-period weight is now carried as `lag_weight` alongside
  `lag_src` in `.fg_growth_calc_individual()`, where every period is still
  present, and lagged **within the series** (`.by` plus the proxy's own
  grouping columns) rather than within the aggregation group.
- Three private helpers extracted so both call sites share one definition:
  `.fg_lag_one_period()` (lag a fixed set of columns in one pass, keeping
  every lag pointing at the same source row), `.fg_weighted_growth()` (the
  weighted mean, with its documented fallback to the unweighted mean when no
  member has a usable weight) and `.fg_series_vars()` (what a series is —
  `.fg_growth_prep()` now uses it too instead of its own copy).

`R/utils.R` — `lag_weight` appended to `utils::globalVariables()`.

`NEWS.md` — entry describing the defect and stating that no published values
change.

`tests/testthat/test_gapfilling.R` — one shared `weighted_proxy_fixture()`
tribble plus four tests: the crash regression (and that `"gdp[pop]"` matches
the unweighted `"gdp"` when each group has a single member), the grouped
weighted mean, the ungrouped `"var:[weight]"` branch, and a proxy series with
an interior gap.

**Classification: mechanical.** Nothing here is a methodological choice — the
weighted mean formula, the `w > 0` validity rule and the unweighted fallback
are all unchanged; only the column rename and which row the weight is read
from moved.

**No published values change.** Grepped every `proxy_col =` in `R/` — the
seven in `build_cbs.R` and three in `build_production.R` are all plain names
(`"pop"`, `"agriland"`, `"Cropland"`, `"yield"`, `"processing"`,
`"Palmkernel_Oil"`). None is weighted, which follows from the bug: any call
that was weighted raised an error.

## What I verified

- **Failing first.** With only the tests added, on unmodified code:
  `[ FAIL 6 | PASS 186 ]`, all six the `V1`/`V2` abort. With only the
  `setnames` line fixed: `[ FAIL 6 | PASS 190 ]`, now failing on the numbers
  in the table above — each one matching the value predicted by hand from the
  bug. With both fixes: `[ FAIL 0 | PASS 196 ]`.
- **Full suite**: `devtools::test()` → `[ FAIL 0 | WARN 11 | SKIP 8 | PASS
  5953 ]`. The 8 skips are the pre-existing `WHEP_*`/local-raster guards.
- **`R CMD check`** on a clean `git archive` export
  (`--no-tests --ignore-vignettes --no-manual`,
  `_R_CHECK_FORCE_SUGGESTS_=false` because `archive` and `RSQLite` are not
  installed here): `0 errors | 0 warnings | 1 note`. The note is
  `.canonicalise_gdp_pop_area: no visible binding for 'key_row' /
  'canonical_area'` — **byte-identical on an `origin/main` baseline export**,
  so pre-existing and untouched by this PR.
- **lint** on the three changed `.R` files with the project's disabled
  linters: no lints. All new lines ≤ 80 columns.
- **`air format`** run on the three changed files (`air.toml` keeps the
  `tribble` alignment). `devtools::document()` produced no diff — no roxygen
  was touched. Every `man/*.Rd` topic still appears in `_pkgdown.yml`
  (0 missing).

## Left undone (out of scope, filed separately)

The **value** lag in `.fg_growth_calc_individual()` is still grouped by the
coarse aggregation group, not by the series, and is protected only by the
`year == lag_yr + 1` adjacency guard. When two series in one group have
disjoint year coverage that happens to be adjacent, that guard passes and the
growth is computed across the series boundary. On `origin/main` and on this
branch alike, a region holding ESP (2000–2002) and FRA (2003–2004) gives FRA
2003 a growth of `7.264 = (1000 − 121) / 121` — FRA's `gdp` over **ESP's**.
That affects the unweighted path too, which #324 explicitly scopes out, so it
is not touched here; filed as #608.

Closes #324

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,21 @@
 # whep (development version)
 
+* **`fill_proxy_growth()` weighted proxy specs work at all, and weight the
+  right period.** The documented `"var[weight]"` and `"var:group[weight]"`
+  forms both aborted with
+  `Items of 'old' not found in column names: [V1, V2]`: the weighted
+  aggregation renamed its output columns from `V1`/`V2`, but data.table names
+  them after the symbols in the returned `list()`. With the crash out of the
+  way, the weight lag proved wrong twice over — it was taken *after* the rows
+  without an individual growth rate had been dropped, and it was grouped by the
+  coarse aggregation group rather than by the individual series. The first
+  member of every group therefore lost its weight entirely and fell out of the
+  weighted mean, the next member picked up the previous member's weight, and a
+  member with an interior gap was weighted by the wrong year. The
+  previous-period weight is now carried alongside the value lag, over the full
+  series and within it. **No published values change**: no pipeline in the
+  package passes a weighted proxy spec, because until now any call that did
+  raised an error.
 * **An aggregation bucket now sums, and comes out under one name.** The reader
   aggregation grouped rows by the member's polity **name** as well as by
   `polity_area_code`, so a bucket folding members that resolve to different
--- a/R/gapfilling.R
+++ b/R/gapfilling.R
@@ -1252,7 +1252,7 @@ fill_proxy_growth <- function(
   }
 
   # 2. Compute Individual Row Growth (with Smoothing)
-  prep <- .fg_growth_calc_individual(prep, spec, smooth_window, time_col)
+  prep <- .fg_growth_calc_individual(prep, spec, smooth_window, time_col, .by)
 
   # 3. Aggregate to Groups
   summary_dt <- .fg_growth_aggregate(
@@ -1286,8 +1286,7 @@ fill_proxy_growth <- function(
     return(data.frame())
   }
 
-  lag_vars <- unique(c(.by, spec$present_group_vars))
-  lag_vars <- lag_vars[lag_vars %in% names(data)]
+  lag_vars <- .fg_series_vars(spec, .by, names(data))
 
   cols <- unique(c(time_col, lag_vars, spec$source_var, spec$weight_col))
   cols <- cols[!is.na(cols)]
@@ -1302,7 +1301,7 @@ fill_proxy_growth <- function(
   as.data.frame(dt)
 }
 
-.fg_growth_calc_individual <- function(data, spec, window, time_col) {
+.fg_growth_calc_individual <- function(data, spec, window, time_col, .by) {
   # Helper to manage lag vars
   by_vars <- if (length(spec$present_group_vars) > 0) {
     spec$present_group_vars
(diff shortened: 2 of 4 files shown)


lbm364dl:
Closing as **superseded by #617**, with #606's tests preserved in #627 (both merged).

#324 and #171 were filed separately but describe the same defect: the weighted
branch of the proxy-growth aggregation, which aborted on every call with
`Items of 'old' not found in column names: [V1, V2]`. Both were fixed
independently, and the two fixes converged — the resulting
`.fg_weighted_growth()` is character-identical apart from one local variable
name (`usable` here, `valid` there) — and both return the same numbers on a
shared fixture:

```
"gdp:region[pop]"  ->  50.000 50.500 51.005 / 80.000 80.800 81.608   (both PRs)
                       ERROR: Items of 'old' not found ...            (main before)
```

#617 landed first, so this PR's code contributes nothing, and its branch predates
today's merges — merging it now would reintroduce superseded versions of
unrelated files.

What was **not** redundant here were the four test cases, which cover the
ungrouped `"var:[weight]"` form that #617's tests never exercise. They were
ported onto #617's implementation unmodified in #627, where they pass — which
also serves as independent corroboration that the two fixes agree beyond the
cases #617 asserts about itself.

No work from this PR is lost. Thanks for it — the numbers in its description were
what made the convergence checkable.

Issue #559, opened 2026-08-05
No validation covers the LPJmL-derived pins, so a 31% carbon-input change is invisible to the harness


lbm364dl:
## The gap

The four LPJmL-output-derived pins were just repointed from 5.9.7 to 6.1.1
(#558), which raises natural-land carbon input **~31%**
(`lpjml-grass-natural-net-c` mean ratio 1.309). Nothing in `validation/` can see
that.

`validation/validate_all.R` sweeps eight variables — production, production_psd,
occupation, land_per_tonne, cropping_intensity, cropping_intensity_obs,
cycle_length, stability — and every one reads FAOSTAT / GAEZ / MapSPAM / USDA PSD.
`validation/variables.R:15` says so explicitly:

```r
# public production pin - no LPJmL inputs).
```

So the harness ran clean through a change that moves every downstream SOC number.
I recommended it as the regression net for this migration; it is not one.

## What is unvalidated

The pins feed three chains, none covered:

| pin | consumer | covered? |
|---|---|---|
| `lpjml-grass-natural-net-c` | `build_grass_natural_carbon_inputs()` → `build_soil_carbon_inputs()` → `build_carbon_balance()` | no |
| `lpjml-soc-hydrology` | `get_soc_climate_drivers()`, `build_water_balance()` | no |
| `lpjml-grass-availability`, `lpjml-grass-productivity` | `R/feed_lpjml.R` | no |

## What was verified by hand instead

Ad hoc, at the pin boundary rather than as a repeatable check:

- Each published pin read back off the board with exact expected row counts
  (86,781,420 for hydrology — row-for-row identical to its predecessor).
- `build_grass_natural_carbon_inputs()` returns 101,186 rows on the new pin,
  `c_input_mgc_ha_yr` mean 6.318 (natural 6.738, grassland 5.735).
- Feed availability returns 41,860 rows.
- `get_soc_climate_drivers()` could **not** be exercised: it aborts on an unset
  `WHEP_CRU_DIR` before doing anything with the pin.

## The propagation is still unmeasured

The input-side change is quantified (1.309×). How it lands in SOC output is not,
because `build_carbon_balance()` needs local raster paths that are unset in a
fresh checkout — `WHEP_HWSD_DIR`, `WHEP_CRU_DIR`, `WHEP_LUH2_DIR`,
`WHEP_POLITY_FRACTION_PATH`, `WHEP_TYPE_CROPLAND_PATH`,
[...]

Issue #598, opened 2026-08-06
WHEP_LPJML_RUN_DIR and lpjml_hydrology.R both still name the 1901-2009 run, while the complete 1901-2023 run sits beside it


eduaguilera:
`read_lpjml_hydrology()` reads whatever `WHEP_LPJML_RUN_DIR` points at, and both that variable and the reader's own documentation still name the LPJmL 5.x **1901–2009** run — while the 6.x **1901–2023** run exists alongside it and is complete.

## Measured

```
WHEP_LPJML_RUN_DIR -> .../LPJmL_runs/global_1901-2009_spinup_200_our_inputs_grassland_livestock_npp_vegc_fix
R/lpjml_hydrology.R:5 -> "1901, lastyear 2009 (monthly time len 1308 = 109*12)"

available on disk:
  global_1901-2009_spinup_200_our_inputs_grassland_livestock_npp_vegc_fix
  global_1901-2023_spinup_200_our_inputs
  global_1901-2018_spinup_200_our_inputs
```

**The 2023 run carries every variable `.hydro_var_map()` reads** — `mseepage`, `mtransp`, `mevap`, `minterc`, `mprec`, `mrain`, `mirrig`, `mrunoff`, `mdischarge`, `mswc`, plus `cft_nir` and `cft_consump_water_b`/`_g`. Nothing is missing.

**And the sizes confirm it really is the longer series**, rather than a re-run of the same span: the monthly files go 995.1 MB → 1123.0 MB, a ratio of **1.1285**, against a year-count ratio of 123/109 = **1.1284**. Same grid, 14 more years.

## Why it matters

There is no warning anywhere. A reader pointed at the 2009 run returns 2009-truncated results and the roxygen agrees with it, so the wiring and the documentation corroborate each other on the stale answer. Anyone measuring the water balance today gets 5.x/2009 output and has no signal that a 6.x/2023 run is sitting next to it.

This is adjacent to but distinct from the two existing issues: **#558** (merged) re-pointed the four LPJmL-*derived pins* at the 6.1.1 run, and **#559** reports that no validation covers those pins. Neither touches the run directory the hydrology reader resolves at call time, nor the docstring.

## Suggested fix

1. Point `WHEP_LPJML_RUN_DIR` at the 6.x 1901–2023 run, and correct `R/lpjml_hydrology.R:5`'s "lastyear 2009" in the same change so the two cannot drift apart again.
[...]

Issue #596, opened 2026-08-06
test_carbon_balance.R errors instead of skipping when the local HWSD extract lacks t_clay


eduaguilera:
`tests/testthat/test_carbon_balance.R:833` **errors** rather than skipping when `WHEP_HWSD_DIR` points at an HWSD extract that exists but lacks the column the code needs.

## What happens

```
Error in `dplyr::filter(.read_hwsd_attributes_local(hwsd_dir), !is.na(.data$t_clay))`:
Caused by error in `.data$t_clay`: ! Column `t_clay` not found in `.data`.
```

On this machine `C:/XL_files/HWSD/hwsd_data.csv` carries `mu_global, t_usda_tex, share, t_ph_h2o` — and no `t_clay`, which is what `.cb_hwsd_clay()` reads.

## Why it is a defect rather than a local data problem

The test's `skip_if` guard checks only that `hwsd_data.csv` **exists**. It does not check that the file carries the column the test is about to require. So an incomplete extract — which is a perfectly ordinary state for a locally-derived file — produces a hard error in the suite instead of a skip, and the error names a `dplyr` internal rather than the actual cause.

That is the inverse of the more familiar failure mode. A guard that is too weak to *fail* lets a defect through; this one is too weak to *skip*, so it converts a missing-input condition into what looks like a code fault. Anyone whose HWSD export predates the `t_clay` column gets a red suite and a misleading message.

It reproduces identically on the parent commit, so it is not introduced by any recent work — it was found while running a full-suite control for an unrelated branch, in an environment where `WHEP_HWSD_DIR` was actually visible. Most runs never see it because the variable is unset and the block skips for the right reason by accident.

## Two candidate fixes, and the choice is a judgement call

1. **Re-export the CSV with `t_clay`** via `inst/scripts/export_hwsd_attributes.R`, so the local extract matches what the reader expects. Fixes this machine and anyone who re-runs the export, but leaves the guard just as weak for the next missing column.
2. **Widen the `skip_if` to test for the required column**, not merely for the file. Makes the test honest about its own precondition and fails loudly only when the data is genuinely wrong rather than merely partial.

Doing both seems right — (2) is the one that stops this recurring.

## Related

Same family as the guard problems being cleaned up elsewhere: a precondition check that reads as though it protects the test and does not. Worth pairing with #559's point that missing validation lets large changes pass unnoticed — here the opposite happens, and a missing *column* masquerades as a broken function.

PR #618, opened 2026-08-06
fix(soc): stop the unspatialized-carbon warning aborting multi-year runs

lbm364dl:
`build_carbon_balance(years = 1901:2023)` died after ~10 minutes of work with

```
Error in post_process_plurals(pstr, values): Multiple quantities for pluralization
```

from inside `.sci_warn_unspatialized()`. **The abort came out of a warning** — a
message that merely failed to format destroyed a three-hour computation.

## Cause

The message left `cli` to infer which value each plural marker refers to, and it
cannot. The first bullet interpolates **two** numbers (`nrow(lost)` and the rounded
mass), the second a vector — and `item_prod_code` is **integer**, so `cli` treats
the crop vector as another quantity candidate too.

Depending on the values it then throws one of two errors, which is why this was
awkward to pin down:

| values | error |
|---|---|
| the real 123-year run | `Multiple quantities for pluralization` |
| integer crops, length ≥ 2 | `length(object) == 1 is not TRUE` |

Two faces of the same ambiguity.

## Fix

Every marker gets its quantity pinned with `cli::qty()`, and the numbers are
precomputed into scalars instead of interpolated inline, so nothing depends on
`cli` guessing. `polity_folds.R` already pins a marker this way, so the pattern was
established in the codebase.

Pluralisation still reads correctly:

```
7488 polity-crop carbon components (4.3e+08 Mg C) ... were dropped.
1 polity-crop carbon component (4.3e+08 Mg C) ... was dropped.
```

## Why no test caught it

**Only multi-year spans reach it**: `1901-1905` warns fine, `1901-1910` aborts. A
single-year run cannot trigger it, and single-year is what everything tested.

The regression test uses integer `item_prod_code` with more than one unmatched crop
— the minimal trigger — and **fails without the fix**:

```
Error in `make_quantity(values$qty)`: length(object) == 1 is not TRUE
```

Verified by stashing the `R/` change and re-running: 1 failure, then passing with
it restored. 66 tests in the file, `lint_package()` clean.

## Note on the wider defect

The narrow bug is one message. The shape of it is worth noticing: a `cli` message
with more than one interpolated quantity is a latent abort in any code path that
only shows up with the right data, and it lands on the caller rather than on the
warning. There are other messages in `R/` with two interpolations and a plural
marker; I have not audited them, and a sweep might be worth its own issue.

Found while measuring what the LPJmL 6.x pin swap does to soil carbon (#559),
which this was blocking.

https://claude.ai/code/session_01TvvcyfcfaSxW9pywZsdvAY

### DIFF
--- a/R/soil_carbon_inputs.R
+++ b/R/soil_carbon_inputs.R
@@ -317,12 +317,27 @@ build_soil_carbon_inputs <- function(
     return(invisible(NULL))
   }
   crops <- sort(unique(lost$item_prod_code))
+  # Every plural marker gets its quantity pinned with cli::qty(), and the numbers
+  # are precomputed into scalars instead of being interpolated inline.
+  #
+  # Left to infer, cli has to decide which value each marker refers to, and with
+  # two numbers in the first bullet and a vector in the second it can conclude
+  # the answer is ambiguous and throw "Multiple quantities for pluralization" --
+  # from inside the warning, so the abort lands on the caller. That killed
+  # build_carbon_balance(1901:2023) outright after ~10 minutes of work, and only
+  # on multi-year spans: 1901-1905 warns fine, 1901-1910 dies. No single-year
+  # test would have caught it. polity_folds.R already pins a marker this way.
+  n_lost <- nrow(lost)
+  n_crops <- length(crops)
+  lost_mass <- round(sum(lost$c_mass_mg), 3)
   cli::cli_warn(c(
-    "!" = "{nrow(lost)} polity-crop carbon component{?s}
-           ({round(sum(lost$c_mass_mg), 3)} Mg C) had no crop-pattern cells and
-           {?was/were} dropped from the gridded soil carbon input.",
-    i = "Unspatialized item_prod_code{?s}: {.val {crops}}. Add {?its/their}
-         cells to {.field crop_patterns} to retain the carbon."
+    "!" = "{cli::qty(n_lost)}{n_lost} polity-crop carbon component{?s}
+           ({lost_mass} Mg C) had no crop-pattern cells and
+           {cli::qty(n_lost)}{?was/were} dropped from the gridded soil carbon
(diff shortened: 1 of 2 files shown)


eduaguilera:
Heads-up on a file collision, not an objection — the fix itself is clearly right.

`R/soil_carbon_inputs.R` is being rewritten on `edu/polycell-spatial-support` right now, as part of migrating the carbon path off the centroid grid and onto the polycell (the crosswalk half of EA4, plus rescaling the LUH2 class areas onto polycell land). That touches `.sci_*` and the surrounding join paths, so our diff and yours will overlap in that file.

Ours is not merged yet and yours is small, so the sensible order is **yours first** — we will absorb it. Flagging only so the conflict is expected rather than discovered.

One thing from your diagnosis worth keeping in view while we are both in there: *"the abort came out of a warning — a message that merely failed to format destroyed a three-hour computation."* That is the same shape as several defects this branch has been clearing, where the guard is the thing that fails rather than the thing it guards. Two others in the same family, already filed, in case they are worth batching with this:

- **#596** — `test_carbon_balance.R` errors instead of skipping when the local HWSD extract lacks `t_clay`. The `skip_if` checks that the file *exists* but not that it carries the column the test then requires, so a partial extract reports as a code fault.
- **#598** — `WHEP_LPJML_RUN_DIR` and `R/lpjml_hydrology.R:5` both still name the 1901–2009 run while the complete 1901–2023 run sits beside it, with no warning either way. The suggested fix there is the same principle as yours: have the reader **report the span it actually found** rather than documenting a constant.

Also relevant to your multi-year run specifically: `.cb_drop_uncovered_climate()` drops polycells behind a `cli_warn` that reads like an ordinary climate-coverage gap. On a half-migrated crosswalk that silently loses about 14% of them, which is why we are moving the carbon path as a single commit rather than file by file.

Issue #529, opened 2026-08-04
cshapes straight borders render as great circles: 123,276 km2 of Canada is booked to the USA, invisible to every conservation check


eduaguilera:
**The US–Canada border in `whep::polities` is displaced north by up to ~0.85 degrees (~94 km), moving roughly 123,276 km² of Canadian prairie into the United States.** No conservation check can detect it. Found while building the polycell spatial support table on branch `edu/polycell-spatial-support`; every figure below was measured independently by a second agent against shipped package data.

## Mechanism

`polygon_source` for both polities is `cshapes-2.0`, a planar GIS product. It encodes the 49th-parallel border as a **single segment between two vertices 27.6 degrees of longitude apart**:

```
(-122.757794, 48.997353)  ->  (-95.154173, 48.999436)
```

In a planar viewer that renders as the straight line the treaty describes. Under spherical geometry — which `sf` uses by default via s2, and which any geodesic area calculation requires — a segment between two points is a **great circle**, and a great circle between two points at equal latitude **bulges poleward** between them. So the rendered border leaves the parallel and arcs north into Canada.

## Measurement

Fraction of each 0.5 degree cell at latitude 49.25 assigned to each polity — these are cells lying **entirely north of the treaty line**, so the correct answer is `CAN = 1.0000` in every row:

```
      lon     lat     USA     CAN     sum
  -119.75   49.25  0.6447  0.3553  1.0000
  -117.25   49.25  1.0000  0.0000  1.0000
  -114.75   49.25  1.0000  0.0000  1.0000
  -112.25   49.25  1.0000  0.0000  1.0000
  -109.75   49.25  1.0000  0.0000  1.0000
  -107.25   49.25  1.0000  0.0000  1.0000
  -104.75   49.25  1.0000  0.0000  1.0000
  -102.25   49.25  1.0000  0.0000  1.0000
```

Total displaced area on that single edge: **12,327,586.7 ha = 123,276 km²**.

The same mechanism appears wherever a straight border is stored with sparse vertices. Egypt/Sudan on the 22nd parallel: **235,891 ha** across two edges (143,079.4 + 92,811.7).

## Why nothing catches it

`USA + CAN = 1.0000` exactly in every row. There is **no overlap and no gap**, and total land is conserved. So every structural check a spatial product would normally rely on passes:

- area decomposition holds;
- re-aggregation to the polity is exact;
- nothing crosses a boundary in the sense of double-counting;
[...]

PR #619, opened 2026-08-07
Replace the two cell-polity crosswalks with the polycell, and move the area denominators onto it

eduaguilera:
Replaces the two incompatible cell↔polity crosswalks with one unit, and moves every area denominator on the extent side of #458 onto it.

**Draft** — the identity half is not here, and two known gaps are called out at the bottom. Please read those before the diff.

## The unit

The **polycell** is the intersection of a 0.5° cell with a polity over that polity's validity interval, `polycell_id = f(cell_id, polity_code)`. It carries territory decomposed into three separately addressable categories:

```
polity_area_ha = land_area_ha + inland_water_ha + ice_area_ha
```

Per-hectare biogeochemical rates apply to `land_area_ha`; `polity_area_ha` is what reconciles with official national areas. Inland water and ice are territory that receives real fluxes, not rounding error — deposition to freshwater is a live eutrophication term.

## What was wrong

- **A per-hectare rate multiplied by whole-cell area over-counts by 11.0%** — global terrestrial area is 12.9931 Gha against 14.3195 Gha of whole cells. `polity_frac` *is* a valid partition of the land; it was simply multiplied by the wrong base.
- **Two crosswalks were live in one package.** The carbon path read a centroid grid where `cell_area_frac` defaults to 1, so a whole border cell went to one polity; the nitrogen path split fractionally. No test detected it, because each path was internally self-consistent.
- Border splits were quantised to **1/36** of a cell (`subcells = 6L`), not the 1/144 often assumed.
- Two live `end_year` fenceposts read the boundary inclusively, so on a succession year both epochs activated and the first-sorting row took every cell — **and the dissolved predecessor sorts first in 315 of 315 cases**, so the successor got no row and cells outside its smaller extent were dropped. That lost mass, not just labels.

## Measured movement, per consumer

Each consumer moved on its own commit, against a characterisation baseline pinned first, so the movement is attributable rather than aggregate.

| consumer | movement |
|---|---|
| deposition, split key | **mass conserved exactly** — new − old = 0 t; rate `identical()` over 63,065 cells |
| deposition, categories | scope selectable, defaults to whole territory → **ledger bit-identical** |
| synthetic | per-polity totals moved by **exactly 0**; weighted share movement 1.99% |
| carbon | **+1.32%** on masses (stock, C input, SON) |
| spatialize | **zero** — enforcement only |
| whole ledger | −0.504% on N input, of which **95.6% is the identity gap, not geometry** |

**Carbon is the one to look at.** Per-hectare densities are *invariant under a pure re-split*, so a density check reports ~0 whether or not the migration is correct — and here the **global mean stock density falls while the mass rises**, so a density check would have reported the wrong *sign*. All carbon movement is therefore measured on `sum(density × area_ha)`.

## Two decisions worth disagreeing with, if you do

**The ledger keeps the whole territory.** Restricting deposition to the land share would remove 0.89 Tg N (−1.44%), concentrated on lake and glacier states. We measured it, then didn't take it: **N deposited on a lake or glacier still drives indirect N₂O and still reaches the eutrophication pathway**, so the terrestrial-only scope is wrong for those terms. Scope is an explicit argument defaulting to `"territory"`, recorded in an output column.

**The table keys on `polity_code`, not the bucket.** `polity_area_code` folds 544 codes into 201, with 206 holding Sudan and South Sudan — the fold #591 caught corrupting published series. Consumers convert at their own boundary, so the fold happens visibly at a join rather than irrecoverably in the grid. This is why #579's `grid_area_code` as an *added* column was the right shape.

## Evidence

The producer went through **five adversarial verification rounds**; four found a real defect, two of them blocking, and **every one was invisible to a green test suite**. Silent territory deletion (Greece losing 10% while reporting healthy status); an abort that made the production call impossible at any year; five live mutants; and dead gap code whose only guard could detect duplication but never omission.

So the acceptance check here is **mutation testing, not the suite**. Every commit carries a baseline-gated sweep in a sandbox that cannot write the repo, with sources checksummed either side. Surviving mutants are either killed or *proved* equivalent by measurement — one over 228,175 comparisons, another over 102,988.

Every commit also proves "no new failures" against a `git archive` of its own parent, in the same environment, **diffed at block level rather than by totals** — two offsetting changes would cancel in a total. One earlier control was invalidated because the two sides ran in different environments; the tell was a skip-count mismatch, and it is now caught by printing environment fingerprints on both sides.

`rcmdcheck` **run with tests** 0 errors / 0 warnings / 1 note; `devtools::test()` `FAIL 0 | PASS 8459` in the offline job's own configuration; `lint_package()` 0; pkgdown complete (302 topics, 302 `_pkgdown.yml` entries, 0 missing and 0 extra).

One disclosure about that evidence, because it is the reason this PR first went red on seven checks. Every local check on this branch ran `rcmdcheck(args = c("--no-tests", …))`, which **skips the suite entirely** — so a set of census assertions pinned against an older polity vintage stayed green locally and failed everywhere in CI. `whep-polities` #551 flipped `GRC-1830-1913` to `wiki_status = "superseded"`; the live filter then drops it, `build_polycell_support()` returns **0 rows with no error and no warning**, and 13 assertions ran against an absent frame. They are re-pinned on the live successor `GRC-1881-1913` against an **engine-independent** reference (planar clip measured in Lambert azimuthal equal-area, 7.4e-05 relative), so the pin constrains neither the clipping nor the area engine. Two of them were passing *vacuously* on empty frames and now cannot: under a mutant that kills the fixture the same way #551 did, they still pass while the new guards fail.

## Known, and deliberately not fixed here

- **8 of those 9 polities still lose their deposition, and #459 did not fix it.** Re-measured after the fourth reconciliation with `main`, against the deployed `cell_polity_fraction.parquet` (191 area codes). The codes that still receive no polycell are **61 Equatorial Guinea, 153 New Caledonia, 154 North Macedonia, 209 Eswatini, 212 Syria, 276 Sudan, 277 South Sudan, 299 Palestine**. **Only Fiji is fixed** — the #662 snapshot refresh leaves `FJI-1800-2025` s2-readable (0 s2-invalid on the 751-row table) and the crosswalk maps area 66 onto it. #459 is closed and #628 is merged, but they promoted these territories on the **`reporting_polity_code`** axis only: `polity_area_crosswalk` still carries `polity_code = ROW-1850-2025` and `polity_area_code = 999` for all six, and carries **no row at all** for `SYR-1967-2025`, `PSE-1948-2025`, `SWZ-1894-2025` or `GNQ-1968-2025`. The support keys on `polity_code → polity_area_code`, and the producer drops `polity_type == "aggregate"`, so the promotion never reaches it. Sudan and South Sudan stay folded on 206 — #414 is now **closed**, and the open decision is costed in **#680** and `lbm364dl/whep-polities#139`.
- **What carbon drops is larger than this PR claimed, and two different defects were being summed as one.** At the carbon support year 2015, **32** clipped polities carry no `area_code` and are deleted by `.carbon_drop_unkeyed()`. The six named above are **28.2 Mha** of territory (Syria **18.70**, Equatorial Guinea 2.70, North Macedonia 2.54, New Caledonia 1.88, Eswatini 1.71, Palestine 0.62) — the old 26.6 Mha over seven, minus Fiji, on a different basis. The other 26 are **252.5 Mha**, almost all of it **Greenland 213.2**, **Western Sahara 26.8** and **French Guiana 8.4**. None of those 26 resolves to a reporting bucket at all, so they are territory outside the reporting vocabulary rather than a crosswalk gap, and no deposition row ever existed for them to lose. **Basis warning**: the GLWD water layer and the Natural Earth ice layer are unreadable in this environment (`WHEP_LPJML_INPUT_DIR` and `WHEP_NATURALEARTH_DIR` unset), so every figure in this bullet is **territory, an upper bound on land**. That is a tight bound for the six (Syria 18.70 territory against 17.91 land measured previously) and a poor one for Greenland, most of which is ice. Measured on a polity-subset build, proved subset-invariant by a control that rebuilds the same targets with five neighbouring polities present (`max |land diff| = 0 ha`). The **"single largest lever left in the ledger" claim is withdrawn** rather than restated: it was never measured against the ledger's other terms, and these two populations should not be added into one lever.
- **Water and urban are not migrated.** Both are gated on data layers on your side — water on what the LPJmL depths are referenced to (see #598: the wiring and the docstring both still name the 2009 run while the complete 2023 run sits beside it), urban on HYDE. The decisions are recorded: regenerate `landfrac` and rerun for water, key at HYDE's native 5 arcmin for urban.
- **`feed_lpjml.R` is frozen.** Its `cell_area_ha` multiplication *inverts a normalisation applied upstream*, so substituting land area there would under-count global grass — the mirror image of the defect this PR fixes.
- **Six `no visible binding` lines are deliberate** — three in `water_balance.R`, one in `feed_intake_redistribute.R`, one in the frozen `feed_lpjml.R`, and, as of the fourth reconciliation with `main`, one in main's own new `.read_fraction_country_grid()`. Dropping the `polity_frac` global was the mechanism that proved no consumer was missed: with it present a planted missed consumer is invisible, with it removed the planted site is named — which is exactly how main's new fractional-grid reader announced itself here. The note count is unchanged at 1.
- The great-circle border displacement (#529, ~123,000 km² of Canadian prairie booked to the USA) is **detected and reported, not corrected** — the fix belongs in the polygons.

## Also in here

Fixes for two live epoch fenceposts (1,912 polity-years move, 1,103 off a dead label); a status-aware tie-break so a retired interval can no longer outrank a live one (Montenegro 1913–14 was resolving to a 37.6% smaller polygon); the open-interval rule so 2025 resolves at all; and `inst/scripts/reconcile_polity_areas.R`, which reconciles every polity against FAOSTAT land + inland waters and attributes all 45 material discrepancies to polygon quality, identity, or convention.

Small polities are the headline there: Kiribati drew **34.3×** its official area under the old crosswalk and now draws 1.18×.

@lbm364dl — review please. The `data$cell_polity` contract is measured to the byte if you want to sequence the `build_cell_polity()` re-key against it.


eduaguilera:
Final post-main acceptance (head 202e5d65, 2026-08-08):

- Merged origin/main at f8e55e57 and re-pinned PR #662's polity snapshot: 751 snapshot rows, 692 live polities, 414,479 measured/retained pieces, 21 terra pieces across 9 polities, and zero DA-22 long edges. The exact full clip took 3,843.01 s.
- The census verifier now gates the checked blob, requested/output code-set identity, coverage/status completeness, engine census, S-A2 exceptions, and the all-runtime terra/s2 residual envelope. Its advertised GRC subset reaches Done with 63 pieces, 4 terra pieces, and 227,311.6 ha.
- The macOS sf/GEOS failure is closed: missing or unusable intersection idx metadata now falls back to exact per-source intersections. The focused polycell file is green locally with 424 expectations.
- Final CI is green: format 31221505737, lint 31221506959, offline tests 31221507000, coverage 31221506978, pkgdown 31221506957 (attempt 2), and all five R-CMD-check platforms including macOS 31221506958.

eduaguilera:
@lbm364dl — one figure in here I could not settle myself, and it is on your side of the polygon question rather than the extent side. Flagging it for review rather than changing it, because I do not think it is wrong so much as **two different quantities being compared**.

`NEWS.md:22` gives the union of the live polity polygons as **13.2795 Gha**. The whole-domain polycell build at 2015 sums to **13.4599 Gha** (= 12.9253 land + 0.3032 inland water + 0.2314 ice, S-A1 exact).

Those should not be expected to agree: a **union** counts overlapping ground once, whereas a **polycell sum** counts it once per claiming polity. This build reports **34,526 overlap rows**, which is more than enough to account for the 0.18 Gha gap in the right direction.

So the question is not "which number is right" but **which one `NEWS.md:22` should be quoting**, and whether the union figure is still current on the 753-row snapshot — it predates the refresh in #734, and #529's densification changed polygon geometry along every straight border.

Two things that would settle it, both cheap on your side:

- Recompute `sf::st_union()` over the live polygons on the current `polities.rda` (blob `4f1fa941`) and see whether 13.2795 still holds.
- Decide whether the NEWS sentence wants the union (unique ground) or the polycell sum (attributed territory). They answer different questions, and the surrounding paragraph is about attribution, which argues for the sum — but that is your call, not mine.

Everything else in the body has been re-measured against this HEAD, with a `### Measurement basis` section naming the three classes I could **not** reproduce here and why (HaNi-gated, LUH2-gated, and the deployed parquet). This one is listed there too.

Separately filed rather than left here, because it outlives the PR: #758, on the repeated-key guard not catching *overlapping* intervals of one polity in one cell. Latent — the shipped table cannot trigger it — but it is a contract question about what a caller-supplied `geometries` table may contain.

lbm364dl:
Ran the recompute you asked for. Both of your bullets, plus the conservation check underneath them. Pushed as 877cfad9.

**The support table conserves area exactly.** Sum of the live polity polygons at 2015 against the sum of `polity_area_ha` off the pinned support (`20260810T165201Z-7caf6`):

| | Gha |
|---|---|
| Σ polygon areas @2015 | 13.459855 |
| Σ polycell `polity_area_ha` @2015 | 13.459855 |
| **difference** | **0.000000 (0.0000%)** |

223 live polities on both sides, zero invalid polygons at 2015. The intersection neither loses nor invents ground, which is the property the whole table exists to provide.

**On the union: it is stale, and overlap is not what moved it.** Recomputed with `sf::st_union()` over the live polities on the current `polities.rda` (blob `4f1fa941`, 753 rows → 694 live → 223 at 2015):

- union @2015 = **13.4267 Gha**, against the 13.2795 in `NEWS.md:22`
- overlap, polygon side (Σ − union) = **0.0332 Gha**
- overlap, pin side, measured independently (territory claimed beyond the cell) = 0.0308 Gha over 30,855 cells

So overlap is real but roughly a fifth of what the gap needs. Of the ~0.180, about **0.147 is vintage drift** (#734's refresh + #529's densification, your second bullet) and only ~0.033 is double-claimed ground. The 34,526 figure is an all-epoch row count, and a row count doesn't bound an area — 34.5k slivers are still slivers.

**On which quantity `NEWS.md:22` should quote — I went with the union**, against your lean, because rows 1, 2 and 4 of that ladder (whole cells, HaNi's mask, LUH2 terrestrial) are all unique-ground masks. A count-once-per-claimant sum in row 3 makes the ladder incomparable with itself, which is the opposite of what the section is for. The sum is now stated next to it as the attribution quantity, so both are on the page.

While doing that I think I found what invited the confusion: the roxygen said `polity_area_ha` **is** the third row. It carries the third row's territory, but summing it isn't the union. That sentence is rewritten in both `NEWS.md` and `R/polycell_support.R`.

**One thing I did not touch.** `NEWS.md:14` says the whole-cell convention over-counts by **11.0%**, citing 14.3195 against 12.9931 Gha — that pair gives **10.2%**. I left it alone rather than guess which of the three numbers is off, since `WHEP_LUH2_DIR` is unset here and you listed that row as gated. Your call.

Green locally on the merge commit: `R CMD check` `Status: OK` (0/0/0), lintr clean, `air` clean, pkgdown reference index complete, and 1106 passing / 0 failing across the polycell and polity suites.

Issue #624, opened 2026-08-07
build_carbon_balance needs ~89 GB for 1901-2023: .sci_to_grid fans out every year at once


lbm364dl:
## Symptom

`build_carbon_balance(years = 1901:2023)` — the full historical march, i.e. the
normal production configuration — reaches **89.3 GB** and gets OOM-killed:

```
Out of memory: Killed process (R) total-vm:105036576kB, anon-rss:89312576kB
```

That is not a machine that most users have. A 63-year span (1961-2023) fits, but
only by giving up half the period.

## Cause

`.sci_to_grid()` (`R/soil_carbon_inputs.R:252`) spatialises polity-crop carbon to
cells with one many-to-many join:

```r
components |>
  dplyr::inner_join(
    weights,
    by = c("area_code", "item_prod_code"),
    relationship = "many-to-many"
  )
```

`components` carries a **year** dimension. `weights` does **not** — it comes from
`crop_patterns`, which is time-invariant (my note on the gridded surfaces records
it as static, and the schema has no year). So the join replicates the same spatial
weights across every requested year **simultaneously**:

| span | fan-out rows | payload at 8 numeric cols |
|---|---|---|
| 1 year | 6,741,717 | ~0.4 GB |
| 63 years | 424,728,171 | ~25 GB |
| **123 years** | **829,231,191** | **~49 GB** |

= 2,247,239 static cell-crop weights x 3 `input_type`s (`crop_residue`, `root`,
`manure`) x years. The 49 GB of payload becomes the observed ~89 GB once dplyr
holds its copies through the join, the `.sci_rescale_cell_area()` step and the
`mutate()`.

Peak memory is therefore **O(years)** in a quantity that has no year dependence.

## Fix

Chunk the fan-out by year. The weights are identical for every year, so joining
one year at a time and binding the results gives **bit-identical output** with
peak memory O(1 year) instead of O(all years) — roughly a 123x reduction in the
dominant intermediate, taking the full span from ~89 GB to a few GB.

This is the same shape as the fix already applied to the SOC hydrology pin
generator in `whep_inputs`, where reading all years at once used 113 GB and
per-year chunking with a spill to disk holds it flat at ~2 GB.

Two implementation notes:

- Chunk **inside** `.sci_to_grid()` (or in `build_soil_carbon_inputs()` around
  it), not in callers, so every caller benefits and none has to know.
- `.sci_warn_unspatialized()` is called once before the join and reports across
[...]

PR #629, opened 2026-08-07
perf(soc): spatialize and aggregate one year at a time

lbm364dl:
Addresses #624. **Makes long spans possible, not faster** — stating that up front
because the `perf` prefix invites the wrong expectation.

## The problem

`build_carbon_balance(years = 1901:2023)` — the production configuration — reached
**89.3 GB** and was OOM-killed. 1961–2023 was killed too.

A dimension mismatch in the spatialization: `components` carry a `year`, the
weights from `crop_patterns` do not (that raster is time-invariant). So the
many-to-many join replicated the same **2.25 M cell-crop weights across every
requested year simultaneously**:

| span | fan-out rows | payload |
|---|---|---|
| 1 year | 6.7 M | ~0.4 GB |
| 123 years | **829 M** | **~49 GB** |

## The change

Chunked per year, with four details that matter:

- **The chunk wraps the join *and* `.sci_finalise()`.** Chunking the join alone
  would not help — `.sci_finalise()` collapses only `input_type` and keeps the crop
  dimension, so its output is still cell × crop × year and the caller receives an
  O(years) table regardless. The crop dimension only collapses later, in
  `.ci_cropland_class()`.
- Weights built **once**, not per year.
- `.sci_warn_unspatialized()` stays **outside** the loop: it reports totals, so
  warning per year would spam the caller *and* change the reported numbers.
- Components are `split()` once, not filtered per year. Filtering inside the loop
  rescans the whole table every iteration — 829 M row-scans where one pass does
  6.7 M — so the memory fix would have cost time.

## Verified identical, not assumed

Same 3-year span through old and new in separate processes:

| | rows | columns | max abs diff (11 numeric cols) |
|---|---|---|---|
| old vs new | 4,861,893 both | identical | **0** |

66 unit tests pass, `lint_package()` clean.

## What this does not do

**Not a speed-up.** 3 years: 7.7 min chunked vs 7.4 min unchunked.

**Does not touch the fixed floor.** At 3 years the fan-out payload is ~1.2 GB while
peak RSS is 29.4 GB — the rest is `build_primary_production()` +
`build_commodity_balances()`. So the profile is roughly:

```
peak ~ 29 GB floor (input chain)  +  ~0.4 GB per year of fan-out
```

This removes the second term only. **Whether that is enough for 1901–2023 is
untested**, because no long run has completed yet. If the floor turns out to grow
with span, this is necessary but not sufficient and the floor needs its own issue.

Separately, profiling suggests the *runtime* cost of a long span is the
`build_carbon_balance` march (~1.5 min/year), which this does not address either.
That is being measured now.

https://claude.ai/code/session_01TvvcyfcfaSxW9pywZsdvAY

### DIFF
--- a/R/soil_carbon_inputs.R
+++ b/R/soil_carbon_inputs.R
@@ -70,20 +70,47 @@ build_soil_carbon_inputs <- function(
   }
   d <- .sci_resolve_inputs(data, years)
   components <- .sci_assemble_components(d$npp, d$manure)
-  gridded <- .sci_to_grid(
-    components,
-    d$country_grid,
-    d$crop_patterns,
-    d$harvested_area
-  )
-  .sci_finalise(gridded, resolution, d$residue_humification) |>
+  .sci_grid_and_finalise(components, d, resolution) |>
     .add_reporting_polity_columns()
 }
 
 # Private helpers ----
 
+# Spatialize and aggregate ONE YEAR AT A TIME, then bind.
+#
+# The spatial weights are time-invariant (crop_patterns has no year), so joining
+# every requested year at once inflates the intermediate by the number of years
+# for no benefit: 2.25e6 cell-crop weights x 3 input types x 123 years is 8.3e8
+# rows, ~49 GB of payload, and dplyr's copies through the join took a 1901-2023
+# run to 89 GB before the kernel killed it (#624). The crop dimension only
+# collapses later in .ci_cropland_class(), so chunking the join alone would not
+# help -- .sci_finalise() has to be inside the loop too, which is why this wraps
+# both.
+#
+# Output is unchanged: `year` is one of the .sci_finalise() grouping keys, so
+# grouping within a year gives exactly what grouping across years gave, and the
+# weights each year sees are identical.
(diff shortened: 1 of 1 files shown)


lbm364dl:
## Re-verified against current main

The earlier identity proof was against a main **43 files / 2,039 insertions** older,
so it no longer covered this base. Re-ran old vs new on current main
(`c14d61c9`), sequentially in separate processes:

| | rows | columns | max abs diff (11 numeric cols) |
|---|---|---|---|
| old vs new | **4,874,649 both** | identical | **0** |

Worth noting the row count moved from 4,861,893 to 4,874,649 purely from main's
own changes — so re-running was not ceremony; the previous number would have
looked like a discrepancy to anyone checking later.

Also confirmed after merging main that both this restructuring and #618's
`cli::qty()` pins survived — they live in the same function and were the obvious
thing to lose in a merge.

`lint_package()` clean, 66 tests in the file pass, 10/10 CI green.

Still true, and stated in the description: this removes the span-scaling fan-out
term only. The ~29 GB / ~8 min floor is the year-unaware build chain, which is
being handled separately in #570 / #367.

PR #508, opened 2026-08-04
Resolve livestock GLEAM/IPCC regions from the polity, not just ISO3

lbm364dl:
Part of the polity migration epic #458. Closes #465.

## What was wrong

`prepare_livestock_emissions()` attaches `iso3` from the area crosswalk, and the
Tier-1 emission-factor joins looked that `iso3` up in `gleam_geographic_hierarchy`:

- `.add_ipcc_region()` in `R/livestock_enteric.R`
- the N-excretion join and the manure-CH4 EF join in `R/livestock_manure.R`
- `.join_weights()` in `R/livestock_energy.R`

`gleam_geographic_hierarchy` lists 204 modern sovereign states. Any polity whose
ISO3 is not one of them resolved to `region = NA` and silently took the **Global
default** emission factor.

## Evidence it reproduced BEFORE the change

On this branch, at the parent commit:

- 59 of the 265 area codes `.current_area_lookup()` can resolve carry an ISO3
  with no row in `gleam_geographic_hierarchy`.
- Six of those areas actually carry livestock heads in the FAOSTAT production
  input (`primary_prod`), by head-years 1961-2023:

  | area | name | head-years |
  |---|---|---|
  | 228 | USSR | 10.2e9 |
  | 999 | RoW | 1.45e9 |
  | 248 | Yugoslav SFR | 6.7e8 |
  | 51 | Czechoslovakia | 4.1e8 |
  | 15 | Belgium-Luxembourg | 3.0e8 |
  | 186 | Serbia and Montenegro | 1.1e8 |

- All of them returned the **Global** dairy EF of 80 kg CH4/head/yr:

  ```
   15 Belgium-Luxembourg     iso3=BLX  ef=80
   51 Czechoslovakia         iso3=CSK  ef=80
  151 Netherlands Antilles   iso3=ANT  ef=80
  186 Serbia and Montenegro  iso3=SCG  ef=80
  228 USSR                   iso3=SUN  ef=80
  248 Yugoslav SFR           iso3=YUG  ef=80
  999 RoW (aggregate)        iso3=ROW  ef=80
  ```

Three of the five new tests fail at the parent commit (two of them error,
because the helper does not exist and because `.add_ipcc_region()` could not
accept a frame without `iso3`). The other two are pins that hold before and
after by design.

## What changed

`.gleam_region_of()` resolves the GLEAM region per row:

1. the **ISO3 leg** runs first, so present-day polities resolve to exactly the
   region they resolved to before;
[...]

Issue #492, opened 2026-08-04
Aggregate reporting buckets have no GLEAM grouping, so area 999's meat production drops silently


lbm364dl:
Found while implementing #207 (PR #475). Distinct from #415, which is about individual reporting areas: this is about the **aggregate buckets themselves**.

## Problem

The regional buckets — `ROW`, `RAFR`, `RASI`, `REUR`, `RNAM`, `ROCE` — have **no GLEAM row**, so the emission intensity for FAOSTAT area `999` and its siblings resolves to `NA` and their meat production is dropped by an inner join. Silently.

That is not a rounding error. Bucket `999` aggregates 62 FAOSTAT areas, 16 of which have real polities of their own (`BMU-1684-1968`, `FRO-1800-2025`, `GNQ-1886-1968`, `GUF-1816-1946`, Guadeloupe, Guam, Cook Islands...). Whatever they produce lands in `999`, and `999` has no grouping, so it leaves the extension entirely.

#382 excluded aggregates from the GLEAM classification deliberately, so this is a live decision rather than an oversight — but the consequence was not stated where anyone would see it.

## Decision needed

1. **Give the buckets a GLEAM grouping.** Cheapest, and defensible for `ROW` if a world-average intensity is acceptable. Wrong in spirit for `RAFR` vs `REUR`, whose members differ by more than the regional averages do.
2. **Derive the bucket's grouping from its members**, weighted by production. More defensible, more work, and it needs a rule for members that have no grouping either.
3. **Accept the drop and report it.** Currently the only honest reading of the code, but then the size of the drop should be published, not inferred.

Whichever is chosen, the drop should be **warned about rather than silent**. PR #475 adds a warning for individual live reporting areas GLEAM cannot classify; extending it to the buckets was left out of that PR because it is this decision, not plumbing.

## Related

- **#415** — five individual areas, and see the comment I am adding there: its list is now stale.
- **#419** — if the RoW fold stops, most of this evaporates, because the 16 members with their own polities would carry their own groupings. That makes this partly downstream of #419 rather than independent.

Refs #458.

Issue #465, opened 2026-08-04
Route livestock EF/region lookups through a polity-keyed region crosswalk instead of ISO3


lbm364dl:
**Problem**
`prepare_livestock_emissions()` attaches `iso3` via the non-year-aware `.current_area_lookup()`,
and three downstream functions (`.add_ipcc_region` in `livestock_enteric.R`, the Nex/MCF/MMS joins
in `livestock_manure.R`, `.join_weights` in `livestock_energy.R`) join that `iso3` against
`gleam_geographic_hierarchy`, which lists only modern sovereign states. Any polity whose historical
(or dependent-territory) ISO3 is absent silently falls back to the Global default emission factor.
MEASURED: 38 distinct historical/dependent polity_codes are affected, including 6 substantial
dissolved states with real livestock output — USSR (F228), Czechoslovakia (F51), Yugoslavia (F248),
Serbia & Montenegro (SCG), Belgium-Luxembourg (BLX), Netherlands Antilles (ANT).

**Evidence**
- `R/livestock_bridge.R:72-84`, `R/polities.R:11-35`, `R/livestock_enteric.R:69-143`,
  `R/livestock_manure.R:51-52,192-263,358-364`, `R/livestock_energy.R:275-296`.
- MEASURED 38-polity list as above.

**Target state**
Derive the GLEAM/IPCC region from `polity_code`/`polity_area_code` directly: add a
`polity_code`→gleam_region override table (or column) for the historical/dissolved polities
identified, and change the joins to use it instead of `iso3`.

**Acceptance criteria**
- The 6 named dissolved states (and the other 32 historical/dependent polities) resolve to a real
  GLEAM region, not the Global fallback, wherever they have production.
- Existing GLEAM-region tests pass unchanged for present-day polities.

**Depends on**: none. **Decision needed**: single-region assignment for the aggregate "Rest of
Africa/Asia/…" polities (RAFR/RASI/REUR/RLAM/RNAM/ROCE), which are aggregates by construction.

**Moves published values**: yes. **Port from #382**: no.

---

---
Part of the polity migration epic #458.

Issue #415, opened 2026-07-30
Five live reporting areas get no GLEAM/IEA grouping: Bermuda, Guam, Nauru, Palau, Tuvalu


lbm364dl:
`.energy_country_grouping()` derives all three GLEAM regional schemes (`development3`,
`region5`, `detailed15`) by transmuting `gleam_geographic_hierarchy`. That table therefore
defines the country universe for the energy extension, and an area absent from it gets **no
row at all** — not a wrong group — so a join on the grouping drops it silently.

### Measured

18 of the 217 self-reporting areas (`area_code == polity_area_code`, i.e. not folded into a
bucket) have no row in that table. They fall into three classes:

| class | areas | verdict |
|---|---|---|
| regional aggregates | `RAFR RASI REUR RLAM RNAM ROCE ROW` | fine — a country table should not carry aggregates |
| dissolved entities | Belgium-Luxembourg, Czechoslovakia, Netherlands Antilles, Serbia and Montenegro, USSR, Yugoslav SFR | fine — GLEAM is a present-day table |
| **live territories** | **Bermuda (17), Guam (88), Nauru (148), Palau (180), Tuvalu (227)** | **unclassifiable** |

Each of those five exists today, has its own polity, and reports under its own area code —
they are not folded into `ROW`. Pinned by identity in
`tests/testthat/test_gleam_hierarchy_coverage.R` so a sixth would fail rather than hide in a
count.

Tuvalu is the sharpest case: `.energy_ldc_iso3()` lists `TUV` among the least-developed
countries, so the code asserts a classification for a country the table it joins against
cannot represent. (All 46 LDC codes do resolve to polities, so that list is sound — the gap
is GLEAM's.)

### Why nothing is edited in the table

`gleam_geographic_hierarchy` is parsed from the GLEAM Excel source (`tab_sa1sa2` in
`data-raw/livestock_coefficients.R`). Adding five rows would make the package's copy diverge
from the published source with nothing recording it — the opposite of what the polities work
is for.

### Options

1. **Classify from the polities database when GLEAM has no row.** The schemes need
   `continent` (which `polity_area_crosswalk` carries) plus the `oecd`/`eu27` flags, and none
   of these five is OECD or EU27, so `continent` is sufficient. Consumer-side, no divergence
   from the source, and it removes the silent drop. Needs a decision: Oceania → "Non-OECD
[...]

PR #645, opened 2026-08-07
fix(energy): price the dissolved federations from the membership they held

### LINKED ISSUE #553: Energy CO2 extension loses 569 Mt of dissolved-entity meat production: it never got #508's polity-keyed GLEAM regions
Found while measuring #492 (PR #552). Distinct from both issues it sits between:

- **#415** classes the dissolved entities as "fine -- GLEAM is a present-day table". That is true
  about the *classification* and false about the *consequence*: their production leaves the energy
  extension entirely.
- **#465 / PR #508** solves exactly this problem, but only for the **livestock GHG** extension
  (`R/livestock_enteric.R`, `R/livestock_manure.R`, `R/livestock_energy.R`). PR #508 states
  explicitly that `R/energy_co2_extension.R` "keeps its own ISO3-keyed `iso2reg`" and that wiring
  it is not that PR's call.

So nobody owns the energy extension's copy of the problem.

## Measured

Driving `build_energy_co2_extension()` over the real `get_primary_production()` output
(1850-2023), the meat carcass production that gets **no energy intensity at all** -- no row in
`gleam_geographic_hierarchy` -- and therefore leaves the extension:

| area | iso3 | carcass Mt | years |
|---|---|---|---|
| 228 USSR | SUN | 436.8 | 1961-1991 |
| 15 Belgium-Luxembourg | BLX | 43.9 | 1961-1999 |
| 51 Czechoslovakia | CSK | 38.1 | 1961-1992 |
| 248 Yugoslav SFR | YUG | 37.8 | 1961-1991 |
| 186 Serbia and Montenegro | SCG | 12.8 | 1992-2005 |
| **subtotal, dissolved entities** | | **569.4** | |
| 999 RoW bucket | ROW | 25.5 | 1850-2023 |
| 227 Tuvalu + 148 Nauru | TUV, NRU | 0.010 | 1961-2023 |
| **total** | | **595.0 (3.48% of 1850-2023)** | |

The dissolved entities are **96% of the loss**, and **15.3% of all world meat carcass tonnage in
1961** (14.2% in 1980, 13.9% in 1990) disappears from the extension this way. After PR #552 the
loss is at least reported with its size rather than silent, and can be opted out of with
`unclassified = "global_mean"` -- but a *world* mean is a poor instrument for these five, because
their successors' regions are known.

## Why this is cheap now

PR #508's `.gleam_region_of()` already resolves precisely these five from a
`polity_area_code`-keyed override table, and #508 documents the reasoning for each: BEL/LUX are
both Western Europe; CZE/SVK both Eastern Europe; all six Yugoslav successors Western Europe; the
USSR takes its largest successor's region. Those are the same GLEAM regions the energy extension's
schemes need as input, so the work is a consumer-side wiring job, not new science, *for the four
whose successors agree*.

## The decision it still needs

The energy extension's three schemes (`development3`, `region5`, `detailed15`) are keyed on
`oecd` / `eu27` / `continent` / `faostat_region`, not on `gleam_region`, so a GLEAM region alone
does not fill them: someone has to say what `oecd`/`eu27` mean for the USSR in 1975 or for
Czechoslovakia in 1990. That is a real methodological call (`needs-expert`), and it is why this is
an issue rather than a follow-up commit:

1. **Resolve the flags from the successor territories** the same way #508 resolves the region,
   weighting by production where successors disagree.
2. **Treat every dissolved entity as "Others" / non-OECD / non-EU27** for the period it existed,
   which is historically right for CSK/SUN/YUG and wrong for BLX (Belgium and Luxembourg were both
   founding OECD members).
3. Keep them out and accept the reported loss (the status quo after PR #552).

Blocked on #508 merging first, since it owns `.gleam_region_of()`.

Refs #458, #492, #415, #465.

lbm364dl:
## What was wrong

`gleam_geographic_hierarchy` is the country universe of the whole energy
extension: all three grouping schemes in `.energy_country_grouping()` are
`transmute()`d off it. It is a **present-day** table of 204 sovereign states, so
the dissolved federations have no row in it, get no energy intensity, and their
meat production leaves the extension without a number.

PR #508 (#465) solved exactly this for the **livestock GHG** extension with
`.gleam_region_of()` and a `polity_area_code`-keyed override table, and said
explicitly that `R/energy_co2_extension.R` "keeps its own ISO3-keyed `iso2reg`"
and that wiring it was not its call. PR #552 (#492) then made the loss
**visible** and gave it a world-mean escape hatch, and said the strictly better
fix for 96% of the tonnage was to resolve these five properly. Nobody owned
that, which is what #553 is.

The issue asks a real question the region alone cannot answer: the three schemes
are keyed on `oecd` / `eu27` / `continent` / `faostat_region`, not on
`gleam_region`, so someone has to say what OECD and EU mean for the USSR in 1975
or for Czechoslovakia in 1990.

At the base commit (`de7e2ff9`), driving `build_energy_co2_extension()` over the
real `get_primary_production()` output (**6,305,656 rows, 1850-2023, 217 areas**):

```
carcass rows entering .energy_co2e_by_group: 111715   tonnes: 17,084,808,616
after inner_join(area_code -> iso3):         111715   tonnes: 17,084,808,616   (loses nothing)
rows left with NA ef_total (these disappear):   840   tonnes:    569,447,998   = 3.3331%
```

| area | iso3 | carcass Mt | years |
|---|---|---|---|
| 228 USSR | SUN | 436.8 | 1961-1991 |
| 15 Belgium-Luxembourg | BLX | 43.9 | 1961-1999 |
| 51 Czechoslovakia | CSK | 38.1 | 1961-1992 |
| 248 Yugoslav SFR | YUG | 37.8 | 1961-1991 |
| 186 Serbia and Montenegro | SCG | 12.8 | 1992-2005 |
| 227 Tuvalu | TUV | 0.007 | 1961-2023 |
| 148 Nauru | NRU | 0.004 | 1961-2023 |
| **dissolved subtotal** | | **569.4** | |

Share of world meat carcass tonnage lost: **15.2% in 1961**, 14.1% in 1980,
13.8% in 1990, 0.41% in 2000, ~0 from 2010 on.

**The issue's headline holds exactly (569 Mt), but one line of its table no
longer does, and re-measuring mattered:** it reported 595.0 Mt including 25.5 Mt
of bucket 999. Since the Rest-of-World fold was lifted (#628), **bucket 999
contributes nothing to this loss at all** — its 21 reporting members now price
under their own ISO3 codes. The dissolved entities are therefore no longer 96%
of the loss, they are **99.998%** of it, and #492's `"global_mean"` is now a
world mean applied to five European federations and two Pacific islands.

## What I changed

`R/energy_co2_extension.R` only, plus its tests, docs and `NEWS.md`.

A fourth value of the existing argument, `unclassified = "historical_region"`,
which is a **superset** of `"polity_region"`: it does everything that does for
the live areas GLEAM omits, and additionally builds one
`gleam_geographic_hierarchy`-shaped row per **dissolved** entity so
`.energy_country_grouping()` runs GLEAM's own `case_when()` on it. As with #415's
treatment, **no grouping label is typed in** — only the inputs those rules read:

* `continent` from the polity crosswalk;
* `gleam_region` from **#508's merged `.gleam_region_overrides()`**, not a second
  copy of that decision (it only feeds the dressing fraction) — this is the
  compose-don't-duplicate the issue asks for;
* `oecd` / `eu27` from a new six-row table of the membership each entity held
  **while it existed**.

Which entities qualify is derived, not listed: a self-reporting area with no
GLEAM row **all of whose polity periods ended before the crosswalk's open end**.
That is what separates the dissolved five (plus the Netherlands Antilles) from
the live omissions and from buckets 901-906/999, whose periods run to the open
end — so this cannot silently pre-empt #415's or #492's decisions. An entity the
membership table does not cover falls out of the join and keeps being reported
rather than defaulting to zeros nobody chose.

Rows carry `method_energy = "GLEAM_3.0_energy_meat_historical_region"`, per row,
so a coarser treatment is never a silent fallback and the three treatments stay
distinguishable inside one build.

One incidental correctness fix the new path needed: `.energy_gleam_continent()`
translates the crosswalk's `"North America"`/`"South America"` into GLEAM's
`"Americas"`. Without it an American area matches no `region5` branch and falls
through `.default` to **"Non-OECD Asia"**. Measured: no area reaching either
derived treatment today is in the Americas with production, so this maps a
latent mis-grouping, not a live one, and it moves no number.

### The memberships, each with its date (verified, not assumed)

| entity | oecd | eu27 | why |
|---|---|---|---|
| 15 Belgium-Luxembourg | 1 | 1 | BEL and LUX both signed the OECD Convention on 14 Dec 1960 (founding members, in force 30 Sep 1961) and are both EEC founders (Treaty of Rome, in force 1 Jan 1958) — so OECD and EU across its whole 1961-1999 span |
| 51 Czechoslovakia | 0 | 0 | Czechia joined the OECD 21 Dec 1995, Slovakia 14 Dec 2000, both the EU 1 May 2004 — all after the 1992 dissolution |
| 228 USSR | 0 | 0 | no Soviet successor is an OECD member; the Baltics joined the EU in 2004 |
| 248 Yugoslav SFR | 0 | 0 | Slovenia joined the OECD in 2010, Croatia the EU in 2013 |
| 186 Serbia and Montenegro | 0 | 0 | neither successor is in either body |
| 151 Netherlands Antilles | 0 | 0 | the OECD member is the Kingdom of the Netherlands; its Caribbean territories are EU Overseas Countries and Territories, associated with the Union but outside its territory |

No successor weighting is involved: for the four non-members the successors
agree, they were all outside.

## How I verified

**The default is bit-identical on the real input.** Full
`build_energy_co2_extension(data = list(primary_prod = get_primary_production()))`
before and after: 181,831 rows both times, `identical(before, after)` **TRUE**,
`sum(impact_u) = 6.530863856531e12` both times.

**What opting in does** (measured, so the decision is not taken blind):

* +1,190 rows (183,021), +7 areas (204), **no area and no row removed**;
* the 181,831 shared rows are unchanged to the last bit (`max|Δ| = 0`) and not
  one of them changes `method_energy`;
* total energy CO2e **+2.399%** over 1850-2023 — **+12.0% in 1961**, +11.5% in
  1980, +11.3% in 1990, +0.26% in 2000, **0% from 2010**;
* recovered per area (kg CO2e): USSR 1.197e11, Belgium-Luxembourg 1.454e10,
  Yugoslav SFR 1.017e10, Czechoslovakia 9.17e9, Serbia and Montenegro 3.12e9,
  Tuvalu 4.2e5, Nauru 2.9e5;
* the build now emits no unpriced-meat warning at all, because nothing in the
  real input is left unpriced.

**Structural checks** (a PR was reverted this week for conserving mass while
splitting a bucket): 0 duplicated `(year, area_code, item_cbs_code)` keys before
and after; **one `area_code` -> one `polity_area_code`** (0 violations) and one
`reporting_polity_name` per `(area_code, year)` (0 violations, before and
after); `setdiff(before$area_code, after$area_code)` empty. The recovered areas
carry their own labels ("USSR (1945-1991)", "Czechoslovakia (1947-1993)"), not a
shared one.

**Tests**: 6 new blocks in `tests/testthat/test_energy_co2_extension.R`.
**Load-bearing, proved by reverting only `R/energy_co2_extension.R` to
`origin/main` and re-running the file: FAIL 14, PASS 78** (9 assertion failures
plus 5 errors — the helpers and the argument value do not exist). With the
change: **FAIL 0, PASS 114**. One of the six is a status-quo pin that holds
before and after by design (adding dissolved production under the default must
neither add a row nor move another area's value), and one drives the failure
branch: with the membership table mocked down to one entity, the others stay
unpriced and are still named in the warning.

**Gates**: `air format .` (binary) clean; `devtools::document()` run and `man/`
committed; `lintr::lint_package()` with the CI config — **No lints found** (it
caught `.energy_dissolved_hierarchy_rows` at 32 characters; renamed);
`devtools::test()` full suite green; both changed files pure ASCII. No new NSE
symbol (`.data$` throughout), so `R/utils.R` is untouched; no new documented
topic, so `_pkgdown.yml` is untouched (verified: nothing in `man/` missing
from it).

## Moves published values

**No.** The default `unclassified = "drop"` is bit-identical on the full real
input, measured above, which is why this is not a draft. What changes is that a
caller can now opt into pricing 569 Mt of production that today leaves the
extension silently. `NEWS.md` carries the numbers.

## The open decision

This is a **science decision** and the default deliberately does not take it.
Both alternatives are measured:

1. **Contemporaneous membership (what this PR implements when opted into).**
   Czechoslovakia lands on "Others"/"Non-OECD Europe"; Belgium-Luxembourg on
   "OECD"/"EU 27", identical to BEL and LUX.
2. **Successors' present-day flags** (the issue's option 1). The two readings
   disagree on exactly one entity, Czechoslovakia, because CZE and SVK are both
   OECD and EU today: it would be **+38.0%** on Czechoslovakia (9.17e9 ->
   1.266e10 kg CO2e) and +0.052% on the world total. I did not implement it,
   because making Czechoslovakia an OECD and EU-27 member in 1990 is
   anachronistic — but it is one row of `.energy_dissolved_membership()` away if
   you prefer it.
3. **The USSR's `detailed15` bucket.** It resolves to "Non-OECD Europe" from its
   continent, while GLEAM's `detailed15` also has a "Russian Federation" bucket
   that #508's region override points the USSR's *dressing* at. Putting the USSR
   on the Russian factor instead would be **-9.6%** on the USSR (1.197e11 ->
   1.082e11 kg CO2e, -0.18% of the world total): bovine is identical, the
   difference is direct energy on pig (0.139 vs 0.111), poultry (0.338 vs 0.298)
   and mutton/goat (0.261 vs 0.231 kg CO2e/kg LW). Left as it is because
   "Non-OECD Europe" comes out of GLEAM's own rules while "Russian Federation"
   would be a hand-placement, but say the word and it is a one-line change.

Option 3 of the issue (keep them out) remains the shipped default.

- **Did not change any default**, and did not touch `"drop"`, `"polity_region"`
  or `"global_mean"` behaviour.
- **Did not touch `gleam_geographic_hierarchy`.** It is parsed from the GLEAM
  workbook; adding rows would make the package's copy diverge from the published
  source with nothing recording it — the same reasoning #415, #475 and #552
  fixed on.
- **Did not touch `.gleam_region_overrides()` or `R/livestock_enteric.R`.** This
  consumes #508's decision; it does not restate it.
- **Did not assign RAFR/RASI/REUR/RLAM/RNAM/ROCE or bucket 999 anything.** They
  are aggregates, they carry no meat production, and #492 already ruled on 999.
- **Did not guess a single membership.** Every date above is from the OECD's and
  the EU's own published accession records; where I could not attribute a
  membership to the entity itself (the Netherlands Antilles) I said why it has
  none rather than inferring one.
- **Did not run `build_footprint()`.** The extension is exported but not wired
  into the IO core, so there is no published footprint for it to move.
- **Did not widen this PR to the two adjacent gaps I measured.** Filed instead.

Closes #553

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,29 @@
 # whep (development version)
 
+* **`build_energy_co2_extension(unclassified = "historical_region")` prices the
+  dissolved federations instead of losing them (#553).** Measured on the real
+  `get_primary_production()` output (6,305,656 rows, 1850-2023), 569.4 Mt of
+  meat carcass production — 3.33% of all of it, and 15.2% of the world's 1961
+  tonnage — gets no energy intensity and leaves the extension, because
+  `gleam_geographic_hierarchy` is a present-day country table with no row for
+  the USSR, Belgium-Luxembourg, Czechoslovakia, the Yugoslav SFR or Serbia and
+  Montenegro. Those five are now 99.998% of the loss: since the Rest-of-World
+  fold was lifted (#628) bucket 999 no longer contributes to it at all. The new
+  treatment groups them by running GLEAM's own scheme rules on the OECD and EU
+  membership they themselves held while they existed — Belgium and Luxembourg
+  were OECD founding members and EEC founders, so Belgium-Luxembourg is OECD/EU
+  27; no successor of the other four was in either body before the entity
+  dissolved, so they are non-OECD, non-EU. Rows carry
+  `method_energy = "GLEAM_3.0_energy_meat_historical_region"`, and the option
+  is a superset of `"polity_region"`.
+
+  **No published value changes**: the default `unclassified = "drop"` is
+  bit-identical on the full real input (181,831 rows, `sum(impact_u) =
+  6.530863856531e12` before and after, `identical()` TRUE). Opting in adds
+  1,190 rows and 7 areas, moves no shared row by any amount, and raises total
+  energy CO2e by **+2.40%** over 1850-2023 — **+12.0% in 1961**, +11.3% in
+  1990, +0.26% in 2000 and 0% from 2010 on.
+
 * **`get_primary_production()`, `get_wide_cbs()` and `get_processing_coefs()`
   take a `years` argument.** A scoped request now builds only that window
   instead of building 1850-2023 and discarding the rest. Measured for 2010:
--- a/R/energy_co2_extension.R
+++ b/R/energy_co2_extension.R
@@ -34,12 +34,14 @@
 #' 901-906) and the dissolved entities GLEAM's present-day country table cannot
 #' carry (USSR, Czechoslovakia, Yugoslavia, Belgium-Luxembourg, Serbia and
 #' Montenegro). The size of that loss is now **reported** on every build rather
-#' than left to be inferred, and two opt-in treatments recover it instead of
+#' than left to be inferred, and three opt-in treatments recover it instead of
 #' losing it: `unclassified = "polity_region"` groups the **live** reporting
-#' areas GLEAM omits (today Nauru and Tuvalu) from the polity crosswalk, and
+#' areas GLEAM omits (today Nauru and Tuvalu) from the polity crosswalk,
+#' `unclassified = "historical_region"` additionally groups the **dissolved**
+#' entities from the membership they themselves held while they existed, and
 #' `unclassified = "global_mean"` prices **every** unclassifiable area at the
 #' world-mean GLEAM intensity. The default keeps the historical behaviour; see
-#' whep#415 and whep#492.
+#' whep#415, whep#492 and whep#553.
 #'
 #' @param method Estimation method. Only `"gleam"` (default), the GLEAM 3.0
 #'   per-live-weight factors, is currently available.
@@ -54,6 +56,11 @@
 #'   polity in `polity_area_crosswalk`, running GLEAM's own scheme rules on that
 #'   continent, and marks those rows `"GLEAM_3.0_energy_meat_polity_region"`;
 #'   the aggregate buckets and dissolved entities still drop.
+#'   `"historical_region"` does everything `"polity_region"` does and also
+#'   groups the **dissolved** entities (USSR, Czechoslovakia, the Yugoslav SFR,
+#'   Belgium-Luxembourg, Serbia and Montenegro, the Netherlands Antilles) from
+#'   the OECD and EU membership they held while they existed, marking those rows
+#'   `"GLEAM_3.0_energy_meat_historical_region"`.
 #'   `"global_mean"` instead prices every unclassifiable area at the unweighted
 #'   world mean of the published GLEAM factors, marking those rows
 #'   `"GLEAM_3.0_energy_meat_global_mean"`.
@@ -63,7 +70,9 @@
 #' @return A tibble with columns `year`, `area_code`, `item_cbs_code`,
 #'   `impact_u` (energy-use emissions in kilograms CO2e) and `method_energy`
 #'   (`"GLEAM_3.0_energy_meat"`, `"GLEAM_3.0_energy_meat_polity_region"` for
-#'   rows grouped from the polity crosswalk, or
+#'   rows grouped from the polity crosswalk,
+#'   `"GLEAM_3.0_energy_meat_historical_region"` for dissolved entities grouped
+#'   from their own era's memberships, or
 #'   `"GLEAM_3.0_energy_meat_global_mean"` for rows priced at the world mean),
 #'   plus the polity columns below.
 #'
@@ -76,7 +85,7 @@
 build_energy_co2_extension <- function(
   method = c("gleam"),
   data = list(),
-  unclassified = c("drop", "polity_region", "global_mean"),
+  unclassified = c("drop", "polity_region", "historical_region", "global_mean"),
   example = FALSE
 ) {
   method <- match.arg(method)
@@ -251,7 +260,7 @@ build_energy_co2_extension <- function(
   }
   n_gaps <- nrow(gaps)
   areas <- sort(gaps$area_name)
-  if (identical(unclassified, "polity_region")) {
+  if (unclassified %in% .energy_derived_treatments()) {
     cli::cli_inform(c(
       "i" = "{n_gaps} live reporting area{?s} {?has/have} no row in
          {.field gleam_geographic_hierarchy} and {?is/are} grouped from the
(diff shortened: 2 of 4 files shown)


lbm364dl:
Verified independently and merging. The one red check was **CANCELLED**, not failed — superseded, not broken.

Your default-neutrality claim reproduces exactly:

| | rows | `impact_u` |
|---|---|---|
| `main` | 181,831 | 6.530864e12 |
| this PR, `unclassified = "drop"` | 181,831 | **6.530864e12** |

Bit-identical to every digit, so nothing published moves by merging.

And the opt-in does what it says:

| | rows | `impact_u` |
|---|---|---|
| `"drop"` | 181,831 | 6.530864e12 |
| `"historical_region"` | 183,021 | 6.687550e12 (**+2.4%**) |

Recovering exactly the dissolved federations — areas 15, 51, 148, 186, 227, 228, 248. Belgium-Luxembourg, Czechoslovakia, Serbia and Montenegro, the USSR and Yugoslav SFR are the same set #508 measured as silently taking GLEAM's Global default, so the two changes line up.

This is the shape I want more of in this epic: the plumbing lands, the status quo stays the default, and the science decision becomes a one-argument choice backed by a measurement rather than a rewrite someone has to attempt later. Full suite 6436 pass, 0 fail against current `main`.

The decision itself stays yours and I have not taken it.

Issue #514, opened 2026-08-04
Two live polities share one polygon: 441 cells claim up to 2x their own area


eduaguilera:
**Two live polities can be assigned the same polygon**, so the ground inside it is claimed twice. Found while building the polycell spatial support table on branch `edu/polycell-spatial-support`; the numbers below come from a full-domain run of that producer against `whep::polities` on `main`.

## Problem

`GNQ-1968-2025` (Equatorial Guinea) and `STP-1800-2025` (São Tomé and Príncipe) are both live in 2015 and both carry the **same** geometry. Intersecting the 0.5 degree grid against the live polity set therefore hands cell `(10.25, 1.75)` to each of them in full, and the cell's claimed territory comes to **2.0000×** its own area.

This is not a border-sharing artefact — sharing is normal and the intersection handles it. It is the same polygon appearing under two `polity_code`s.

## Evidence

Measured over all live real polities (`wiki_status` not in `retired`/`superseded`, `polity_type != "aggregate"`), intersecting against the 0.5 degree grid:

| Quantity | Value |
| --- | --- |
| Cells in 2015 whose claimed area exceeds the cell | **441** of 67,629 |
| Worst ratio | **2.0000×** (cell `(10.25, 1.75)`) |
| Excess territory, 2015 | **12.68 Mha** |
| Cell-intervals affected across all epochs | **126,798** |

## Why it matters

Any area-weighted allocation over these cells double-counts: a per-hectare rate applied to claimed territory delivers the cell's quantity twice, and re-aggregating polycells to a polity inflates whichever polity holds the duplicate. It is invisible today because the current crosswalk renormalises each cell's polity fractions to sum to 1, which silently absorbs the overlap into a plausible-looking split rather than surfacing it.

## What the producer does about it, and what it deliberately does not

`build_polycell_support()` emits the overlap as `attr(., "overlap")` with a warning and **does not renormalise it away**. That is intentional: deciding which of two polities owns a piece of ground is a territorial judgement, and silently rescaling would reproduce exactly the defect that hides it today. So this issue is about fixing the underlying polygon assignment, not about adding a normalisation step downstream.

## Target state

Each live polity has its own geometry, or — where two genuinely share a footprint in the source data — the overlap is explicit and resolved deliberately rather than inherited.

## Acceptance criteria

- No cell's claimed territory exceeds its own area beyond geodesic tolerance.
- A regression check pins the `GNQ`/`STP` pair specifically.
[...]

Issue #425, opened 2026-07-31
CRITICAL: commodity-balance cast returns row COUNTS not quantities — 37% of the CBS frame, 259x on food, affects main


lbm364dl:
> [!CAUTION]
> **Every value produced by `.select_best_source()`'s cast is a ROW COUNT, not a quantity, in every real build — including on `main`.** Not the 31,642 duplicated cells: all of them.

## The proof, in one `dcast()`

`data.table::dcast()` falls back to `fun.aggregate = length()` when *any* duplicate key exists — and it applies that function to **every cell**, not only the duplicated ones.

```r
d <- data.table(k = c("A","A","A","B","C"), source = c("x","x","y","x","y"),
                value = c(10, 20, 5000, 70000, 90000))
dcast(d, k ~ source, value.var = "value", fill = NA)
```

```
        k     x     y
   <char> <int> <int>
1:      A     2     1     <- A.x is the COUNT of its two rows; A.y should be 5000
2:      B     1    NA     <- should be 70000
3:      C    NA     1     <- should be 90000
```

One duplicate anywhere converts the whole table to counts, and the columns come back `<int>`.

## Measured in a real build (2010-2023, `build_commodity_balances`)

```
wide = 1,275,817 rows
classes = FAOSTAT_FBS_New:integer  FAOSTAT_FBS_Old:integer  FAOSTAT_prod:integer
maxima  = FAOSTAT_FBS_New:4        FAOSTAT_FBS_Old:4        FAOSTAT_prod:1
```

These columns hold commodity-balance **quantities in tonnes**. A maximum of 4 is impossible. All three primary-source columns are integer counts across all 1.28M rows. The 31,642 duplicate `(key, source)` combinations reported by the guard (#418) are merely what *triggers* the fallback; the damage is total.

## What it costs downstream, measured

Re-running the full range with `fun.aggregate = sum` (and with `first` — they agree to within 0.1%, since what matters is that values are quantities at all, not which is chosen):

| column | today (counts) | with quantities | ratio |
|---|---|---|---|
| food | 1.212e9 | 3.138e11 | **259×** |
| seed | 2.756e8 | 5.035e10 | **183×** |
| stock_withdrawal | 9.340e9 | 1.492e11 | 16× |
| feed | 8.156e10 | 9.701e11 | 11.9× |
| import | 1.566e10 | 1.110e11 | 7.1× |
| export | 2.995e10 | 7.870e10 | 2.6× |
[...]

Issue #184, opened 2026-07-22
[CRAN] Vignette trade-sources-coverage.Rmd fetches a Google Sheet at build time


lbm364dl:
## Blocker for CRAN submission

`vignettes/trade-sources-coverage.Rmd` setup chunk (`:10-15`) sets only `collapse`/`comment` — **no `eval = FALSE`**. The chunk at `:24` calls `googlesheets4::gs4_deauth()` then `googlesheets4::read_sheet("1UdwgS87...")` over the network, and every downstream chunk depends on the result. `VignetteBuilder: knitr` is set and the vignette is not in `.Rbuildignore`, so `R CMD check` builds it — on CRAN (offline / no Google auth) the build **errors**.

Secondary: `googlesheets4` and `ggplot2` are in Suggests but used unconditionally (no `requireNamespace()` guard).

## Fix

Mirror `vignettes/footprint-analysis.Rmd:10-16`, which correctly sets `knitr::opts_chunk$set(eval = FALSE)`. Either set `eval = FALSE` globally here, or precompute/bundle a small fixture so no network is hit at build, and guard Suggests packages with `requireNamespace(..., quietly = TRUE)`.

---
_CRAN-readiness pass (2026-07-22)._

Issue #183, opened 2026-07-22
[CRAN] Package tarball exceeds the 5 MB limit (polities.rda + large CSVs)


lbm364dl:
## Blocker for CRAN submission

CRAN rejects source tarballs over ~5 MB. Current payload:
- `data/` = **5.1 MB**, dominated by **`data/polities.rda` = 4.7 MB** (already xz-compressed — won't shrink further in the tarball).
- `inst/extdata/` = **4.1 MB** (`coefs/weed_npp_scaling.csv` 1.7 MB, `Crops_manure_N.csv` 928 KB, `mueller_synthetic_n.csv` 272 KB, `Synthetic_N_Grassland_share.csv` 136 KB).

Combined this lands well above the ceiling before `R/` and `man/`.

## Fix options

1. Move the large exported dataset(s) — above all `polities.rda` (geometries) — out of the CRAN package: ship a simplified/slimmed version, or serve the full version through the existing `pins`/`whep_read_file()` remote-data mechanism, or split a companion data package.
2. Downsize `inst/extdata/coefs/weed_npp_scaling.csv` and `Crops_manure_N.csv`.
3. Add `inst/scripts/` (520 KB) and dev markdown (`inst/LIVESTOCK_AUDIT.md`, `inst/performance_profile.md`) to `.Rbuildignore` — they install into the package today but are developer-facing.

Target: source tarball < 5 MB.

---
_CRAN-readiness pass (2026-07-22). Verify final size with `R CMD build` + check the `.tar.gz`._

lbm364dl:
Correction: whep is **already on CRAN** (0.3.0), so this is not necessarily a hard blocker — check whether the payload grew since the last accepted tarball. If 0.3.0 shipped under 5 MB and `polities.rda` / the CSVs have since grown past it, this is a **size regression** to fix before 0.4.0; if the prior version was already near the limit (or CRAN granted a data exception), confirm the current `R CMD build` `.tar.gz` size against that baseline before acting.

PR #662, opened 2026-08-07
Refresh the polities snapshot: geometry now loads under s2

lbm364dl:
> *Filed by Claude (Claude Code). Every figure below is measured against this repo and whep-polities; where a number is an inference rather than a measurement I say so.*

Closes the R-side half of #514.

## The committed snapshot cannot be intersected with R's defaults

On `main`:

```r
sf::st_area(sf::st_geometry(polities)[polities$has_geometry])
#> Error: Loop 3 is not valid: Edge 0 crosses edge 2
```

Any consumer using `st_intersection` or `st_area` over the live polity set either **aborts**, or has already called `sf_use_s2(FALSE)` and is silently computing **planar areas on a global table**. With this snapshot the same call returns for all 713 geometries.

Regenerated by `data-raw/table_mappings.R` against whep-polities [`b25cb7b`](https://github.com/eduaguilera/whep-polities/commit/b25cb7b). **The generator is unchanged** — only its input moved (verified by diffing `data-raw/table_mappings.R` and `data-raw/_labels.R` across the range).

| table | change |
| --- | --- |
| `polities` | 749 → **751** rows, 703 → **713** geometries |
| `polity_label_aliases` | 869 → **902** rows |
| `polity_area_crosswalk` | 596 rows, unchanged |
| `items_cbs` / `items_prod` | unchanged |

## What moved, measured rather than assumed

Upstream repaired 44 GEOS-invalid and 10 s2-unloadable geometries, so 191 rows' planar area changes. The two large ones are both fixes whose declared area now agrees:

| polity | before | after | why |
| --- | --- | --- | --- |
| `SER-1913-1918` | 158,054 km² | **90,538** | carried Bosnia and Montenegro; page declares 90,599 |
| `FID-1887-1954` | 699,133 km² | **753,049** | contained Kerguelen; page declares 753,049 |
| `IDN-OTH-1949-1951` | *(no geometry)* | **1,332,129** | new polygon, excluding West Papua |

Four rows added (`CHL-1810-1884`, `LBY-1949-1951`, `PAPNG-1920-1949`, `TKL-1800-2025`), two removed by rename (`CHL-1810-1883`, `LBY-1950-1951`).

## Published values move — and a green suite is not evidence they didn't

This repo has passed **5,151 tests with a 266× aggregation inflation in place** (#425), so the routing change is measured directly rather than inferred from tests. Of 40,489 `(label, source, year)` triples:

```
re-routed to a different polity          541
newly covered (resolved nothing before)  213
no longer covered                         54
```

Most are the same territory under a new code, or a one-year boundary shift from the `end_year` exclusive/inclusive correction:

```
ROW-1850-2023 -> ROW-1850-2025   173    seven reporting buckets extended to 2025
CAN-1948-2025 -> CAN-1949-2025   113
BLX-1921-1999 -> BLX-1850-1999    39
CAN-1886-1948 -> CAN-1886-1949    30
```

**Two are genuinely different territory, and both are corrections:**

```
TUR-1920-2025 -> TUR-1800-1913    33    pre-1913 "Turkey" is not the republic
FID-1887-1954 -> FCC-1862-1887    21    pre-1887 "Indochina" is Cochinchina
```

## Two consumer-side updates were needed

**`test_resolve_polity_label`** expected `"RoW"` to resolve to `ROW-1850-2023`. That code is `wiki_status: retired` upstream; the live successor is `ROW-1850-2025`.

**`test_build_cbs`** asserted `expect_false("ROW-1850-2023" %in% result$polity_code)`. Once that code was retired the assertion **could never fail again**, because it named a code the table can no longer contain. Replaced with a prefix check, so it keeps testing what was meant — a real country must not fall into the residual bucket — across the next rename too.

Every other occurrence of both codes is in a comment describing past state, and is left alone.

## Verification

- **6,424 tests pass, 0 failures, 4 skips.**
- `st_area` over all 713 geometries returns under s2, where `main` errors.
- One failure seen mid-investigation (`test_n_balance_inputs.R:436`) was **not** from this change — it came from a stale worktree that predated #660, and does not reproduce on this base.


lbm364dl:
**Rebased, not merged — and the reason matters for what is published.**

This branch went `CONFLICTING` on `data/polity_area_crosswalk.rda`. Merging a binary `.rda` conflict would publish a table that **neither generator produced**, so it was not an option.

Main had also changed the generator itself while this sat open — the `mapping_status` reordering that gives 20 rows `not_a_reporting_area` instead of `matched`. My original regeneration therefore used a stale `data-raw/table_mappings.R`, which would have silently reverted that fix.

So: branch reset onto current `main`, the two test edits reapplied (`git apply --3way`, both clean, no markers), and **all five tables regenerated with main's current generator** against whep-polities `793c5fc`. The crosswalk now carries main's `mapping_status` fix rather than reverting it.

Snapshot is now taken after three more upstream merges — whep-polities #182 (the West Papua fix), #183, and #184 (two colonial rows moved from the administering power's iso3 to the successor's).

Re-verified on the new base: **6,635 tests pass, 0 failures, 5 skips** (up from 6,424 because main gained tests meanwhile), and `st_area` over all 713 geometries returns under s2 where `main` errors.

Issue #671, opened 2026-08-07
Year-less outputs (grids, coefficient tables) cannot carry a year-scoped polity: decide what identity they do carry


lbm364dl:
The case #458's rule cannot be applied to literally, and which therefore needs an explicit convention rather than a silent one.

## The problem

A polity code is year-scoped: `ESP-1846-1914` and `ESP-1978-2025` are different rows in `polities`. Several WHEP objects have **no year dimension at all**:

- the cell-polity grid (`build_cell_polity()`, and the polycell table on `edu/polycell-spatial-support`);
- `crop_patterns.parquet`, a static crop-pattern weight;
- coefficient tables keyed by area but not year — the 29 joins in the sibling issue;
- `regions_full` / `polities_cats`, which are present-day grids.

For these, "attach the polity code" has no single answer. #458 already anticipates it for the grid: *"Where no historical boundary exists, resolving to the present-day polity is acceptable — but as an explicit, documented choice with a real `polity_code` column, never as an unstated side effect of joining against a present-day-only lookup."*

That instruction has not been carried out anywhere, and today these objects resolve to whatever the consumer's join happens to imply.

## What needs deciding

For each class, one of:

1. **Present-day polity, documented** — the object describes today's territory and says so.
2. **Polity-period rows** — the object gains a validity interval and becomes year-aware. This is what `build_polycell_support()` does for the grid (`polycell_id = f(cell_id, polity_code)` over the polity's validity interval), so a precedent exists.
3. **Deliberately identity-free** — the object is a coefficient about a *place*, not a *state*, and carrying a polity would be false precision.

Option 3 is a real answer and probably the right one for some coefficient tables; the point is that it should be written down rather than inferred from the absence of a column.

## Why now

PR #579 landed `build_cell_polity(area_key = )` carrying **both** the bucket and the raw grid code, on @eduaguilera's argument that a derived key must not overwrite the raw one. Extending that to the polity is the natural next step, and **#582** is the same seam from the output side. Deciding this once would settle both, plus #461 and #462.

## Measured context

`(area_code, year)` determines `polity_code` with 0 violations of 17,184 — so for any object that *does* have a year, the identity is already recoverable. This issue is only about the objects that do not.

Part of the polity migration epic #458.

PR #674, opened 2026-08-07
feat(water): say when a water-balance cell-year names a polity that did not exist (#462)

### LINKED ISSUE #462: Year-validity check in build_water_balance()'s cell-polity aggregation
**Problem**
`build_water_balance()` (both `resolution="grid"` and `"polity"`) and `get_soc_climate_drivers()`
attach territory purely by joining LPJmL cells to `data$cell_polity` on `(lon, lat)`, then group by
`(year, area_code)` — never checking whether `area_code`'s polity was valid in that row's `year`.
The LPJmL run spans 1901-2009. MEASURED: 54 of 266 crosswalk `area_code`s have no polity valid
before some year after 1901 (e.g. Azerbaijan, `area_code 52`, only has `AZE-1991-2025` from 1991 —
so every 1901-1990 water-balance row for Azerbaijani cells is reported under a country identity
that did not yet exist).

**Evidence**
- `R/water_balance.R:444-456` (`.wb_attach_polity`, lon/lat-only join, no year),
  `:539-565` (`.wb_aggregate_polity`, groups by year+area_code with no validity check).
- MEASURED: `polity_area_crosswalk[area_code==52,]` → single row, `polity_start_year=1991`.
- MEASURED: 54/266 area_codes have `min(polity_start_year) > 1901`.
- `R/lpjml_hydrology.R:4-5` — run spans 1901-2009.

**Target state**
`.wb_aggregate_polity()`/`.assemble_soc_drivers()` resolve `(area_code, year)` through
`whep::polity_area_crosswalk` (reusing `.add_polity_columns_dt()`'s pattern from
`R/read_raw_inputs.R`), warning and dropping any cell-year whose `area_code` has no polity valid at
that year, in the same style as the existing `.wb_warn_uncovered_cells()`.

**Acceptance criteria**
- No water-balance "polity" row is attributed to a polity outside its valid year window.
- A visible warning names the count/years/area_codes dropped.

**Depends on**: blocked by #460 (needs `data$cell_polity` to carry polity columns first).

**Decision needed**: (a) route pre-validity cell-years to a predecessor polity (not possible from
grid geometry alone) vs. (b) drop/flag them — (b) is the pragmatic default given present-day-
polity-fallback policy, but changes any already-published pre-1991-Azerbaijan-style totals.

---

---

lbm364dl:
## What was wrong

`build_water_balance()` and `get_soc_climate_drivers()` attach territory by
joining LPJmL cells to `data$cell_polity` on `(lon, lat)`. That crosswalk is a
**present-day rasterization with no year dimension**, while polity validity
**is** year-scoped. So a cell labelled `area_code` 52 carries that label in 1901
exactly as it does in 2009, and `.add_reporting_polity_columns()` then resolves
1901 through the nearest-period fallback in `.add_polity_columns_dt()` to
`AZE-1991-2025` — a state that did not exist in 1901.

The substitution is *already computed*: `.add_polity_columns_dt()` records it as
`mapping_status == "out_of_span"`. But that column is stripped from published
outputs by design (`R/polity_columns_doc.R` says so), so on the water-balance
path it is **silent**. Nothing warns, and there was no way to ask for the rows
to be excluded.

**Re-derived the "blocked by #460" claim: it is stale.** #460 is closed and
PR #579 landed `build_cell_polity(area_key = )`. More to the point, the
dependency was never the real one — `build_water_balance()` already ends in
`.add_reporting_polity_columns()`, which *is* year-aware. What is missing is not
polity columns on the grid; it is that the consumer never says when the year
could not be honoured. This PR is unblocked and does not touch the grid.

**Corrected from the issue text.** #462 says "54 of 266 crosswalk `area_code`s".
Measured on the deployed parquet the number is **21 of the 182 area codes the
grid actually carries** — the issue counted the whole crosswalk, not the grid.

At base commit `99dd49f0`, against the real deployed
`cell_polity_fraction.parquet` (`WHEP_POLITY_FRACTION_PATH`: 62,808 rows /
58,791 cells / 182 area codes) over the LPJmL run's 1901-2009 span
(`R/lpjml_hydrology.R`):

```
out-of-span (area_code, year) pairs : 1,948 of 19,838
distinct area codes affected        : 21 of 182
cells behind those codes            : 14,761 of 58,791  (25.1%)
year range                          : 1901-2009
```

| area_code | resolves to | out-of-span years |
|---|---|---|
| 277 South Sudan | `SSD-2011-2025` | 1901-2009 (**all 109**) |
| 272 Serbia | `SRB-2006-2008` | 1901-2005 |
| 273 Montenegro | `MNE-1913-1918` | 1901-2005 |
| 167, 199 Czechia / Slovakia | `CZE-1993-2025`, `SVK-1993-2025` | 1901-1992 |
| 178 Eritrea | `ERI-1889-1952` | 1901-1992 |
| 80, 98, 198 BiH / Croatia / Slovenia | `BIH-1992-2025` etc. | 1901-1991 |
| 1, 52, 57, 73, 108, 113, 146, 185, 208, 213, 230, 235 | the 12 post-Soviet successors | 1901-1990 |

`polity_coverage_gaps()` could already report this for a *built* table; the
build itself said nothing.

## What I changed

`R/water_balance.R` only. Both entry points gain
`polity_validity = c("keep", "flag", "drop")` and route their tail through one
new private helper, `.wb_resolve_polity_validity()`:

- **`"keep"` (default)** — today's rows, today's numbers, plus a `cli_warn()`
  naming the row count, the year range and every area code involved, in the
  style of the neighbouring `.wb_warn_uncovered_cells()`.
- **`"flag"`** — same rows and same numbers, plus the per-row logical
  `reporting_polity_out_of_span`. This is the multi-method "record the choice in
  a column" requirement, satisfied by a per-row truth rather than a constant
  string: it says *which* rows, not just which mode ran.
- **`"drop"`** — removes exactly those `(area_code, year)` rows.

The gap set is resolved by the **exported** `polity_coverage_gaps()`, on
`distinct(area_code, year)` rather than on the rows, so a multi-million-row
gridded table costs one non-equi join over ~20k pairs. Using that function (and
not a second hand-rolled lookup) is what guarantees the report matches what
`.add_reporting_polity_columns()` actually did — confirmed below.

## How I verified

**Gates, on this branch:**

- `air format .` (the binary) — clean, then `devtools::document()`; `man/`
  committed. No new topic, so `_pkgdown.yml` is untouched and the `comm` check
  is unaffected (only `build_water_balance.Rd` and
  `get_soc_climate_drivers.Rd` changed, both already listed).
- `lintr::lint_package(...)` with the four disabled linters: **0 lints**;
  re-checked per file after the last edit: `R/water_balance.R` 0,
  `tests/testthat/test_water_balance.R` 0.
- `devtools::test(filter = "water_balance")`: **135 passing, 0 failed, 1
  skipped** (the pre-existing CRU-file skip), up from 116.
- Full `devtools::test()`: **6,651 passing, 0 failed, 0 errors**, 9 skipped
  (all pre-existing env-var/local-file guards). `test_water_balance.R`
  contributes 0 warnings.
- No new `utils::globalVariables()` entries: `area_code` and `year` are the only
  NSE symbols used and both are already declared.

**Proving the guard is load-bearing.** I reverted only the two
`.wb_resolve_polity_validity(polity_validity)` calls back to
`.add_reporting_polity_columns()` and re-ran the file: **19 failures + 1 error
across 6 of the 8 new tests** (`a pre-independence cell-year is named, not
silent`, `the validity warning names the count and the codes`, `flag marks
exactly the out-of-span rows`, `drop removes only the out-of-span rows`, `the
grid resolution reports validity too`, `keep is the default…`, `SOC drivers
honour polity_validity as well`). Restored, back to 135/0.

**Structural checks, not a totals diff.** I could not run
`build_water_balance()` end to end — it needs the multi-GB LPJmL NetCDFs, which
are out of budget per the epic — so I drove `.wb_resolve_polity_validity()` over
the **real** key space the polity resolution would produce: every
`(area_code, year)` the deployed parquet can emit across 1901-2009.

| | `"keep"` | `"flag"` | `"drop"` |
|---|---|---|---|
| rows (keys) | 19,838 | 19,838 | **17,890** |
| area codes | 182 | 182 | **181** |
| distinct `reporting_polity_code` | 222 | 222 | 219 |
| distinct `polity_area_code` buckets | 181 | 181 | 181 |
| **area codes mapping to >1 bucket** | **0** | **0** | **0** |
| sum of the carried value | 19,838,000 | 19,838,000 | 17,890,000 |
| flagged rows | — | **1,948 over 21 codes** | — |

- **`"keep"` and `"flag"` are numerically identical** — same rows, same keys,
  same sum. `"flag"` adds one logical column and nothing else.
- **The flag count equals the gap count exactly** (1,948 = 1,948). That is the
  internal-consistency check that matters: the warning and the drop describe the
  same rows `.add_reporting_polity_columns()` actually substituted, not a second
  opinion about the crosswalk.
- **One `area_code` → one label holds in all three modes** (0 codes map to more
  than one `polity_area_code`), which is the #561/#563 failure mode this epic
  reverts PRs for. No bucket is split.
- `"drop"` removes 1,948 pairs over 21 codes; **exactly one area code vanishes
  entirely** — 277 South Sudan, whose only polity `SSD-2011-2025` postdates the
  whole run. Weighting each dropped code by its share of the grid,
  **1,307,298 of 6,408,219 cell-years (20.4%)** would leave the run.

The exact warning on the real grid:

```
! 1948 rows over 21 area codes resolve to a polity that did not exist in that
  row's year (years 1901-2009); they are kept as-is.
i The cell-polity crosswalk has no year dimension, so an early cell carries its
  present-day territory. Area codes: 1, 52, 57, 73, 80, 98, 108, 113, 146, 167,
  178, 185, 198, 199, 208, 213, 230, 235, …, 273, and 277.
i `polity_coverage_gaps()` names the polity each one landed on;
  `polity_validity = "drop"` removes them.
```

**Offline.** The new tests build their own two-cell, two-year fixture and inject
it through `data =`; nothing reads a `WHEP_*` path or the network, so the
`offline-tests` job is unaffected.

## Moves published values

**No, not on the default path.** `polity_validity = "keep"` is the default and
reproduces today's output row-for-row and value-for-value; the only change on
that path is a warning. `"flag"` is identical plus one column.

`"drop"` does move values — it removes 20.4% of the run's cell-years and all of
South Sudan — which is precisely why it is opt-in and not the default.

- **I did not make `"drop"` the default.** The issue offers (a) route
  pre-validity cell-years to a predecessor polity and (b) drop/flag them, and
  calls (b) the pragmatic default. I implemented (b) but did **not** flip it on:
  the rows are not wrong about the *water*. The cells are real territory, the
  hydrology is per cell, and only the polity *name* is anachronistic. Silently
  deleting a fifth of a physical simulation to fix a labelling problem is a
  worse trade than labelling it. Option (a) remains impossible from grid
  geometry alone, as the issue says.
- **I did not touch the cell-polity grid.** #671 is deciding the general
  convention for year-less objects and is running concurrently; my scope is this
  one consumer. Nothing here presumes an answer — if the grid later becomes
  year-aware (polycell, or #671 option 2), the gap set simply goes empty and all
  three modes converge.
- **I did not change the back-cast anchor**, though it is arguably wrong here:
  `.add_polity_columns_dt(backcast_anchor = 1961)` floors 1901-1960 to 1961
  because *FAOSTAT* series are back-cast onto 1961 territory. An LPJmL water
  balance is not back-cast — it is simulated for its own year on present-day
  polygons — so neither 1961 nor the data year is obviously right. Filed as a
  follow-up rather than decided here.
- **I did not migrate the other `data$cell_polity` consumers**
  (`build_n_deposition()`, `build_urban_n()`, `build_ag_land_support()`,
  `aggregate_grass_to_polity()`, the carbon clay path). Same defect class, but
  doing them in one commit is the per-consumer-attribution failure #579
  documented. Filed as a follow-up.
- **I did not add a constant `method_*` column.** The multi-method rule is met
  by `"flag"`'s per-row `reporting_polity_out_of_span`, which is strictly more
  informative than a repeated string, and by the argument itself. Flagging this
  as a conscious reading of the rule, following the same call PR #579 made in
  this subsystem.
- **I did not run `build_water_balance()` end to end.** It needs the multi-GB
  LPJmL rasters; the epic puts that out of budget. Everything above is measured
  at the aggregation/resolution helper over the real deployed grid, which is the
  boundary the change lives at. What I therefore **could not** measure: the
  actual mm/yr each dropped cell-year carries, and whether any downstream
  consumer of a `resolution = "polity"` table breaks on the missing keys under
  `"drop"`.

## The open decision

**Should a gridded physical simulation drop, flag, or keep an anachronistic
polity label — and is the FAOSTAT back-cast anchor the right convention for it?**

The default here is "keep and say so", on the argument that the hydrology is
right and only the name is wrong. The counter-argument is real: a downstream
consumer joining on `reporting_polity_code` gets 1,948 keys whose polygon,
population and period describe a different era than the value does, and 20.4% of
the run is a lot of quietly mislabelled territory. If this project's rule is
that a published row must never name a polity that did not exist, `"drop"` is
the correct default and this PR should flip it in a follow-up — that is a
science decision for the maintainer, not one to take on green CI.

Underneath it: the back-cast anchor. Flooring 1901-1960 to 1961 is right for
back-cast FAOSTAT series and questionable for a year-by-year LPJmL simulation.
Changing it would *increase* the gap count (more years would fail the span), so
the status quo is the conservative choice, but it is a choice.

Closes #462


lbm364dl:
Follow-ups filed rather than widening this PR:

- **#675** — the same year-validity check for the other `data$cell_polity` consumers (`build_n_deposition()`, `build_urban_n()`, `build_ag_land_support()`, `aggregate_grass_to_polity()`, the N-spatialize and HWSD clay/pH paths). One PR per consumer, per #480/#563.
- **#676** — whether the 1961 FAOSTAT back-cast anchor is the right lookup convention for a gridded physical simulation that is not back-cast. Sits next to #671, from the time axis rather than the territory axis.

Scope note for the concurrent work: this PR touches **only** `R/water_balance.R` and its test file. It does not touch `build_cell_polity()` or decide #671's convention — if the grid later becomes year-aware, the gap set here simply goes empty and all three `polity_validity` modes converge.

Issue #680, opened 2026-08-07
Un-folding FAOSTAT 276/277 out of bucket 206 loses 4.2% of the region's tonnage and breaks the ISO3 lookup


lbm364dl:
Option 1 of #414, costed on a real build. Filing rather than doing it, because it is a
modelling decision and — unlike the Rest-of-World un-fold in #628 — it is **not** mass-neutral
as things stand.

## What was measured

`get_primary_production(years = 2015L)`, twice in one session, with
`.polity_crosswalk()` patched between runs so FAOSTAT areas 276 Sudan and 277 South Sudan keep
their own `polity_area_code` instead of folding into bucket 206. Everything else identical.

## Outside the Sudan region, nothing moves at all

| | fold (today) | un-fold |
|---|---|---|
| rows outside areas 206/276/277 | 47,943 | 47,943 |
| values differing on a matched key | — | **0** |
| keys present in only one run | — | **0** |

So the change is strictly local. That is the good news, and it is what makes the option
tractable at all.

## Inside the region, mass is LOST

| unit | fold | un-fold | change |
|---|---|---|---|
| tonnes | 33,975,970 | 32,543,030 | **−4.22%** |
| heads | 211,118,900 | 206,439,200 | **−2.22%** |
| ha | 82,847,640 | 82,639,290 | −0.25% |
| LU | 51,542,520 | 51,477,000 | −0.13% |
| slaughtered_heads | 100,259,100 | 100,259,100 | 0 |

Published area count goes 208 → 210 (2015 build); bucket 206 survives with 2 rows, both
`LUH2_grassland` hectares.

Two series account for all of it:

| `item_prod_code` | item | fold | un-fold | delta |
|---|---|---|---|---|
| 651 | Forage products | 1,432,940 t | **0 t** | −1,432,940 t |
| 1052 | Chickens, layers | 14,118,716 heads | 9,439,000 heads | −4,679,716 heads |

Forage products disappears entirely. Both are derived rather than reported series, so the
mechanism is the one #628 had to work through for bucket 999: a downstream step keyed on the
**bucket** code has an entry for 206 and none for 276/277, so promoting the members drops them
out of that step rather than moving them. Nothing warns; the rows simply are not there.

## And the ISO3 lookup stops being unique by construction

`.iso3c_area_code_lookup()` maps ISO3 → `polity_area_code` and is unique today *because* of
this fold: `SDN` appears in `regions_full` as both area 206 "Sudan (former)" and area 276
[...]

PR #690, opened 2026-08-07
Count only the areas bucket 206 really folds, and name the defect it has (#414)

### LINKED ISSUE #414: Bucket 206 sums Sudan + South Sudan but is attributed to SDN-2011-2025: the value and its polity describe different territories

lbm364dl:
## What was wrong

#414 asks whether bucket 206's value and its polity describe different territories.
`polity_bucket_coverage()`, the diagnostic #547 shipped to answer that, gets both halves of
the answer wrong on today's `main`:

1. **It reports the fold in 65 years; it happens in 14.** The report counts a member whenever
   the year-aware resolver can name a polity for `(area_code, year)`. That resolver answers
   for *every* pair, standing in with the nearest period, so asking it about area 276 in 1961
   returns `SDN-2011-2025` — a territory that does not report then and did not exist.
2. **It calls the label an extent mismatch; the extent is right.** Since #591 the bucket
   carries `SUD-1956-2011`. The polities database publishes that polity's `successor` set as
   `SDN-2011-2025; SSD-2011-2025` — *exactly* the two the bucket folds — so its territory is
   the sum. The defect is temporal: that polity had ended. The warning nonetheless says "The
   value and its `reporting_polity_code` describe different extents."

Both overstate the last identity gap in #458 and point at the wrong repair.

At `99dd49f0`:

```
> polity_bucket_coverage() |> dplyr::count(polity_area_code, coverage)
  polity_area_code coverage     n
             206   partial     65      # 1961-2025, n_member_polities = 3 in every year
```

The 3 members in 1961 are area 206 (real) plus areas 276 and 277 as `out_of_span` stand-ins.

Ground truth, from the `faostat-production` pin (all elements):

| `area_code` | years reported | rows |
|---|---|---|
| 206 Sudan (former) | 1961-2011 | 13,759 |
| 276 Sudan | 2012-2024 | 3,467 |
| 277 South Sudan | 2012-2024 | 2,170 |

The reporting spans **do not overlap in a single year**, so before 2012 bucket 206 sums exactly
one territory. Post-2011 it sums two, split 80.07% Sudan / 19.93% South Sudan by production
tonnage over 2012-2024.

And the extent claim, from `whep::polities`:

```
polity_code    successor
SUD-1956-2011  SDN-2011-2025; SSD-2011-2025
```

## What I changed

`R/polity_folds.R` only, plus its docs and tests.

* **A member counts only when the area reports.** Two bounds, because the resolver applies
  neither for this purpose: `out_of_span` drops a stand-in above a period, and the upstream
  map's first reported year drops one below it (area 276 resolves to `SDN-2011-2025` from
  2011, because that polity starts then, while FAOSTAT begins reporting the area in 2012).
* **New coverage class `"predecessor"`**: the bucket's label has ended *and* its published
  successors are exactly the member set. Narrow by construction — a live polity has no
  successors, and an ended one whose successors merely overlap the members stays `"partial"`.
* **`"partial"` now means only what it says**, a label covering less than the value does. No
  bucket is in it today.
* **The warning names the defect** rather than asserting the wrong one, and fires for both
  classes so nothing that warned before stops warning.

I did **not** decide the science. No live polity means "Sudan and South Sudan"; this makes the
report accurate about that, it does not fill the hole.

## How I verified

Gates, all on `Rscript --vanilla`:

* `air format .` (binary) — clean.
* `devtools::document()` — `man/polity_bucket_coverage.Rd`, `man/polity_coverage_gaps.Rd`.
* `lintr::lint_package(...)` with the four project exclusions — **No lints found.**
* `devtools::test()` — **`[ FAIL 0 | WARN 47 | SKIP 9 | PASS 6628 ]`**.
* `rcmdcheck::rcmdcheck(...)` — **0 errors, 0 warnings, 0 notes** (with
  `_R_CHECK_FORCE_SUGGESTS_=false`; `archive` and `RSQLite` are not installed here, CI has
  them).
* pkgdown: every `man/*.Rd` still in `_pkgdown.yml`, `character(0)` from the `comm` check. No
  new topic — the two new helpers are private.

Structural checks, on a real `get_primary_production(years = 2015L)`:

* **No published value changes**, and checked rather than assumed. `get_primary_production(years
  = 2015L)` built at `99dd49f0` and at this branch: 48,189 rows both ways, and **all 12 columns
  `identical()` elementwise** — 0 columns differ, and the whole frame is `identical()` once the
  `.cb_extracts` attribute is dropped. (That attribute's cached `fbs_new` extract came back in a
  different row order, an artefact of my own sorting of the outer tibble; it is `all.equal()`
  once both are row-sorted, 92,431 rows each.)
* One `area_code` → one `area` label: **0 violations**, before and after.
* `polity_bucket_coverage()`: 65 rows → **14**, all bucket 206, all `"predecessor"`,
  2012-2025, `member_area_codes` `"276, 277"`. `"partial"` count 1 → **0**.

Proving the guards load-bearing — I re-introduced each defect and watched the tests fail:

* Replacing the member filter with `TRUE` (the old counting rule) → **9 failures** across all
  three new tests: `nrow(coverage)` 14 → 65, `min(year)` 2012 → 1961, the class flips to
  `"partial"` and `"no bucket is labelled with a polity covering less than it sums"` breaks.
  Restored; suite green again.
* `.bucket_is_predecessor()` is tested directly on the four cases that must not reach the
  class: in-span label, a successor set that only overlaps the members, and a polity with no
  successors published.

The old assertion `all(sudan$coverage == "partial")` is now `== "predecessor"`. That is the
finding, not a weakened test: it is pinned alongside `bucket_mapping_status == "out_of_span"`
and the exact member set, which the old test did not check.

## Moves published values

**No.** Diagnostic and warning text only. `NEWS.md` says so explicitly.

* **Did not mint a polity downstream.** Option 2 of #414 needs an upstream row
  (`F206-2011-2025`, `"Sudan (combined reporting: ...)"`, `type = aggregate`), and
  whep-polities is the source of truth for territorial identities. It is proposed there as
  lbm364dl/whep-polities#139. Note the shape is not quite the `F237` (Vietnam) / `F249`
  (Yemen) precedent: those combine territories that reported together and separated later,
  whereas here FAOSTAT reports the successors separately and WHEP's own fold re-sums them, so
  the aggregate would be live *simultaneously* with both its members. That is the substantive
  question, and it is upstream's to answer.
* **Did not un-fold 276/277.** I costed it instead — see the open decision below — and filed
  #680 with the numbers.
* **Did not touch the resolver's bounds.** `polity_start_year` as the lower bound is right for
  a resolver (a row that exists must resolve) and wrong only for this report, so the extra
  bound lives here.
* **Did not change `folded_reporting_areas()`**, `regions_full`, or any `polity_area_code`.

## The open decision

Unchanged and still unmade: **no live polity means "Sudan and South Sudan"**, and one of three
things has to give.

I costed option 1, un-folding, on a real build (`get_primary_production(years = 2015L)`, twice
in one session with `.polity_crosswalk()` patched between runs). It is **not** the mass-neutral
switch-flip #628 was for bucket 999:

* Outside areas 206/276/277: 47,943 rows both ways, **0** differing values, **0** unmatched
  keys. Strictly local.
* Inside the region: tonnes **−4.22%**, heads **−2.22%**, ha −0.25%. Two derived series carry
  all of it — `Forage products` goes 1,432,940 t → **0**, and `Chickens, layers` 14,118,716 →
  9,439,000 heads — because a downstream step keyed on the bucket code has an entry for 206
  and none for the promoted members.
* `.iso3c_area_code_lookup()` stops being unique by construction: `SDN` is both area 206 and
  area 276 in `regions_full`, and the fold is what makes them one bucket. It goes to 258 rows
  for 257 ISO3 codes, and #540's pin `.iso3c_to_area_code("SSD") == 206` moves too.

Option 2 (mint the aggregate upstream) changes no WHEP value at all. Option 3 (leave it) is
what this PR makes honest. I have no basis for choosing between them and have not.

Closes #414


lbm364dl:
Merged `origin/main` (through `f8e55e57`) into the branch to clear a `NEWS.md` conflict — both entries kept, no marker left, file parses as Markdown.

That merge brought in **#662, which refreshes the polities snapshot** (`data/polities.rda`, `data/polity_area_crosswalk.rda`), i.e. exactly the tables every number in this PR is measured from. Re-measured on the merge commit:

```
polity_bucket_coverage() -> 14 rows, all bucket 206, all "predecessor",
                            2012-2025, members SDN-2011-2025 + SSD-2011-2025
"partial" count           -> 0
```

Unchanged. `devtools::test()` on the merge: **`[ FAIL 0 | WARN 47 | SKIP 9 | PASS 6641 ]`**.

Issue #742, opened 2026-08-10
The crosswalk cannot say whether a row answers for a reporting area or for the bucket it folds into


lbm364dl:
Split out of #741, which fixed the half of this that was actionable now.

## The gap

`polity_area_crosswalk` rows are keyed on `(area_code, polity_code)`, and
`.add_polity_columns_dt()` is called with **two different kinds of code** in that
first slot:

- a raw FAOSTAT **reporting area** code -- `R/build_cbs.R:766, 892, 1393, 2425`,
  `R/build_trade.R:191, 201`;
- a **bucket** code, i.e. `polity_area_code` -- `.bucket_area_labels()`
  (`R/polity_folds.R:585`), which is where every published
  `reporting_polity_code` is actually decided.

Nothing in the table says which of the two a row is meant to answer, so one row
has to serve both. `(238, ETH-1952-1993)` is the case:

- **as a bucket row it is right.** FAOSTAT area 62 (Ethiopia PDR) folds into
  bucket 238, so bucket 238 sums Ethiopia's 1961-1992 data, and this row is the
  bucket's only period before 1993. Deleting it sends 176 of the 352
  (area, year) pairs for areas 62/238 to `ETH-1993-2025` as an out-of-span
  stand-in -- measured in #741, which is why that PR kept it.
- **as an area row it is not.** Reporting area 238 does not exist before 1993;
  upstream's map gives it 1993-2024 only.

#741 removed the mirror row `(62, ETH-1993-2025)` precisely because it has no
bucket reading -- nothing folds into bucket 62 -- but the asymmetric rule it
shipped is a workaround for the missing distinction, not the distinction itself.

## Scope

Small and static. Only three areas fold into a bucket that is not their own code
(62 into 238, 276 and 277 into 206), and `ETH` is the only ISO3 prefix family the
upstream map splits across two reporting areas, so exactly **one** row is
currently ambiguous.

## Not urgent, and why

The two readings happen to give the same answer today. Measured over all 266
area/bucket codes x 1850-2025 (46,816 pairs) and confirmed on a real
6,310,390-row `get_primary_production()`, every pre-1993 Ethiopia row reaches
`reporting_polity_code` through the **bucket** route, and `ETH-1952-1993` is the
right answer for it. A raw area-238 row dated before 1993 -- which FAOSTAT never
emits -- would get the same code, correctly by extent but silently as a period
hit rather than as the stand-in it is. So this is labelling precision, not a
wrong number.

[...]

PR #743, opened 2026-08-10
fix(regions): stop the prefix expansion awarding a polity outside the area's fold

### LINKED ISSUE #741: The prefix expansion hands an area the polity another area owns upstream, because it only tests each area against its own map span
Filed from the epic #458 review on 2026-08-10. This is the small, immediately
fixable piece of #740, separable from the modelling question in
#739.

## The defect

`data-raw/table_mappings.R:485-511` builds the `prefix_outside_map` rows by taking
every polity sharing an area's ISO3 prefix and removing the ones "shadowed" by the
map:

```r
shadowed_by_map <- prefix_candidates |>
  ...
  dplyr::inner_join(
    faostat_area_map |> dplyr::select("area_code", "map_year_start", "map_year_end"),
    by = "area_code", relationship = "many-to-many"
  ) |>
  dplyr::filter(
    .data$polity_start_year <= .data$map_year_end,
    .data$polity_end_year - 1L >= .data$map_year_start
  )
```

The join is `by = "area_code"`, so a candidate is only ever tested against **its own
area's** map span. Nothing checks whether the upstream map has already given that
polity to a **different** area. So an area is handed exactly the polity a sibling
owns.

## Reproduced on `main`

Upstream is unambiguous -- one row per Ethiopia code, `confidence: high`:

```
62,  1961,1992, ETH-1952-1993, Ethiopia PDR, ETH, iso-equal, high, 85546
238, 1993,2024, ETH-1993-2025, Ethiopia,     ETH, iso-equal, high, 102786
```

whep's crosswalk turns that into nine rows per code:

```
 area_code   polity_code     mapping_source
        62 ETH-1952-1993       upstream_map      <- correct
        62 ETH-1993-2025 prefix_outside_map      <- 238 owns this
        62 ETH-1800-1889 ... ETH-1941-1952  prefix_outside_map  (7 more)
       238 ETH-1993-2025       upstream_map      <- correct
       238 ETH-1952-1993 prefix_outside_map      <- 62 owns this
       238 ETH-1800-1889 ... ETH-1941-1952  prefix_outside_map  (7 more)
```

Why each direction slips through: `ETH-1993-2025` starts 1993, which is **not**
`<= 1992`, so it is not shadowed for area 62. `ETH-1952-1993` ends 1993, so
`1992 >= 1993` is false and it is not shadowed for area 238. Each is invisible to
the other's test.

## Scope: exactly 2 rows, and it is latent

Across the whole crosswalk, prefix rows landing on a polity another area owns
upstream:

```
rows: 2   distinct polities: 2   distinct areas: 2   ISO3 families: ETH
```

It does not currently reach published output. Area 62's `polity_area_code` is
**238**, so Ethiopia PDR folds into Ethiopia's bucket and publishes **0 rows of its
own**; year-aware resolution then splits the bucket correctly (pre-1993 ->
`ETH-1952-1993`, the larger Ethiopia including Eritrea; post-1993 ->
`ETH-1993-2025`). Measured on a real 6,310,390-row `get_primary_production()`:

```
(polity, year) pairs with >1 area_code : 174   -- all ROW-1850-2025 (#717)
area 62 rows in published output       : 0
years where both 62 and 238 have rows  : 0
```

So: wrong in the crosswalk, harmless in the output, and worth fixing because the
next consumer to read the crosswalk directly (as #646 and #716 did) inherits it.

## Fix

Exclude a prefix candidate when the upstream map assigns that polity to a different
`area_code`. Guard: assert that no `(polity_code)` is claimed by an `upstream_map`
row for one area and by a non-upstream row for another.

Prove the guard is load-bearing by restoring the current `shadowed_by_map` join and
confirming it fails on Ethiopia -- this epic has shipped three guards that could not
fail on the bug they existed for.

lbm364dl:
## What was wrong

`data-raw/table_mappings.R` builds the `prefix_outside_map` crosswalk rows by
taking every polity that shares an area's ISO3 prefix and removing the ones
"shadowed" by the upstream FAOSTAT map. The shadowing join is `by = "area_code"`,
so a candidate is only ever tested against **its own area's** map span. Nothing
ever asked whether the map had already awarded that polity somewhere else.

Upstream is unambiguous, one `confidence: high` row per Ethiopia code:

```
62,  1961,1992, ETH-1952-1993, Ethiopia PDR
238, 1993,2024, ETH-1993-2025, Ethiopia
```

Both areas carry the `ETH` prefix, so each is a prefix candidate for the other's
polity, and each escapes the same-area test on a boundary year: `ETH-1993-2025`
starts in 1993, which is not `<= 1992`, so it was not shadowed for area 62;
`ETH-1952-1993` ends (exclusively) in 1993, so `1992 >= 1993` is false and it was
not shadowed for area 238.

On the committed `polity_area_crosswalk` at the base commit (596 rows), joining
every non-`upstream_map` row to the map's owner of its polity:

```
    polity_code area_code     mapping_source area_code_owner
1 ETH-1952-1993       238 prefix_outside_map              62
2 ETH-1993-2025        62 prefix_outside_map             238

rows: 2  distinct polities: 2  distinct areas: 2  ISO3 families: ETH
```

Exactly the 2 rows #741 measured, both Ethiopia.

## What I changed

`outside_map_areas` now also anti-joins `owned_elsewhere`: a prefix candidate is
dropped when the map's owner of that polity is **outside the candidate's own
fold**, i.e. when `area_code` is neither the owner area nor the bucket the owner
folds into. The crosswalk goes **596 -> 595 rows**; the row removed is
`(62, ETH-1993-2025)`, no row is added, and every one of the 595 survivors is
`all.equal()`-identical across all 26 columns.

**The exception is load-bearing, and it is why one row goes rather than two.**
`.bucket_area_labels()` resolves a published row's `reporting_polity_code` from
the `polity_area_code` — the *bucket* code — not the reporting area's own code.
Area 62 folds into bucket 238, so bucket 238 sums Ethiopia PDR's 1961-1992 data
and `(238, ETH-1952-1993)` is the **only period that bucket has before 1993**.
I implemented the issue's literal rule first ("a different `area_code` owns it")
and measured it: it removes both rows, and 176 of the 352 (area, year) pairs for
areas 62/238 change — every Ethiopia row from 1850 to 1992 flips from
`ETH-1952-1993` / `matched` to `ETH-1993-2025` / `out_of_span`. That is a
published value moving, so the shipped rule spares the candidate whose owner
folds into it. The reverse row has no such role: nothing folds into bucket 62,
area 238 is its own bucket, and area 62 stopped reporting in 1992.

Only three areas fold into a bucket that is not their own code (62 into 238,
276/277 into 206), and `ETH` is the only prefix family the map splits across two
areas, so the two readings differ on Ethiopia alone.

`data/regions_full.rda` is regenerated too. `data-raw/harmonization_tables.R`
loads the crosswalk and stamps each area's current/latest polity onto
`regions_full`, so area 62's row moves with it — see *Moves published values*.
`test_data_raw_freshness.R` and `test_territorial_identity.R` both caught the
under-regeneration before I did, which is what those gates are for.

The regeneration is attributable rather than a silent upstream re-sync: upstream
was pinned at **whep-polities `09d7eaf`**, the revision named in whep 1b752a3c,
extracted to a temp dir and passed through `WHEP_POLITIES_GPKG` /
`WHEP_POLITIES_FAOSTAT_MAP` / `WHEP_POLITIES_LABEL_ALIAS_MAP`. Before the fix
that reproduced all five `table_mappings.R` `.rda` **byte-identical** to `main`;
after it, and after re-running `harmonization_tables.R`, exactly two of the 56
datasets differ by value (`polity_area_crosswalk`, `regions_full`). The seven
other `.rda` that came back byte-different but `all.equal()`-identical were
reverted, so the diff carries no compression noise.

## How I verified

- **Full suite**: `devtools::test()` — 0 failures (see below), matching main.
- **`air format .`** (binary), `devtools::document()` (two unrelated `man/*.Rd`
  whitespace drifts from a local roxygen version were reverted), `lintr` with the
  four configured exclusions: clean. Every `man/*.Rd` is in `_pkgdown.yml`.
- **Join audit unaffected**: `sum(baseline$n) == sum(found$n) == 58`, unchanged.
  `R/join_audit.R` reads the NAMESPACE, and this change is in `data-raw/`.
- **The guard is load-bearing, proved by restoring the bug.** I commented out the
  new `anti_join`, regenerated the crosswalk (back to 596 rows) and re-ran the
  tests. They go red, and the guard named the offender rather than counting it:

  ```
  Failure ('test_polity_faostat_map.R:322:3'): no area is handed a polity owned outside its own fold
  `offenders` (`actual`) not equal to character(0) (`expected`).
  `actual`:   "area 62 <- ETH-1993-2025 (owned by 238)"
  ```

  The pre-existing test `prefix inference never contradicts the upstream map`
  **passed** on the injected bug — it is the same-area check, and that is exactly
  the blind spot. Re-run at the end against the final state (both generators
  re-run with the bug in place), the injection fails **11 expectations across 3
  files**. Then I restored the `anti_join`, regenerated both, and confirmed the
  data is `all.equal()` to the verified fixed version.
- Three tests were added and four pinned enumerations updated
  (`matched`/`prefix_outside_map` 247 -> 246, and the two bucket-conflict
  enumerations described below). The third new test, `bucket 238 can still date
  Ethiopia before 1993`, asserts the *consequence* through `add_polity_code()`,
  so the asymmetry cannot be "simplified" away later without a red test.

## Moves published values

**One identity column, one row. No quantity.**

`regions_full` row 77, FAOSTAT area 62 "Ethiopia PDR":

```
reporting_polity_code : ETH-1993-2025 -> ETH-1952-1993
reporting_polity_name : Ethiopia      -> Ethiopia (1952-1993)
```

That is the correction, not a side effect: area 62 dissolved in 1993 and was
never the modern republic. It is the only differing row in the only other
dataset that moves, checked column-by-column over all 272 rows.

Nothing numeric moves, and I checked that two ways rather than assuming it —
the crosswalk row count alone would not have caught the regression the literal
fix introduced.

1. **The whole resolution domain.** Every published `reporting_polity_code` comes
   from `.add_polity_columns_dt()` keyed on an area or bucket code and a year, so
   I resolved **all 266 codes in the crosswalk x 1850-2025 = 46,816 (area, year)
   pairs** under the old and new crosswalk and compared all seven resolved
   columns. **33 pairs differ, all of them area 62 in 1993-2025**
   (`ETH-1993-2025`/`matched` -> `ETH-1952-1993`/`out_of_span`), plus the one
   year-free lookup for area 62 — which is the `regions_full` change above.
2. **A real build.** `get_primary_production()` -> **6,310,390 rows** (the figure
   #741 measured). `area_code == 62`: **0 rows**, and 62 is absent from the
   published code set entirely — `R/build_production.R` publishes
   `area_code = polity_area_code`, and `.iso3c_area_code_lookup()` maps `ETH` to
   238 only. Ethiopia's bucket-238 rows still split exactly as before: 27,084
   (<=1960) + 8,474 (1961-1992) on `ETH-1952-1993`, 10,057 (>=1993) on
   `ETH-1993-2025`, never both in one year.

### One diagnostic now reports more, on purpose

`.bucket_year_polity_conflicts()` goes from 65 rows to 98: bucket 238 joins
bucket 206, for 1993-2025. Its two members disagree there — dissolved area 62
stands in with `ETH-1952-1993` while area 238 is a period hit on
`ETH-1993-2025`. **The removed row was manufacturing the agreement** by handing
a dead reporting area the live polity the map awards to another, so this is the
detector doing its job, not a regression. It cannot reach published output:
area 62 contributes no rows after 1992 and `.bucket_area_labels()` labels a
bucket from the bucket code regardless. `test_polity_resolution_uniqueness.R`
and `test_polity_resolution_coverage.R` now enumerate both buckets with the
reason, in the same style bucket 206 was already pinned in.

- **Did not remove `(238, ETH-1952-1993)`**, which #741 also lists. It is wrong
  read as "area 238 in 1970", right read as "bucket 238 in 1970", and the
  resolver only ever asks the second. Making the area reading true as well needs
  the crosswalk to distinguish an area key from a bucket key, which is a
  published schema change well outside this issue — filed as #742.
- **Did not touch `R/build_production.R`**, `R/polities.R` or the resolver, and
  did not widen `.polity_stand_in_mode()`.
- **Did not make `.bucket_year_polity_conflicts()` count a member only in the
  years it reports**, which would drop bucket 238's 33 rows *and* shrink bucket
  206's 65 — i.e. it would silently redefine an already-filed finding (#414).
  Noted in #742 instead.
- **Did not re-sync upstream.** The regeneration is pinned to `09d7eaf`, so the
  diff carries the fix and nothing else.
- **Did not rewrite the pre-existing same-area test.** It still earns its place;
  the new check is a sibling, not a replacement.

Closes #741.


lbm364dl:
Tightest form of the crosswalk claim, for the record:

```r
drop <- which(old$area_code == 62L & old$polity_code == "ETH-1993-2025")   # 148
all.equal(old[-drop, ], new, check.attributes = FALSE)                       # TRUE
```

The shipped table is the old one **minus row 148 and nothing else** — not one other cell in the remaining 595 rows x 26 columns moves.

lbm364dl:
Labelled `needs-expert` rather than `mechanical`, deliberately, even though the diff is one `anti_join` and one crosswalk row.

The **defect** is mechanical: an area holding a polity the upstream map awards to another is wrong under any reading. The **fix** is not, and the choice is not visible from the diff:

1. Which of the two rows to remove. The symmetric rule #741 prescribes removes both and moves 176 (area, year) pairs of published Ethiopia attribution. The asymmetric rule shipped here spares the row that serves the bucket. Both are defensible readings of "owns"; they differ in what 1850-1992 Ethiopia is attributed to.
2. What a dissolved reporting area should stand in with after it dissolves. Area 62 now resolves to `ETH-1952-1993` / `out_of_span` for 1993-2025 instead of `ETH-1993-2025` / `matched`. That is the honest answer and it matches how area 52 was handled earlier in this epic, but it is a convention, not an identity.
3. Accepting bucket 238 as a second entry in `.bucket_year_polity_conflicts()`. It is a true statement that was previously masked, but it changes what a filed finding (#414) looks like.

Green CI proves the code runs and that the regressions are pinned. It does not settle 1-3.

Issue #740, opened 2026-08-10
288 of 595 polity_area_crosswalk rows have no upstream authority: whep invents them by ISO3-prefix match


lbm364dl:
> ## ⚠️ CORRECTION (2026-08-10): the central claim below is REFUTED
>
> The row counts are right; **the conclusion drawn from them is wrong.** I wrote that
> "111 of the 174 published years are resolved by prefix guessing". They are not.
>
> `.add_polity_columns_dt()` floors every polity lookup at `backcast_anchor = 1961L`
> (`R/polities.R`), so a pre-1961 row is resolved **at 1961**, which is inside the
> upstream map's coverage. Measured on a real 6,310,390-row
> `get_primary_production()` (PR #744):
>
> | era | upstream row | whep ROW bucket | whep prefix guess |
> |---|---|---|---|
> | pre-1961 | 3,519,684 (**96.5%**) | 100,835 (2.76%) | 27,084 (0.74%) |
> | 1961+ | 2,542,473 (95.5%) | 111,840 (4.20%) | 8,474 (0.32%) |
> | all | 6,062,157 (**96.06%**) | 212,675 (3.37%) | 35,558 (0.56%) |
>
> **257 of the 262 `prefix_outside_map` rows resolve nothing at all.** In published
> output the prefix rule decides **exactly one area: 238 Ethiopia**. And that one row
> is load-bearing -- deleting it moves 143 of Ethiopia's 174 area-years to
> `out_of_span`, which is why #741/#743 spared it.
>
> **So the "definition of done" below is wrong**: this is not 288 rows of debt to
> delete. It is ~5 live rows, one of which needs an upstream answer
> (eduaguilera/whep-polities#200), and ~257 inert rows whose only cost is confusion.
>
> Two real findings did survive, and both are now separate issues:
> * **#745 — whep's crosswalk is built from a stale upstream vintage.** 245 rows /
>   197 areas committed, against 281 / 228 on `whep-polities@main`. 31 of the areas
>   with "no upstream row" already have one upstream today.
> * Of the areas genuinely missing upstream, only **two publish anything**: 94 Holy
>   See (268 rows) and 22 Aruba (244). Everything else is whep's own ROW bucket or
>   its 901-906 constructs.
>
> Everything below is kept for the record but should be read through this correction.

---

Filed from the epic #458 review on 2026-08-10, on the owner's rule: **whep should not
be fabricating crosswalks.** If a mapping cannot be made from the upstream polities
[...]

Issue #739, opened 2026-08-10
A pre-1961 row is a reconstruction on present-day borders, so 'source' should decide which polity vocabulary it gets


lbm364dl:
> ## ⚠️ CORRECTION (2026-08-10): the premise below is REFUTED
>
> **whep already does what this issue proposes.** `.add_polity_columns_dt()`
> (`R/polities.R`) has carried `backcast_anchor = 1961L` since `256b37a1` and floors
> every polity lookup at it:
>
> ```r
> year = pmax(as.numeric(get(year_col)), as.numeric(backcast_anchor))
> ```
>
> with a comment stating this issue's own reasoning: *"a 1900 'Austria' figure
> represents 1961 Austria, not the 1900 Habsburg crownland ... so pre-anchor data
> maps to the entity active in 1961 instead of a larger historical-extent period."*
>
> So the claim below that area 238's 1850 row resolves to `ETH-1800-1889` is **false**.
> Verified by resolution: area 238 publishes `ETH-1952-1993` at **every year
> 1850-1992**. Class 3 (pure back-cast) already takes the anchor polity.
>
> Measured on a real 6,310,390-row build (#744): pre-1961 output is **96.5%
> resolved through an `upstream_map` row**, and **257 of the 262
> `prefix_outside_map` rows are never selected at all**.
>
> Two further corrections from the same measurement:
> * **There are no `historical_*` rows in the published artifact** — `historical_data`
>   defaults to `NULL`, so class 2 is empty and class 1 is `FAOSTAT_prod` at year
>   >= 1961 only. The rule has one branch in practice.
> * `LUH2_grassland` is a **fourth** reconstruction source the table below omits.
>
> **The real defect this uncovered is #748**, and it is sharper: a pre-1961 row has
> **two** territorial anchors that disagree. The *level* is the 1961 value (polity
> anchored at 1961) while the *movement* is a present-day-ISO3 LUH2 ratio. For the
> **43 areas whose polity changes after 1961** those extents differ -- Ethiopia's
> tonnage is grown on present-day Ethiopia (no Eritrea) while labelled
> `ETH-1952-1993` (with Eritrea). 747,758 rows, 20.5% of the pre-1961 block.
> Independently confirmed: 43 of 263 areas resolve to a different polity at 1961
> than at 2023 (`DZA-1919-1962`->`DZA-1962-2025`, `EGY-1925-1967`->`EGY-1979-2025`,
[...]

PR #746, opened 2026-08-10
Let `source` decide the pre-1962 polity vocabulary — and a refutation of #739's premise

lbm364dl:
> **Draft, deliberately.** #739's central factual claim is refuted below, and
> the code half of this PR is inert on today's data. The measurement is the
> deliverable; the switch is offered, not argued for, and can be dropped
> without losing anything above it.

## What was wrong

**#739's premise is false, and I verified it before implementing anything.**

The issue says a pre-1962 row is given a *historical* polity — "area 238's 1850
row resolves to `ETH-1800-1889`, whose polygon is a different territory" — and
proposes that pure back-cast rows take the 1961-anchor polity instead.

whep has done exactly that since `256b37a1`. `.add_polity_columns_dt()`
(`R/polities.R`) carries `backcast_anchor = 1961L` and floors every polity
lookup at it:

```r
year = pmax(as.numeric(get(year_col)), as.numeric(backcast_anchor))
```

with a comment giving the issue's own reasoning as its rationale ("a 1900
'Austria' figure represents 1961 Austria, not the 1900 Habsburg crownland").
Measured on the published artifact: **area 238 carries `ETH-1952-1993` at every
year from 1850 to 1992**, and area 62 publishes 0 rows. The proposed class-3
branch is the status quo, not a change.

What *is* missing is the other branch: the floor is **blanket**, so a row
genuinely observed in 1850 (via `historical_data`) would be given the 1961
polity too. And the vocabulary is currently a per-**dataset** constant — the one
caller that wants the other answer switches it off wholesale
(`.resolve_hist_trade_polities()` passes `backcast_anchor = -Inf`) — where #739
argues it should be a per-**row** consequence of `source`.

**And the question underneath is sharper than the issue states.** A pre-1961 row
is neither "on present-day borders" nor "on 1961 borders". It is a hybrid: its
**level** is the area's 1961 reported value, its **year-on-year movement** is a
LUH2 land ratio, and LUH2 is keyed on present-day ISO3. Two anchors, and they
disagree for every territory that changed after 1961.

Real `get_primary_production()` at `origin/main` (`4b1d2633`), 6,310,390 rows,
warm pins.

**The published artifact has four `source` values, and #739's three classes are
not evenly populated:**

| # | class | `source` values actually present | rows | share |
|---|---|---|---|---|
| 1 | observed | `FAOSTAT_prod` — **every one at year ≥ 1961** | 2,640,245 | 41.84% |
| 2 | historically anchored (`historical_*`) | *none* | **0** | 0.00% |
| 3 | pure back-cast | `LUH2_cropland` 1,987,344 · `LUH2_agriland` 1,621,821 · `LUH2_grassland` 60,980 | 3,670,145 | 58.16% |

**#739's open question 3, answered: the artifact contains ZERO `historical_*`
rows.** `historical_data` defaults to `NULL` and nothing else emits the prefix,
so class 2 is empty and class 1 lives entirely at or after the anchor year,
where flooring the lookup year is a no-op. **The rule has one branch in
practice today.**

**What the anchor floor is currently worth.** Resolving the 35,821 distinct
`(area_code, year)` pairs with the floor and without it:

| | |
|---|---|
| pairs where the two answers differ | **7,282**, all at year < 1962 |
| published **rows** whose `reporting_polity_code` would move | **1,614,244 (25.58%)** |
| reporting areas affected | **85** |
| distinct anchor → own-year polity substitutions | **231** |
| `polity_area_code` (the aggregation bucket) pairs that move | **0** |

The bucket never moves: this is an identity question, not an aggregation one.

**The second-order effect #740 depends on, verified — and it is bigger than the
issue claims.** Joining each published row's polity to the crosswalk row that
authorises it (596 rows: 245 `upstream_map`, 262 `prefix_outside_map`, 62
`fabio_row_fold`, 27 `prefix_fallback`):

| authority | anchor route (i.e. today) | own-year route |
|---|---|---|
| `upstream_map` | 6,062,157 rows / 241 pairs | 4,483,681 / 241 |
| `fabio_row_fold` | 212,675 / 24 | 212,675 / 24 |
| **`prefix_outside_map`** (fabricated) | **35,558 / 1 pair** | **1,614,034 / 231 pairs** |

1961 is the upstream map's **first** year (`map_year_start` spans 1961-2014, no
NAs), so the anchoring **already** retires 261 of the 262 fabricated rows. The
262nd is `238 → ETH-1952-1993` — precisely **#741**: upstream gives that polity
to area **62** for 1961-1992, area 238's only `upstream_map` row is
`ETH-1993-2025` (1993-2024). It is load-bearing for 35,558 published rows, so it
wants upstream stating area 238's pre-1993 span (whep-polities#200), not a
deletion.

**The two anchors, verified on Ethiopia.** Area 238's published harvested-area
series moves year on year *exactly* with LUH2's **ETH-only** (present-day, no
Eritrea) cropland proxy:

| year | published ha | LUH2 ETH cropland Mha | value ratio y/y+1 | proxy ratio y/y+1 |
|---|---|---|---|---|
| 1955 | 26,132,703 | 9.978 | 0.9802472 | 0.9802927 |
| 1957 | 27,207,349 | 10.387 | 0.9793245 | 0.9793825 |
| 1959 | 28,388,302 | 10.836 | 0.9778245 | 0.9779047 |
| 1960 | 29,032,104 | 11.081 | 0.9773180 | 0.9770967 |

max ratio difference over 1955-1960 = **2.2e-04**. Meanwhile the row is labelled
`ETH-1952-1993`, which **includes** Eritrea; LUH2 carries ERI as a separate
present-day mask at 4.8% of ETH+ERI cropland in 1961, and area 178
(`ERI-1993-2025`) publishes its own 1850-1992 grassland rows *inside* the
polygon area 238 claims.

**Enumerated: 43 of 217 published areas have a 1961-anchor polity that is a
different entity from their present-day one — 747,758 pre-1961 rows, 20.5% of
the pre-1961 block, 11.85% of the artifact, 35.1 of 391.2 Gt of pre-1961
tonnage.** Largest by row count:

```
 59 EGY-1925-1967 -> EGY-1979-2025   36,408      4 DZA-1919-1962 -> DZA-1962-2025   27,417
143 MAR-1958-1975 -> MAR-1979-2025   35,187    238 ETH-1952-1993 -> ETH-1993-2025   27,084
105 ISR-1948-1967 -> ISR-1979-2025   33,744    131 MYS-1957-1963 -> MYS-1965-2025   25,752
 79 DEU-1949-1990 -> DEU-1990-2025   32,856    237 F237-1954-1975 -> VNM-1975-2025   23,532
 16 BGD-1947-1971 -> BGD-1971-2025   30,969    144 MOZ-1891-1975 -> MOZ-1975-2025   22,533
101 IDN-1949-1969 -> IDN-2002-2025   30,081    181 SRH-1953-1964 -> ZWE-1980-2025   20,091
114 KEN-1926-1963 -> KEN-1963-2025   28,638    249 F249-1918-1990 -> YEM-1990-2025   17,783
165 PAK-1949-1971 -> PAK-1971-2025   28,416    248 F248-1947-1991 -> F248-1991-1992   9,990
215 TZA-1961-1964 -> TZA-1964-2025   27,528     15 BLX-1850-1999 -> (none)            8,769
```

`polygon_area_km2` is `NA` for 35 of the 40 in packaged `polities`, so the km²
gap is not quantifiable offline; the polity names carry it (SRH→ZWE, NRH→ZMB,
BEC→BWA, PAK pre/post Bangladesh).

## What I changed

- `.add_polity_columns_dt()` / `.add_reporting_polity_columns()` accept
  `backcast_anchor` as **one value per row**, not only a scalar
  (`.backcast_anchor_by_row()` recycles or aborts). A scalar behaves exactly as
  before, so both existing routes (`1961L`; `-Inf` for historical trade) are
  untouched. This is the mechanism any per-row answer to the vocabulary question
  needs, whichever way it is answered.
- The carried-identity fast path in `.add_reporting_polity_columns()` now only
  serves a request at the default anchor. It republishes an identity resolved at
  1961, so answering a `-Inf` request from it would silently swap the vocabulary.
- `build_primary_production(polity_vocabulary =)`, `rlang::arg_match()`-validated:
  - **`"anchor"` (default, current published behaviour)** — every pre-1962 row
    takes the anchor polity, whatever its source.
  - `"observed_period"` — a genuinely observed historical row (a `historical_data`
    row, which keeps its own `historical_*` source) takes the polity of its
    **own** year; every reconstruction stays anchored.
  - `"historical_period"` — the same, plus the three `historical_LUH2_cropland` /
    `historical_LUH2_agriland` / `historical_fill_linear` labels.
- The two non-default values emit **`method_reporting_polity`** (`"anchor"` /
  `"period"`) per row, per the multi-method rule in `CLAUDE.md`.

**Design notes, since these are choices and not obvious:**

*Three levels, not two.* The `historical_LUH2_*` class is the genuinely
undecided case, so it is the **only** thing separating `"observed_period"` from
`"historical_period"`. The open decision is then a choice between two named
values rather than a comment.

*The period test is positive and closed.* A row earns the own-year polity only
by carrying a source that says it was observed in that year. A source nobody has
classified keeps the anchor, so the failure mode is always the status quo.

*The method column is emitted only when it says something.* Adding a column to
the published 6.3M-row schema is a downstream event; under `"anchor"` it would
carry a fact `year` already carries. (I had written that
`polity_coverage_gaps()` surfaces the anchoring per row — **it does not**:
`.polity_gap_kind()` applies the same floor, so only 4,309 of the 3,647,603
anchored rows are flagged `polity_not_started`. That is an open question below,
not a doc claim.)

*The argument sits on `build_primary_production()`, not `get_primary_production()`,*
matching `federation_land`. The cache key is untouched and the published entry
point cannot change by accident.

## How I verified

| check | result |
|---|---|
| `get_primary_production()`, default, branch vs `main` | **`identical()`**, 6,310,390 rows × 12 columns |
| `main` run 1 vs `main` run 2 (control) | frame `identical()`; **whole object is not** — see below |
| `get_wide_cbs()`, branch vs `main` | **`identical()`**, 2,184,850 rows × 18 columns |
| `polity_vocabulary = "observed_period"` on real data | `identical()` on the published schema; **0** polity codes move |
| `polity_vocabulary = "historical_period"` on real data | `identical()` on the published schema; **0** move |
| `method_reporting_polity` on the real build | 3,647,603 `anchor` / 2,662,787 `period` — exactly the rows before 1961 |
| `devtools::test()` | **FAIL 0 · PASS 7172 · SKIP 9** |
| `lintr::lint_package()` | No lints found |
| `air format .` | clean |
| every `man/*.Rd` in `_pkgdown.yml` | `character(0)` |

The `identical()` control matters: **`main` is not `identical()` to itself**
across two runs, on the `.cb_extracts` attribute only — its four tables carry
the same rows in a different order run to run. The published frame is
`identical()` in all three pairings. Filed as a follow-up; it is upstream of
anything here (`.read_production()` sets it before this code runs).

Tests added, all failing against the pre-change code:

- `.prod_period_rows()` over all nine source labels × three vocabularies,
  including that `NA` is not an own-year observation.
- `.backcast_anchor_by_row()` recycles a scalar, passes a per-row vector, and
  **aborts** on a length that is neither.
- Two rows with the *same* `(area_code, year)` and different anchors resolve to
  `ETH-1952-1993` and `ETH-1800-1889` — so the difference is the anchor's doing
  and nothing else's — with `polity_area_code` unchanged.
- A `-Inf` request is not served from the carried fast path.
- End-to-end `build_primary_production(.raw_data =)` on a four-row fixture, one
  row per source class, under each vocabulary; plus the bucket invariant across
  all three, and the `arg_match` abort.

All offline: package crosswalk data and `tribble()` fixtures only, no pin, no
`WHEP_*` path, no join added (`test_join_audit.R` green, baseline untouched).

## Moves published values

**No — default unchanged, asserted by `identical()`** on the full 6,310,390-row
`get_primary_production()` output and on `get_wide_cbs()`. And because the
published build carries no `historical_*` row at all, **both non-default
vocabularies also produce `identical()` published values today**: the switch is
inert until someone passes `historical_data`. That is stated here rather than
dressed up — it is why this is a draft.

- **Did not flip the default**, and did not implement an answer to the
  two-anchors question. Resolving that needs a **ceiling** on the lookup year
  (name a back-cast row by its *present-day* polity, matching the trajectory),
  not a floor; `pmax()` cannot express it and it is a decision, not a patch.
- **Did not invent a change to justify the issue.** The switch closes the one
  branch of #739's rule that is genuinely unimplemented, and nothing more. If
  the answer to the real question makes it redundant, drop the code and keep the
  measurement.
- **Did not touch `data-raw/table_mappings.R` or `data/*.rda`** (#741's owner)
  and **did not delete any crosswalk-building code** (#740/#741). The authority
  table above is the input those two want, not a change to them.
- **Did not route observed rows through `label_alias_map.csv`.** #739 proposes
  resolving them through upstream's
  `(source_label, source, year_start, year_end, polity_code)` map, because
  `pipelines/pre1961-matching/` resolves by `(iso3c, year)` with real historical
  routing (`TUR < 1913 -> OTT`). This resolves them through the **same FAOSTAT
  area crosswalk, minus the floor** — right vocabulary, routing only as right as
  the area crosswalk. Two join paths for two populations is a bigger change than
  a default-preserving switch, and there is no such row to test it on.
- Did not extend the switch to `build_commodity_balances()` or the trade chain,
  and did not re-measure the footprint chain — the bucket does not move, so the
  IO layer cannot see this.

## The open decision

1. **The real one: 1961 anchor or present-day anchor?** A back-cast row's level
   describes the 1961 territory and its movement the present-day one. 43 areas /
   747,758 rows sit on the gap. The anchor matches the level; the present-day
   polity would match the movement. **Neither matches both**, and this PR does
   not choose — it does not even offer the second option.
2. **Which polity a `historical_LUH2_*` row should carry** — the same
   level-vs-movement split, one layer down, and the only difference between
   `"observed_period"` and `"historical_period"`. 0 rows today.
3. **Should the default flip to a source-driven rule?** It costs nothing today
   and takes effect the day `historical_data` is wired in. Deciding now means
   the first historical build is right by default.
4. **Is the anchor's temporal absurdity acceptable, and should it be visible?**
   Labelling an 1850 row `ETH-1952-1993` is territorially defensible and
   temporally nonsense. #739 argues for embracing it. Note that it is currently
   **invisible**: `.polity_gap_kind()` floors at the same anchor, so only 4,309
   of the 3,647,603 anchored rows are flagged `polity_not_started`. Making
   `method_reporting_polity` unconditional is the candidate fix and is a schema
   change worth deciding on purpose.
5. **Whether `LUH2_grassland` is a reconstruction.** Treated as one here (LUH2
   pasture on present-day ISO3, 38,438 pre-1961 rows). #739's class table does
   not mention it.
6. **A non-default vocabulary does not survive downstream.**
   `.carried_reporting_polity()` re-resolves at the default anchor when a
   consumer re-adds polity columns, and warns when the two answers contradict.
   Loud, not silent, and moot while the observed class is empty — but it means
   the vocabulary is a property of the production build, not of the pipeline.

Part of the polity migration epic #458. Refs #739.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,24 @@
 # whep (development version)
 
+* **`build_primary_production()` gains `polity_vocabulary`, which lets a row's
+  `source` decide whether its polity is looked up at its own year or at the
+  1961 back-cast anchor (#739).** The anchor floor itself is not new -- it has
+  applied to every pre-1962 row since `256b37a1`, whatever the source -- so
+  `"anchor"`, the default, keeps doing exactly that. `"observed_period"` gives
+  the own-year polity to genuinely observed `historical_data` rows and
+  `"historical_period"` extends that to the three `historical_LUH2_*` /
+  `historical_fill_linear` labels; both add a `method_reporting_polity` column
+  saying which route each row took. **No published value changes:** the default
+  `get_primary_production()` output is `identical()` across all 6,310,390 rows
+  and 12 columns, `get_wide_cbs()` likewise, and because a default build
+  carries **no** `historical_*` row at all, so are the two non-default
+  vocabularies -- the switch is inert until `historical_data` is wired in.
+  The open question it does not settle is a different one from the one #739
+  states: a back-cast row's *level* is its area's 1961 value while its
+  *year-on-year movement* is a present-day-ISO3 LUH2 land ratio, so for the
+  43 of 217 published areas whose territory changed after 1961 (747,758
+  pre-1961 rows) the value and its label describe two different extents.
+
 * **`build_carbon_inputs()` no longer attaches reporting polity columns to an
   intermediate that discards them.** `build_soil_carbon_inputs()` produced a
   5.0e7-row gridded table for a 40-year span, and adding the four reporting
--- a/R/build_production.R
+++ b/R/build_production.R
@@ -37,6 +37,37 @@
 #'     its successor states' LUH2 land, resolved from the `successor` relation
 #'     published in [polities]. This back-casts 14.3% more of the 1961-62
 #'     production tonnage and therefore moves published pre-1962 values.
+#' @param polity_vocabulary Character. Which polity a pre-1962 row is named by.
+#'   A pre-1962 WHEP row is a reconstruction, not an observation: its **level**
+#'   is the area's 1961 reported value, and its **year-on-year movement** is the
+#'   ratio of LUH2 land, which is keyed on present-day ISO3. Neither is the
+#'   territory that governed the ground in the row's own year, so the polity
+#'   lookup floors its year at the 1961 anchor -- and has done since `256b37a1`,
+#'   for every pre-1962 row whatever its source. This argument makes that floor
+#'   answer to `source` instead, which is what whep#739 asks for; what it does
+#'   **not** address is that the level and the movement describe two different
+#'   extents wherever a territory changed after 1961 (43 of 217 published areas,
+#'   747,758 pre-1961 rows -- see whep#739).
+#'   * `"anchor"` (default, current published behaviour) names **every**
+#'     pre-1962 row by the polity active at the 1961 anchor, whatever its
+#'     source.
+#'   * `"observed_period"` names a genuinely observed historical row -- one
+#'     supplied through `historical_data`, which keeps its own
+#'     `historical_*` source -- by the polity active in its **own** year, and
+#'     leaves every reconstruction anchored.
+#'   * `"historical_period"` does the same and additionally gives the own-year
+#'     polity to `historical_LUH2_cropland`, `historical_LUH2_agriland` and
+#'     `historical_fill_linear` rows, which are a real historical observation
+#'     grown by a modern-border land ratio. Whether those belong with the
+#'     observations or with the reconstructions is an open question: their
+#'     level describes the historical territory and their year-on-year movement
+#'     the modern one.
+#'
+#'   The two non-default values additionally emit a `method_reporting_polity`
+#'   column recording, per row, whether its polity was resolved at the anchor
+#'   (`"anchor"`) or in its own year (`"period"`). With
+#'   `historical_data = NULL` all three values give the same result, because
+#'   the build then contains no observed pre-1962 row at all.
 #' @param .raw_data Optional tibble with the same structure as the output
 #'   of the internal `.read_production()` step. When supplied, the
 #'   remote-data read is skipped entirely and the pipeline starts from
@@ -52,7 +83,11 @@
 #'   `live_anim_code`, `unit`, `value`, and `source`.
 #'   Item names can be recovered via [add_item_prod_name()] and related helpers.
 #'   When `show_duplicates = TRUE`, returns a wide tibble with one
-#'   column per source showing the competing values.
+#'   column per source showing the competing values. A non-default
+#'   `polity_vocabulary` adds `method_reporting_polity`; the default schema is
+#'   unchanged, because under `"anchor"` the rule is one fact about the whole
+#'   build -- every row before the anchor year is anchored -- and `year` already
+#'   says which rows those are.
 #'
 #' @export
 #'
@@ -66,12 +101,14 @@ build_primary_production <- function(
   show_duplicates = FALSE,
   historical_data = NULL,
   federation_land = c("none", "successor_union"),
+  polity_vocabulary = c("anchor", "observed_period", "historical_period"),
   .raw_data = NULL
 ) {
   if (example) {
     return(.example_build_primary_prod())
   }
   federation_land <- rlang::arg_match(federation_land)
+  polity_vocabulary <- rlang::arg_match(polity_vocabulary)
   cli::cli_h1("Building primary production")
   if (is.null(.raw_data)) {
     raw <- .read_production(
@@ -122,7 +159,7 @@ build_primary_production <- function(
       source,
       dplyr::any_of("fao_flag")
     ) |>
-    .add_reporting_polity_columns()
+    .resolve_prod_polities(polity_vocabulary)
 
   attr(result, ".cb_extracts") <- cb_extracts
   result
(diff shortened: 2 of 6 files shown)


lbm364dl:
Filed the two follow-ups this measurement produced:

- **#748** — the real modelling question, split out so it can be decided on its own evidence: a pre-1961 row's *level* is its area's 1961 value while its *movement* is a present-day-ISO3 LUH2 land ratio, and for 43 of 217 published areas (747,758 pre-1961 rows, 35.1 Gt) those are different territories. Three options costed there. This PR does not choose, and does not even offer option 2 — that needs a **ceiling** on the polity-lookup year, which `pmax()` cannot express.
- **#747** — `main` is not `identical()` to itself across sessions, on the `.cb_extracts` attribute only (same rows, unpinned order, all four tables). It is why the `identical()` assertions above are stated on the frame rather than the whole object, and it will bite the next PR that tries the obvious check.

lbm364dl:
Not merging this as it stands, and the PR's own framing is why: it is **inert** — 0 rows move under either non-default `polity_vocabulary` value — so merging it would add public API surface that changes nothing, while the finding it produced (#748) is the part with value.

Leaving it open as a draft for @eduaguilera rather than closing it, because the refutation write-up in the body is the clearest record of why #739's premise was wrong. Two options when you get to it:

1. **Drop the code, keep the measurement.** #748 already carries the numbers; this PR can close unmerged.
2. **Keep the argument** if the `observed_period` branch is wanted for the day `historical_data` is actually supplied — today the published artifact has **zero** `historical_*` rows, so that branch is unreachable.

The thing that does want a decision is #748, and it needs a *ceiling* on the polity-lookup year, not the floor `backcast_anchor` gives — `pmax()` cannot express it, so it is not a small change.

lbm364dl:
Closing: the switch is inert. Measured 0 rows moving under either non-default value, because the branch it was built for is the status quo (whep already anchors pre-1961 rows at 1961 via `backcast_anchor = 1961L`) and the branch that would differ needs `historical_*` rows, of which the published artifact has none.

Merging it would make a permanently public argument that does nothing — and an exported formal is hard to withdraw later. Better to rebuild it against real data if the historical corpus is ever ingested.

Nothing is lost: the measurements and the refutation of #739's premise are preserved in the correction header on #739, in #748, and in #761, which carries the actual implementation.

